Issue #79, opened 2020-06-25
apiserver and cleanup Cloud Run services fail to start due to csrf token 'secretmanager.versions.access' denied


zssz:
Using Terraform to deploy I get the following error:

```
google_cloud_run_service.cleanup: Destroying... [id=locations/us-central1/namespaces/sandbox2-cw-en-vs/services/cleanup]
google_cloud_run_service.apiserver: Destroying... [id=locations/us-central1/namespaces/sandbox2-cw-en-vs/services/apiserver]
google_cloud_run_service.cleanup: Destruction complete after 1s
google_cloud_run_service.cleanup: Creating...
google_cloud_run_service.apiserver: Destruction complete after 1s
google_cloud_run_service.apiserver: Creating...
google_cloud_run_service.cleanup: Still creating... [10s elapsed]
google_cloud_run_service.apiserver: Still creating... [10s elapsed]
google_cloud_run_service.apiserver: Still creating... [20s elapsed]

Error: Error waiting to create Service: resource is in failed state "Ready:False", message: Cloud Run error: Container failed to start. Failed to start and then listen on the port defined by the PORT environment variable. Logs for this revision might contain more information.

  on service_apiserver.tf line 47, in resource "google_cloud_run_service" "apiserver":
  47: resource "google_cloud_run_service" "apiserver" {



Error: Error waiting to create Service: resource is in failed state "Ready:False", message: Cloud Run error: Container failed to start. Failed to start and then listen on the port defined by the PORT environment variable. Logs for this revision might contain more information.

  on service_cleanup.tf line 47, in resource "google_cloud_run_service" "cleanup":
  47: resource "google_cloud_run_service" "cleanup" {


zsombor@MacBook-Pro terraform % 
```

Digging into the Cloud Run logs, it looks like it's related to permission denied while accessing the csrf-token:

```
2020-06-25 15:52:01.749 EEST Cloud Run DeleteService apiserver zsombor@covidwatch.org {@type: type.googleapis.com/google.cloud.audit.AuditLog, authenticationInfo: {…}, authorizationInfo: […], methodName: google.cloud.run.v1.Services.DeleteService, request: {…}, requestMetadata: {…}, resourceLocation: {…}, resourceName: namespaces/sandbox2-cw-en-vs/services/apiserver, serviceName: run…


[...]

Issue #71, opened 2020-06-23
Break apart monoconfig


sethvargo:
The current config requires firebase, database, and csrf. Not all services need that (e.g. cleanup only needs the database, apiserver doesn't need csrf), but we have to set all those values on the service because they are required fields in the mega config.

mikehelmick:
oh, now prow yet

PR #80, opened 2020-06-25
Give all services access to the token

sethvargo:
This is necessary until GH-71 is fixed. Fixes GH-79.


sethvargo:
It looks like prow is broken. Gonna merge since the GitHub tests passed.

PR #88, opened 2020-06-30
Feature/issue api

ibeckermayer:
Restructures the server configuration structs so that both the `cmd/server` and the `cmd/apiserver` can implement the `/issue` endpoint to issue verification codes.

The motivation behind this is that at Covid Watch we already have a [permission portal](https://github.com/covid19risk/permission-portal) that we plan to give to our customers as a custom branded interface for doing everything that the included interface does (sans API keys), as well as expose various settings specific to Covid Watch's product.

Given our current codebases, we think the simplest way for us to get verification codes into our permission portal would be to call an api-key protected `/issue` endpoint (this PR) from our backend (ultimately it will be a firebase cloud function).

On a related note, we also hope to make other settings (such as those related to risk scoring) available in the permission portal, and so would like to work with you all to make those easy to configure from our custom UI.


mikehelmick:
/hold

Can you explain why you think this change is needed?

ibeckermayer:
> Can you explain why you think this change is needed?

Apologies, mistakenly pulled the trigger too early. See the updated description.

mikehelmick:
Thanks!

I will take a closer look later and I have some questions Still as well

.gitignore
```diff
@@ -21,3 +21,6 @@
 terraform/state.tf
 
 local/
+
+# OSX
+*.DS_Store
```
sethvargo, line 26: This is fine here, but it's also a good practice to add this to your global `.gitignore`

ibeckermayer: I've been using git for 5+ years and TIL there's a global `.gitignore`... LOL, thanks for the tip.

ibeckermayer:
> hello - general question in addition to comments:

I'll hold off on responding to individual comments until we figure out if we're going to go with this approach at all.

> The API Key protection level isn't sufficient here. A compromised API Key would let someone generate new verification codes relatively rapidly. Once generated, those codes can be use to certify keys that would be accepted by the exposure notification server.

Ah good point, I hadn't considered that.

> I'm open to this being an API that can be invoked by a different front-end, but I think it might need human scale auth (i.e. oauth) or a different API authenticator.

Got it, it would be a huge boost to us to make this invoke-able by our frontend.

One option I'd considered while sifting through the codebase: we are already using Firebase auth in our permission portal, and doing something similar to what you're doing with controlling app-specific permissions with a mirrored table. (In our case, instead of mirroring users in a SQL db, we're mirroring them in a Firestore collection, and applying a hacky `onCreate` trigger and security rules to effectively turn off public sign-ups.) 

I was thinking that perhaps there's a not-to-difficult way to point this server to our firebase project, and use the Firebase jwt tokens directly for auth, rather than session cookies. I think we could substitute the checks you're making in the database (i.e. for `user.Admin`) with checking [custom claims](https://firebase.google.com/docs/auth/admin/custom-claims), which we already have set up in our code in anticipation of such a situation.

OTOH, that approach would need to be in parallel to what you already have going, and as the project evolves it could get quite cumbersome to maintain. I'm generally relatively naive on the subject of API authentication, if you see a path forward with more integrity I'm very open minded to better ideas.

mikehelmick:
Sorry for the delay - long weekend + vacations last week for a couple of team members. Catching up.

I'm going to move this conversation to issue #95

PR #96, opened 2020-07-05
Fix typo

zssz:


sethvargo:
@crwilcox it looks like prow is being wonky. Any ideas?

PR #99, opened 2020-07-07
Set PROJECT_ID on services

sethvargo:
/assign @mikehelmick


sethvargo:
Merging since it's the flakey db test

PR #105, opened 2020-07-10
Make verification server multi-tenant

mikehelmick:
Currently work in progress - initial commit shows database changes.

### DIFF
--- a/cmd/add-realm/main.go
+++ b/cmd/add-realm/main.go
@@ -12,52 +12,43 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// A binary for running database migrations
+// Adds a new realm.
 package main
 
 import (
 	"context"
 	"flag"
-	"fmt"
+	"log"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
-	"github.com/google/exposure-notifications-verification-server/pkg/logging"
-	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/sethvargo/go-envconfig/pkg/envconfig"
-	"github.com/sethvargo/go-signalcontext"
 )
 
 func main() {
-	target := flag.String("id", "", "migration ID to move to")
-	rollback := flag.Bool("rollback", false, "if true, will run a rollback migration towards the target")
+	name := flag.String("name", "", "name of the realm to add")
 	flag.Parse()
-	ctx, done := signalcontext.OnInterrupt()
 
-	err := realMain(ctx, *target, *rollback)
-	done()
-
-	logger := logging.FromContext(ctx)
-	if err != nil {
-		logger.Fatal(err)
+	if *name == "" {
+		log.Fatalf("--name must be passed and cannot be empty")
 	}
-	logger.Info("migrations complete")
-}
 
-func realMain(ctx context.Context, target string, rollback bool) error {
+	ctx := context.Background()
 	var config database.Config
 	if err := envconfig.Process(ctx, &config); err != nil {
-		return fmt.Errorf("failed to process config: %w", err)
+		log.Fatalf("config error: %v", err)
 	}
 
 	db, err := config.Open(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to connect to database: %w", err)
+		log.Fatalf("db connection failed: %v", err)
 	}
 	defer db.Close()
 
-	if err := db.MigrateTo(ctx, target, rollback); err != nil {
-		return fmt.Errorf("failed to run migrations: %w", err)
+	realm := database.Realm{
+		Name: *name,
 	}
-
-	return nil
+	if err := db.SaveRealm(&realm); err != nil {
+		log.Fatalf("unable to create realm: %v", err)
+	}
+	log.Printf("created new realm: %v (%v)", realm.Name, realm.ID)
 }
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -68,13 +68,34 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	authApp, user, err := ic.getAuthorizationFromContext(r)
+	if err != nil {
+		ic.logger.Errorf("failed to get authorization: %v", err)
+		controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("invalid request: %v", err))
+		return
+	}
+	var realm *database.Realm
+	if authApp != nil {
+		realm, err = authApp.GetRealm(ic.db)
+		if err != nil {
+			controller.WriteJSON(w, http.StatusUnauthorized, api.Error(http.StatusText(http.StatusUnauthorized)))
+			return
+		}
+	} else {
+		// if it's a user logged in, we can pull realm from the context.
+		realm = controller.RealmFromContext(ctx)
+		if realm == nil {
+			controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("no realm selected"))
+			return
+		}
+	}
+
 	// Verify SMS configuration if phone was provided
 	var smsProvider sms.Provider
 	if request.Phone != "" {
-		var err error
-		smsProvider, err = ic.db.GetSMSProvider(ctx, "") // TODO(sethvargo): pass in realm
+		smsProvider, err = realm.GetSMSProvider(ctx, ic.db)
 		if err != nil {
-			ic.logger.Errorf("otp.GetSMSProvider: %v", err)
+			ic.logger.Errorf("GetSMSProvider: %v", err)
 			controller.WriteJSON(w, http.StatusInternalServerError, api.Error("failed to get sms provider"))
 			return
 		}
@@ -119,25 +140,6 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	expiration := ic.config.GetVerificationCodeDuration()
 	expiryTime := time.Now().UTC().Add(expiration)
 
-	authApp, user, err := ic.getAuthorizationFromContext(r)
-	if err != nil {
-		ic.logger.Errorf("failed to get authorization: %v", err)
-		controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("invalid request: %v", err))
-		return
-	}
-	var realmID uint
-	if authApp != nil {
-		realmID = authApp.RealmID
-	} else {
-		// if it's a user logged in, we can pull realm from the context.
-		realm := controller.RealmFromContext(ctx)
-		if realm == nil {
-			controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("no realm selected"))
-			return
-		}
-		realmID = realm.ID
-	}
-
 	// Generate verification code
 	codeRequest := otp.Request{
 		DB:            ic.db,
@@ -148,7 +150,7 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		MaxSymptomAge: ic.config.GetAllowedSymptomAge(),
 		IssuingUser:   user,
 		IssuingApp:    authApp,
-		RealmID:       realmID,
+		RealmID:       realm.ID,
 	}
 
 	code, err := codeRequest.Issue(ctx, ic.config.GetColissionRetryCount())
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -125,6 +125,18 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("invalid request: %v", err))
 		return
 	}
+	var realmID uint
+	if authApp != nil {
+		realmID = authApp.RealmID
+	} else {
+		// if it's a user logged in, we can pull realm from the context.
+		realm := controller.RealmFromContext(ctx)
+		if realm == nil {
+			controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("no realm selected"))
+			return
+		}
+		realmID = realm.ID
+	}
 
 	// Generate verification code
 	codeRequest := otp.Request{
@@ -136,6 +148,7 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		MaxSymptomAge: ic.config.GetAllowedSymptomAge(),
 		IssuingUser:   user,
 		IssuingApp:    authApp,
+		RealmID:       realmID,
 	}
 
 	code, err := codeRequest.Issue(ctx, ic.config.GetColissionRetryCount())
--- a/pkg/controller/middleware/auth.go
+++ b/pkg/controller/middleware/auth.go
@@ -32,16 +32,68 @@ import (
 )
 
 var (
-	ErrUserNotFound = errors.New("user not found")
-	ErrUserDisabled = errors.New("user disabled")
+	ErrUserNotFound  = errors.New("user not found")
+	ErrUserDisabled  = errors.New("user disabled")
+	ErrNotRealmAdmin = errors.New("not realm reamin")
 )
+
+const (
+	ContextRealm = "realm"
+)
+
+type RequireRealmAdminHandler struct {
+	logger *zap.SugaredLogger
+}
+
+// RequireRealmAdmin verifies that a user is an admin in the selected realm.
+// It must be used AFTER the RequireAuth and RequireRealm middlewares.
+func RequireRealmAdmin(ctx context.Context) *RequireRealmAdminHandler {
+	return &RequireRealmAdminHandler{logging.FromContext(ctx)}
+}
+
+func (rra *RequireRealmAdminHandler) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := func() error {
+			// Get user from context.
+			user := controller.UserFromContext(r.Context())
+			if user == nil {
+				return fmt.Errorf("missing user in session")
+			}
+
+			// Get realm from context.
+			realm := controller.RealmFromContext(r.Context())
+			if realm == nil {
+				return fmt.Errorf("missing realm in session")
+			}
+
+			if !user.CanAdminRealm(realm.ID) {
+				return ErrNotRealmAdmin
+			}
+
+			return nil
+		}(); err != nil {
+			rra.logger.Errorw("RequireRealmAdmin", "error", err)
+
+			if errors.Is(err, ErrNotRealmAdmin) {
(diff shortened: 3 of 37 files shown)


mikehelmick:
Not ready to merge yet

mikehelmick:
Still a WIP - but this is essentially working end to end.
Will round out the testing in the morning.

pkg/controller/issueapi/issue.go
```diff
@@ -127,6 +127,18 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
+	var realmID uint
+	if authApp != nil {
+		realmID = authApp.RealmID
+	} else {
+		// if it's a user logged in, we can pull realm from the context.
+		realm, err := controller.GetRealm(w, r, user)
+		if err != nil {
+			controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("no realm selected"))
```
sethvargo, line 137: How would a user specify their realm? I'm not sure it makes sense to extract the realm from the user? I actually don't understand this CUJ much.

pkg/controller/middleware/auth.go
```diff
@@ -34,16 +34,76 @@ import (
 )
 
 var (
-	ErrUserNotFound = errors.New("user not found")
-	ErrUserDisabled = errors.New("user disabled")
+	ErrUserNotFound  = errors.New("user not found")
+	ErrUserDisabled  = errors.New("user disabled")
+	ErrNotRealmAdmin = errors.New("not realm reamin")
```
sethvargo, line 39:
```suggestion
	ErrNotRealmAdmin = errors.New("not realm admin")
```

sethvargo: Or just `user not an admin`

pkg/controller/middleware/auth.go
```diff
@@ -34,16 +34,76 @@ import (
...
 )
 
+const (
+	ContextRealm = "realm"
```
sethvargo, line 43: Doc. Maybe `RealmContextKey`?

pkg/controller/middleware/auth.go
```diff
@@ -34,16 +34,76 @@ import (
...
+			if !user.CanAdminRealm(realm.ID) {
+				return ErrNotRealmAdmin
+			}
+
+			return nil
+		}(); err != nil {
+			rra.logger.Errorw("RequireRealmAdmin", "error", err)
+
+			if errors.Is(err, ErrNotRealmAdmin) {
+				flash.FromContext(w, r).Error("You are not authorized to admin the selected realm. Choose another.")
```
sethvargo, line 88:
```suggestion
				flash.FromContext(w, r).Error("You are not authorized to admin that realm.")
```

pkg/controller/middleware/auth.go
```diff
@@ -34,16 +34,76 @@ import (
...
+	})
+}
+
 type RequireAdminHandler struct {
 	logger *zap.SugaredLogger
 }
 
-// RequireAdmin verifies that a user is an admin. It must be used AFTER the
-// RequireAuth middleware.
+// RequireAdmin verifies that a user is a system admin.
```
sethvargo, line 105: What is a "system" admin?

pkg/controller/middleware/auth.go
```diff
@@ -92,6 +152,7 @@ type RequireAuthHandler struct {
 
 // RequireAuth requires a user is authenticated using firebase auth, that such a
 // user exists in the database, and that said user is not disabled.
+// This middleware also write the currently selected realm to the context.
```
sethvargo, line 155:
```suggestion
// This middleware also writes the currently selected realm to the context.
```

sethvargo, line 155: Where does it do that?

pkg/controller/middleware/realm.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/flash"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/logging"
+	httpcontext "github.com/gorilla/context"
+	"go.uber.org/zap"
+)
+
+var (
+	ErrNoRealmSelected = errors.New("no realm selected")
```
sethvargo, line 32: Recommend `ErrMissingRealm` and `"missing realm"`

pkg/controller/middleware/realm.go
```diff
@@ -0,0 +1,91 @@
...
+func RequireRealm(ctx context.Context) *RequireRealmHandler {
+	return &RequireRealmHandler{logging.FromContext(ctx)}
+}
+
+func (h *RequireRealmHandler) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := func() error {
+			cookie, err := r.Cookie("realm")
+			if err != nil {
+				return ErrNoRealmSelected
```
sethvargo, line 50: Maybe log these or wrap the error for debugging purposes?

pkg/controller/realm/cookie.go
```diff
@@ -0,0 +1,32 @@
...
+package realm
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/config"
+)
+
+func setRealmCookie(w http.ResponseWriter, c *config.ServerConfig, realmID uint) {
```
sethvargo, line 24: It's worth noting that if _any_ data is sent before the cookie, this will fail.

pkg/controller/realm/save.go
```diff
@@ -0,0 +1,80 @@
...
+func NewSaveController(ctx context.Context, config *config.ServerConfig, db *database.Database) http.Handler {
+	return &realmSaveController{config, db, logging.FromContext(ctx)}
+}
+
+func (rsc *realmSaveController) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	flash := flash.FromContext(w, r)
+
+	var form formData
+	if err := controller.BindForm(w, r, &form); err != nil {
+		rsc.logger.Errorf("invalid realm select rquest: %v", err)
```
sethvargo, line 48: It's ironic that Russ would make it into our codebase.

pkg/controller/realm/save.go
```diff
@@ -0,0 +1,80 @@
...
+		flash.Error("Invalid realm selection.")
+		http.Redirect(w, r, "/home/realm", http.StatusSeeOther)
+		return
+	}
+
+	rsc.logger.Infof("FORM: %+v", form)
```
sethvargo, line 54: Intended?

pkg/controller/realm/save.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// Verify that the user has access to the realm.
+	if user.CanViewRealm(realm.ID) {
+		setRealmCookie(w, rsc.config, realm.ID)
+		flash.Alert("Selected realm '%v'", realm.Name)
+		http.Redirect(w, r, "/home", http.StatusSeeOther)
+		return
+	}
+	// Not allowed to see the realm selected.
```
sethvargo, line 77: Newline

pkg/controller/user.go
```diff
@@ -39,3 +39,28 @@ func GetUser(w http.ResponseWriter, r *http.Request) (*database.User, error) {
 	}
 	return user, nil
 }
+
+// GetRealm returns the current realm based on the logged in user and the context.
+func GetRealm(w http.ResponseWriter, r *http.Request, u *database.User) (*database.Realm, error) {
```
sethvargo, line 44: The database.User parameter seems unused?

pkg/controller/user.go
```diff
@@ -39,3 +39,28 @@ func GetUser(w http.ResponseWriter, r *http.Request) (*database.User, error) {
...
+	rawRealm, ok := context.GetOk(r, "realm")
```
sethvargo, line 45: Also there's a constant for this, but we don't seem to be using it?

pkg/controller/user/deleteuser.go
```diff
@@ -49,6 +49,12 @@ func (udc *userDeleteController) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 	}
 	flash := flash.FromContext(w, r)
+	realm, err := controller.GetRealm(w, r, user)
+	if err != nil {
+		flash.Error("Select realm to continue.")
+		http.Redirect(w, r, "/realm", http.StatusSeeOther)
+		return
+	}
 
 	udc.logger.Infof("FLASH: %+v", flash)
```
sethvargo, line 59: Debug?

pkg/controller/user.go
sethvargo: Was the intention to verify that the user could actually access said realm?

pkg/controller/user/user.go
```diff
@@ -51,17 +51,28 @@ func (lc *userListController) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
 
 	m := html.GetTemplateMap(r)
 	m["user"] = user
+	m["realm"] = realm
 
-	users, err := lc.db.ListUsers(false)
-	if err != nil {
+	if err := realm.LoadRealmUsers(lc.db, false); err != nil {
```
sethvargo, line 65: Same comment about authorized apps - I think it's fine for the realm to cache these, but I think it would be a cleaner API if this returned users:

```go
users, err := realm.GetUsers(...)
```

sethvargo: Also a bit weird that it's called GetRealmUsers - you're already operating on a realm

pkg/controller/user/user.go
```diff
@@ -51,17 +51,28 @@ func (lc *userListController) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
 		flash.ErrorNow("Error loading users: %v", err)
 	}
 
-	m["users"] = users
+	admins := make(map[uint]bool)
+	for _, au := range realm.RealmAdmins {
```
sethvargo, line 70: Is RealmAdmins populated by LoadRealmUsers? That's not super intuitive to me... I'd rather see this as `RealmAdmins()` that's loaded and cached.

pkg/controller/verifyapi/verify.go
```diff
@@ -73,7 +80,7 @@ func (v *VerifyAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Exchange the short term verification code for a long term verification token.
 	// The token can be used to sign TEKs later.
-	verificationToken, err := v.db.VerifyCodeAndIssueToken(request.VerificationCode, v.config.VerificationTokenDuration)
+	verificationToken, err := v.db.VerifyCodeAndIssueToken(request.VerificationCode, authApp.RealmID, v.config.VerificationTokenDuration)
```
sethvargo, line 83: Feels like realm should be the first parameter?

pkg/database/authorized_app.go
```diff
@@ -77,7 +81,7 @@ func (db *Database) ListAuthorizedApps(includeDeleted bool) ([]*AuthorizedApp, e
 
 // CreateAuthorizedApp generates a new APIKey and assigns it to the specified
 // name.
-func (db *Database) CreateAuthorizedApp(name string, apiUserType APIUserType) (*AuthorizedApp, error) {
+func (db *Database) CreateAuthorizedApp(name string, apiUserType APIUserType, realm *Realm) (*AuthorizedApp, error) {
```
sethvargo, line 84: I think it makes sense to standardize that realm is the first param when a realm is required.

sethvargo: Also, why require a full Realm instead of just the ID?

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+
+	RealmUsers  []*User `gorm:"many2many:user_realms"`
+	RealmAdmins []*User `gorm:"many2many:admin_realms"`
+
+	// Relations to items that blong to a realm.
+	Codes  []*VerificationCode
+	Tokens []*Token
+}
+
+func (r *Realm) AddAuthorizedApp(a *AuthorizedApp) {
```
sethvargo, line 40: Does this save it in the database or is it only in-memory? Comments would be helpful here and throughout 😄

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+	r.AuthorizedApps = append(r.AuthorizedApps, a)
+}
+
+func (r *Realm) AddUser(u *User) {
+	for _, cUser := range r.RealmUsers {
```
sethvargo, line 45: For thousands of users, this is gonna get expensive. Can we use a map or have gorm do something smart like de-dupe at the database layer?

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+func (r *Realm) AddAdminUser(u *User) {
+	for _, cUser := range r.RealmAdmins {
+		if cUser.ID == u.ID {
+			return
+		}
+	}
+	r.RealmAdmins = append(r.RealmAdmins, u)
+}
+
+func (r *Realm) LoadRealmUsers(db *Database, includeDeleted bool) error {
```
sethvargo, line 62: What's a case where we'd include deleted? I feel like the caller should be responsible for passing in an unscoped DB instead.

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+			return fmt.Errorf("unable to remove user from realm: %w", err)
+		}
+		if err := tx.Model(r).Association("RealmAdmins").Delete(u).Error; err != nil {
+			return fmt.Errorf("unable to remove user from realm admins: %w", err)
+		}
+
+		// If the user no has no associations, the user should be deleted.
+		var user User
+		if err := tx.Preload("Realms").Preload("AdminRealms").Where("id = ?", u.ID).First(&user).Error; err != nil {
+			return err
```
sethvargo, line 97: Wrap

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		if len(user.AdminRealms) == 0 && len(user.Realms) == 0 {
+			if err := tx.Delete(user).Error; err != nil {
+				return err
```
sethvargo, line 101: Wrap

pkg/database/realm.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return &realm, nil
+}
+
+func (db *Database) SaveRealm(r *Realm) error {
+	if r.Model.ID == 0 {
+		return db.db.Create(r).Error
+	}
+	return db.db.Save(r).Error
+}
```
sethvargo, line 131: Do we need a DeleteRealm?

pkg/database/sms_config.go
```diff
@@ -35,6 +35,10 @@ type SMSConfig struct {
 	TwilioAccountSid string `gorm:"type:varchar(250)"`
 	TwilioAuthToken  string `gorm:"type:varchar(250)"` // secret reference
 	TwilioFromNumber string `gorm:"type:varchar(16)"`
+
+	// SMS Config belongs to exactly one realm.
+	RealmID uint
```
sethvargo, line 40: Can you move this to the top of the struct? Why do we need both a Realm and RealmID?

pkg/database/token.go
```diff
@@ -46,6 +46,10 @@ type Token struct {
 	SymptomDate *time.Time
 	Used        bool `gorm:"default:false"`
 	ExpiresAt   time.Time
+
+	// Tokens belong to one realm.
```
sethvargo, line 50: Top

pkg/database/token.go
```diff
@@ -105,10 +109,10 @@ func (t *Token) Subject() *Subject {
 
 // ClaimToken looks up the token by ID, verifies that it is not expired and that
 // the specified subject matches the parameters that were configured when issued.
-func (db *Database) ClaimToken(tokenID string, subject *Subject) error {
+func (db *Database) ClaimToken(tokenID string, realmID uint, subject *Subject) error {
```
sethvargo, line 112: Realm first param

pkg/database/vercode.go
```diff
@@ -52,6 +52,9 @@ type VerificationCode struct {
 	ExpiresAt   time.Time
 	IssuingUser *User
 	IssuingApp  *AuthorizedApp
+	// VerificationCodes belong to exactly one realm when issued.
```
sethvargo, line 55: Top

pkg/otp/code.go
```diff
@@ -54,6 +54,7 @@ type Request struct {
 	MaxSymptomAge time.Duration
 	IssuingUser   *database.User
 	IssuingApp    *database.AuthorizedApp
+	RealmID       uint
```
sethvargo, line 57: Top

pkg/database/vercode.go
sethvargo: Why both?

mikehelmick:
removing hold.

Still some room for improvement on the DB interface layer as it relates to multi-tenancy

cmd/add-realm/main.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"flag"
+	"log"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/sethvargo/go-envconfig/pkg/envconfig"
+)
+
+func main() {
```
sethvargo, line 27: It'd be good to use the `realMain()` pattern so we can properly cleanup things (defer's don't run with log.Fatal)

sethvargo: Can be follow-up

pkg/controller/user.go
```diff
@@ -18,18 +18,22 @@ import (
 	"context"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/logging"
 )
 
 // contextKey is a unique type to avoid clashing with other packages that use
 // context's to pass data.
-type contextKey struct{}
+type contextKey string
```
sethvargo, line 26: The context docs explicitly say to use something other than a string type. I think these need to be `type struct` instead of `var struct` - I'll clean this up later.

pkg/controller/user.go
```diff
@@ -63,6 +67,7 @@ func WithUser(ctx context.Context, u *database.User) context.Context {
 // not a user object, the result will be nil.
 func UserFromContext(ctx context.Context) *database.User {
 	v := ctx.Value(contextKeyUser)
+	logging.FromContext(ctx).Infof("GETTING USER FROM CONTEXT, %+v", v)
```
sethvargo, line 70: Remove

pkg/database/sms_config.go
```diff
@@ -35,26 +37,34 @@ type SMSConfig struct {
 	TwilioAccountSid string `gorm:"type:varchar(250)"`
 	TwilioAuthToken  string `gorm:"type:varchar(250)"` // secret reference
 	TwilioFromNumber string `gorm:"type:varchar(16)"`
+
+	twilioAuthSecret string
```
sethvargo, line 41: I don't think we need to persist this on the struct. I'll do a follow-up PR.

PR #106, opened 2020-07-10
Replace rate limiter

sethvargo:
This also changes to limit by API Key on the API server instead of by IP address. Once this is all good and verified, I'll start the process of updating Terraform to deploy a Redis instance and switch to redisstore.


cmd/apiserver/main.go
```diff
@@ -88,3 +99,18 @@ func main() {
...
 	log.Fatal(srv.ListenAndServe())
 }
+
+func apiKeyFunc() httplimit.KeyFunc {
+	ipKeyFunc := httplimit.IPKeyFunc("X-Forwarded-For")
+
+	return func(r *http.Request) (string, error) {
+		v := r.Header.Get("X-API-Key")
+		if v != "" {
+			dig := sha1.Sum([]byte(v))
```
sethvargo, line 109: This is probably overkill since it's all in-memory, but once we switch to Redis, we'll want to apply something like this in case someone gains unauthorized access to Redis.

cmd/apiserver/main.go
```diff
@@ -59,18 +60,28 @@ func main() {
...
-	}, limiter.WithTrustForwardHeader(true))
-	r.Use(limithttp.NewMiddleware(limitInstance).Handler)
+	store, err := memorystore.New(&memorystore.Config{
+		Tokens:   config.RateLimit,
+		Interval: 1 * time.Minute,
+	})
+	if err != nil {
+		log.Fatalf("failed to create limiter: %v", err)
+	}
+	defer store.Stop()
```
mikehelmick, line 72: I guess more of an upstream comment, does it make sense to have this be a Closer?
https://golang.org/pkg/io/#Closer

cmd/cleanup/main.go
```diff
@@ -52,11 +51,20 @@ func main() {
 	r := mux.NewRouter()
 
 	// Setup rate limiter
-	limitInstance := limiter.New(memory.NewStore(), limiter.Rate{
-		Period: 1 * time.Minute,
-		Limit:  int64(config.RateLimit),
-	}, limiter.WithTrustForwardHeader(true))
-	r.Use(limithttp.NewMiddleware(limitInstance).Handler)
+	store, err := memorystore.New(&memorystore.Config{
```
mikehelmick, line 54: now this is overkill :)

cmd/server/main.go
```diff
@@ -74,11 +73,21 @@ func main() {
...
+	store, err := memorystore.New(&memorystore.Config{
+		Tokens:   config.RateLimit,
+		Interval: 1 * time.Minute,
+	})
+	if err != nil {
+		log.Fatalf("failed to create limiter: %v", err)
+	}
+	defer store.Stop()
+
+	httplimiter, err := httplimit.NewMiddleware(store, httplimit.IPKeyFunc("X-Forwarded-For"))
```
mikehelmick, line 85: I think this is where, for logged in actions, we want user based limiting, can be a followup.

PR #107, opened 2020-07-10
add formal API for APIserver cover/chaff requests from devices

mikehelmick:


pkg/controller/cover/cover.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (c *CoverAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var request api.CoverRequest
+	if err := controller.BindJSON(w, r, &request); err != nil {
+		c.logger.Errorf("failed to bind request: %v", err)
+		controller.WriteJSON(w, http.StatusBadRequest, api.Error("invalid request: %v", err))
+		return
+	}
+
+	// Don't immediately return.
```
squee1945, line 85: Can you also say why?

pkg/controller/cover/cover.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		c.logger.Errorf("error getting random int, using min sleep: %v", err)
+		delay = big.NewInt(0)
+	}
+
+	length := time.Duration(minDelayMS+delay.Int64()) * time.Millisecond
+	time.Sleep(length)
+}
+
+func (c *CoverAPI) randomData() (string, error) {
```
crwilcox, line 64: Is there any way we can ensure these get configured well? If they aren't accurate to the live payload, the chaff is easily detected as chaff.

pkg/controller/cover/cover.go
mikehelmick: updated comment

crwilcox:
/lgtm /hold

Comments are all about clarification so shouldn't require further review.

pkg/controller/cover/cover.go
mikehelmick: Yes - will send a followup that tracks and normalizes both request size and latency

https://github.com/mikehelmick/go-chaff

mikehelmick: added necessary TODOs

PR #111, opened 2020-07-14
Switch to redis-based rate limiting

sethvargo:
/assign @mikehelmick

### DIFF
--- a/cmd/cleanup/main.go
+++ b/cmd/cleanup/main.go
@@ -26,8 +26,6 @@ import (
 
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/cleanup"
-	"github.com/sethvargo/go-limiter/httplimit"
-	"github.com/sethvargo/go-limiter/memorystore"
 
 	"github.com/google/exposure-notifications-server/pkg/cache"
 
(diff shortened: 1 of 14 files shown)


cmd/cleanup/main.go
```diff
@@ -50,22 +48,6 @@ func main() {
 
 	r := mux.NewRouter()
 
-	// Setup rate limiter
```
sethvargo, line 53: I took this out since you mentioned it was overkill in the last PR

terraform/services.tf
```diff
@@ -43,6 +43,16 @@ locals {
     FIREBASE_STORAGE_BUCKET    = lookup(data.google_firebase_web_app_config.default, "storage_bucket")
   }
 
+  redis_config = {
+    RATE_LIMIT_TYPE     = "REDIS"
+    RATE_LIMIT_TOKENS   = "60"
+    RATE_LIMIT_INTERVAL = "1m"
+    REDIS_HOST          = google_redis_instance.cache.host
+    REDIS_PORT          = google_redis_instance.cache.port
+    REDIS_MIN_POOL      = 32
```
sethvargo, line 52: Please verify this makes sense for Cloud Run. I load tested it and it was "fine".

mikehelmick: not confident in valid settings, seems fine to start with

Issue #95, opened 2020-07-05
Log into Server via Http


colbymorrison:
### Question
Hi, I'm volunteering over at Covid Watch. As mentioned in my colleague @ibeckermayer 's [PR](https://github.com/google/exposure-notifications-verification-server/pull/88), we have a [web app](https://github.com/covidwatchorg/permission-portal) that we'd like to request verification codes from this server. I wrote a solution for our short-term/demo use that integrates our app with the server as it exists now, using http requests. Though preferably, we'd like to request verification codes via a programmatic interface as opposed to using axios requests. In my short-term solution, we create one user in the verification server for our web app. When a user in our app requests a code, our backend (Firebase Cloud Functions) logs into the server as that user via http and gets a new verification code. Relevant code in our repo is [here](https://github.com/covidwatchorg/permission-portal/pull/307/files#diff-99a7fb22b4b966aca20c8a846a08bdf9). As far as we can tell, the username/password are acting effectively as an API key, and we can use them to generate verification codes programmatically.  We are generally curious as to what differentiates this approach from the API-key approach, particularly from a security perspective. Would our ability to get codes this way not let someone generate new verification codes relatively rapidly, in the manner @mikehelmick was concerned with in [his response](https://github.com/google/exposure-notifications-verification-server/pull/88) to our API-key based approach? Thanks so much!

mikehelmick:
/cc @ibeckermayer

Thanks - I think the use case for unattended access makes sense. It sounds like in your console, there is an authenticated and trusted human who is logged in and responsible for issuing diagnosis verification codes to end users.

Having all users from your frontend come in as a single API key. We're scheduling additional quota and abuse detection work. In that model, having your frontend communicate to our frontend API via an API key would meant that quota would get all bundled up under that one "user." 

The other thing that we're doing is expanding this implementation to be multi-tenant - allowing for multiple PHAs to be present in a single installation. 

This just means that these API keys would need to be scoped correctly (just like user accounts will need to be scoped correctly).

So, if we go this route, I think it needs to be a _different_ API key than the one we're using for exchanges.

Two questions:

1) What's the reason to not have direct case workers to the UI for this server? This seems like an unnecessary layer of indirection given that our API will constrain you to exactly the type of information we accept.

[...]

PR #112, opened 2020-07-15
Admin API Keys

mikehelmick:
* Introduce Admin API keys, that will be used for issuing diagnosis codes
* Existing keys auto migrated to be device keys
* UI for creating these

### DIFF
--- a/pkg/database/authorized_app.go
+++ b/pkg/database/authorized_app.go
@@ -24,6 +24,9 @@ import (
 
 const (
 	apiKeyBytes = 64 // 64 bytes is 86 chararacters in non-padded base64.
+
+	APIUserTypeDevice = 0
+	APIUserTypeAdmin  = 1
 )
 
 // AuthorizedApp represents an application that is authorized to verify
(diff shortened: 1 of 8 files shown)


pkg/database/authorized_app.go
```diff
@@ -29,10 +29,14 @@ const (
...
+//
+// Admin Keys are able to issue diagnosis keys and are not able to perticipate
+// the verification protocol.
 type AuthorizedApp struct {
 	gorm.Model
-	Name   string `gorm:"type:varchar(100);unique_index"`
-	APIKey string `gorm:"type:varchar(100);unique_index"`
+	Name     string `gorm:"type:varchar(100);unique_index"`
+	APIKey   string `gorm:"type:varchar(100);unique_index"`
+	AdminKey bool   `gorm:"default:false"`
```
sethvargo, line 39: Should we consider making this an enum (e.g. KeyType)? The boolean feels weird to me.

sherifkozman:
@mikehelmick will that enable external integration to use API calls to create verification codes without having to relay on the UI ?

pkg/database/authorized_app.go
mikehelmick: changed to int / enum - right now I don't see the need for more than 2 key types, but I'm sure I'll be wrong tomorrow :)

mikehelmick:
> @mikehelmick will that enable external integration to use API calls to create verification codes without having to relay on the UI ?

correct.

pkg/database/authorized_app.go
```diff
@@ -24,6 +24,9 @@ import (
 
 const (
 	apiKeyBytes = 64 // 64 bytes is 86 chararacters in non-padded base64.
+
+	APIUserTypeDevice = 0
```
sethvargo, line 28: iota?

sethvargo: and should this be a custom type?

mikehelmick: made it a custom type , plumbed all the way through

mikehelmick:
Tagging issue #95

Issue #76, opened 2020-06-24
Implement per-user quotas for issuing codes


mikehelmick:
Impose a limit on the number of codes an individual login can issue over a period of time

Provide reports to admins on the number of codes issues by each user

@crwilcox  tracking list of parts:
- [x] Rate Limit Logged In Users on Admin pages. (#113)
- [x] Add Audit Tracing to Keys, marking who generated (#118)
- [x] Add an admin page showing Verification Codes created per user. (#152)
- [ ] Add quotas per user of keys allowed to be created
- [ ] Add quotas per API Key of keys allowed to be created

crwilcox:
Work to surface usage stats is done. Rate limiting of how quickly codes can be created is also implemented. 

Potential work remains in adding daily/weekly quotas for users/api keys.

mikehelmick:
This should be per-realm, not per user

PR #113, opened 2020-07-15
feat: add rate limiting per user email for admin

crwilcox:
Towards #76.


cmd/server/main.go
```diff
@@ -112,12 +112,18 @@ func main() {
...
 
-		// API for creating new verification codes. Called via AJAX.
-		sub.Handle("/issue", issueapi.New(ctx, config, db)).Methods("POST")
-
 		// API for obtaining a CSRF token before calling /issue
 		// Installed in this context, it requires authentication.
 		sub.Handle("/csrf", csrfctl.NewCSRFAPI()).Methods("GET")
+
+		issueSub := sub.PathPrefix("/issue").Subrouter()
+		// TODO(crwilcox): add these to the config.
```
crwilcox, line 120: wanted to get agreement on approach. Can expand this pr to config once we reach concensus.

crwilcox:
@sethvargo I like using a cache for this, but not sure if that works if we also want to track keys issued? Though it would stop the hug of death that will come if someone tries to mass issue since they will get quota'd early on.

sethvargo:
> track keys issued

What kind of tracking are you thinking?

crwilcox:
Part of the original ask was to know how many keys a given account has issued.

pkg/database/migrations.go
```diff
@@ -155,6 +155,31 @@ func (db *Database) RunMigrations(ctx context.Context) error {
 				return nil
 			},
 		},
+		{
+			ID: "00008-AddUserIssuanceQuota",
```
mikehelmick, line 159: needs to be 00009, merged in after my change

pkg/controller/middleware/auth.go
```diff
@@ -165,3 +165,90 @@ func (rah *RequireAuthHandler) Handle(next http.Handler) http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := func() error {
+			// Get the cookie
+			cookie, err := r.Cookie("session")
+			if err != nil {
+				return fmt.Errorf("failed to get cookie: %w", err)
+			}
+
+			// Verify cookie
+			token, err := rah.client.VerifySessionCookie(rah.ctx, cookie.Value)
```
mikehelmick, line 196: hmm, we end up doing this in multiple places - and this also ties it to a user.

Would it be better to have our higher middlewares put the quota identifer into the context and have this enforce quota aginst that.

Then in the DB we can have a quota object that can be tied to an API key or a User

Requires middleware installed int he right order.

sethvargo: That's kinda how the rate limiter works - you define the key function based on "something" in the http.Request object.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/jinzhu/gorm v1.9.12
 	github.com/mitchellh/mapstructure v1.3.2 // indirect
 	github.com/ory/dockertest v3.3.5+incompatible
+	github.com/pkg/errors v0.9.1
```
sethvargo: Any idea why this came in?

pkg/controller/middleware/auth.go
```diff
@@ -166,89 +169,31 @@ func (rah *RequireAuthHandler) Handle(next http.Handler) http.Handler {
...
-
-			if controller.IsJSONContentType(r) {
-				controller.WriteJSON(w, http.StatusTooManyRequests, nil)
-			} else {
-				flash.FromContext(w, r).Error("Quota Exceeded or Unauthorized")
-				http.Redirect(w, r, "/signout", http.StatusFound)
-			}
-		} else {
-			next.ServeHTTP(w, r)
+func userIdKeyFunc() httplimit.KeyFunc {
```
sethvargo, line 182: So, this only limits if a user exists, and if anything fails in that chain, the request is 429-ed.

I'd recommend something like:

- try to get user
- if no user or type assertion fails, rate limit on IP
- rate limit on email

pkg/controller/middleware/auth.go
```diff
@@ -166,89 +169,31 @@ func (rah *RequireAuthHandler) Handle(next http.Handler) http.Handler {
...
+			return "", fmt.Errorf("unauthorized")
 		}
-	})
+		user, ok := rawUser.(*database.User)
+		if !ok {
+			// flash.FromContext(w, r).Error("internal error - you have been logged out.")
+			// http.Redirect(w, r, "/signout", http.StatusFound)
+			return "", fmt.Errorf("internal error")
+		}
+		return user.Email, nil
```
sethvargo, line 192: I recommend we hash this (sha256 is probably fine). Otherwise we risk putting PII in redis.

pkg/controller/middleware/auth.go
```diff
@@ -166,89 +169,31 @@ func (rah *RequireAuthHandler) Handle(next http.Handler) http.Handler {
...
 }
 
-type QuotaKeyIssuanceHandler struct {
-	ctx            context.Context
-	client         *auth.Client
-	db             *database.Database
-	quotaTTL       time.Duration // frequency of quota reset
-	requestsPerTTL int           // count of requests over the ttl window.
-	logger         *zap.SugaredLogger
+// UserRequestLimiter requires a user is authenticated using firebase auth,
```
sethvargo, line 172: Do you plan on sharing this with other packages? If not, I'd recommend moving it inline into main.go.

pkg/controller/middleware/auth.go
crwilcox: I can do this, though today, all of the places that this is used will be behind the auth middleware. But I can add a fall back.

pkg/controller/middleware/auth.go
crwilcox: I moved this closer to where it is used. We can move it back up if needed.

cmd/server/main.go
```diff
@@ -141,3 +145,21 @@ func main() {
 	log.Printf("Listening on :%v", config.Port)
 	log.Fatal(srv.ListenAndServe())
 }
+
+func userEmailKeyFunc() httplimit.KeyFunc {
+	return func(r *http.Request) (string, error) {
+		ipKeyFunc := httplimit.IPKeyFunc("X-Forwarded-For")
```
sethvargo, line 151: This is a nitty performance thing. We should move this outside of the return so it's only allocated once. Here it will allocate on each request.

PR #118, opened 2020-07-17
Adds Tracking Information to Codes to allow auditing

crwilcox:
Towards #76

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -101,15 +101,8 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	expiryTime := time.Now().Add(ic.config.GetVerificationCodeDuration())
 
-	var issuingUserID uint
-	var issuingAuthorizedAppID uint
-
 	authApp, user := ic.getAuthorizationFromContext(r)
-	if authApp != nil {
-		issuingAuthorizedAppID = authApp.ID
-	} else if user != nil {
-		issuingUserID = user.ID
-	} else {
+	if authApp == nil && user == nil {
 		errStr := "unable to identify authorized requestor"
 		ic.logger.Errorf(errStr)
 		controller.WriteJSON(w, http.StatusOK, api.Error(errStr))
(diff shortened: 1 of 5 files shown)


pkg/database/vercode.go
```diff
@@ -45,11 +45,13 @@ var (
...
-	Claimed     bool   `gorm:"default:false"`
-	TestType    string `gorm:"type:varchar(20)"`
-	SymptomDate *time.Time
-	ExpiresAt   time.Time
+	Code                   string `gorm:"type:varchar(20);unique_index"`
+	Claimed                bool   `gorm:"default:false"`
+	TestType               string `gorm:"type:varchar(20)"`
+	SymptomDate            *time.Time
+	ExpiresAt              time.Time
+	IssuingUserID          uint
```
mikehelmick, line 53: w/ Gorm - I think this can be 

```
IssuingUser *User
IssuingApp *AuthorizedApp
```

and I think that simplifies the rest of the code a bit as well.

pkg/controller/issueapi/issue.go
```diff
@@ -99,14 +101,58 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			controller.WriteJSON(w, http.StatusOK, api.Error("error reading user from request context"))
+			return
+		}
+		user, ok := rawUser.(*database.User)
+		if !ok {
+			ic.logger.Errorf("error reading user from request context")
+			controller.WriteJSON(w, http.StatusOK, api.Error("error reading user from request context"))
+			return
+		}
+		got, err := ic.db.FindUser(user.Email)
```
mikehelmick, line 137: I don't think there's a reason to re-find these- as they would have been just loaded/autorized.

pkg/controller/issueapi/issue.go
```diff
@@ -99,14 +101,58 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	expiryTime := time.Now().Add(ic.config.GetVerificationCodeDuration())
 
+	var issuingUserID uint
+	var issuingAuthorizedAppID uint
+
+	// Attempt to find the authorized app.
+	rawAuthorizedApp, ok := httpcontext.GetOk(r, "authorizedApp")
```
mikehelmick, line 108: I would move this out to a utility method - I Think it may end up be using later

authUser, authApp, err := getAuthorizationFromContext()

pkg/controller/issueapi/issue.go
crwilcox: I agree this is a bit pedantic. i can simplify.

PR #120, opened 2020-07-18
Gracefully stop and cleanup servers

sethvargo:
We can merge as-is, or wait until https://github.com/google/exposure-notifications-server/pull/742 and then I'll use that to remove some of the duplication around listening for cleanup signals.


sethvargo:
This is important for the rate limiter and the database since they both use connection pools. It's especially important for the rate limiter because redis makes heavy use of keep-alives.

PR #125, opened 2020-07-19
Add SMS interfaces and Twilio implementation

sethvargo:
This doesn't add any envconfig because we're going to want this to be multitenant and stored in the database.

@mikehelmick do we want to provision a Twilio account so we can inject the test credentials into the build for CI?

### DIFF
--- a/pkg/sms/noop.go
+++ b/pkg/sms/noop.go
@@ -1,0 +1,34 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


sethvargo:
```text
twilio_test.go:23: missing TWILIO_ACCOUNT_SID
```

Expected until we figure out how we want to inject creds for tests.

sherifkozman:
@sethvargo @mikehelmick we have that available if you need help using a test Twilio account for CI and testing

mikehelmick:
> @sethvargo @mikehelmick we have that available if you need help using a test Twilio account for CI and testing

we'll get a google owned account for this. Thanks for the offer.

sethvargo:
@chaodaiG it doesn't look like the envvar is being set still?

chaodaiG:
> @chaodaiG it doesn't look like the envvar is being set still?

They should have been added here: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/614949c999d54867dcc9117474b0f0aceb6e79b6/prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml#L1

sethvargo:
@chaodaiG it worked 🥳

PR #130, opened 2020-07-21
Add more SMS functionality

sethvargo:
/assign @mikehelmick

### DIFF
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -110,6 +110,7 @@
        url: '/home/issue',
        type : "POST",
        dataType : 'json',
+       cache: false,
        contentType: 'application/json',
        data : JSON.stringify(data),
        headers: {"X-CSRF-Token": csrfToken},
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -60,6 +60,21 @@
         </div>
       </div>
 
+      {{ if .smsEnabled }}
+      <div class="row form-group">
+        <label for="symptomDate" class="col-sm-3">Patient phone</label>
+        <div class="col-sm-9">
+          <div class="datepicker date input-group p-0 shadow-sm">
+            <input type="tel" id="phone" name="phone" class="form-control" data-inputmask="'mask': '+33 9 99 99 99 99'" />
+          </div>
+          <small class="form-text text-muted">
+            This must be a phone number capable of receiving SMS text messages.
+            The number should be formatted with the country code.
+          </small>
+        </div>
+      </div>
+      {{ end }}
+
       <div class="row form-group">
         <div class="offset-sm-3 col-sm-9">
           <button id="submit" class="btn btn-primary">Generate code</button>
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -128,10 +128,13 @@
          }
        },
        error: function(xhr, resp, text) {
-         $("#codeerror").text("error: " + text);
+         var message = resp;
+         if (xhr && xhr.responseJSON && xhr.responseJSON.error) {
+           message = message + ": " + xhr.responseJSON.error;
+         }
+         $("#codeerror").text(message);
          $("#codearea").hide();
          $("#codeerror").show();
-         console.log(xhr, resp, text);
        }
      });
    }
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -155,6 +155,7 @@
           url: '/home/csrf',
           type: "GET",
           dataType: 'json',
+          cache: false,
           contentType: 'application/json',
           success: function(result) {
             if (result.error != "") {
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -135,6 +135,7 @@
          $("#codeerror").text(message);
          $("#codearea").hide();
          $("#codeerror").show();
+         $("#submit").prop('disabled', false);
        }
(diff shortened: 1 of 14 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -84,27 +99,28 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if request.SymptomDate != "" {
 		if parsed, err := time.Parse("2006-01-02", request.SymptomDate); err != nil {
 			ic.logger.Errorf("time.Parse: %v", err)
-			controller.WriteJSON(w, http.StatusOK, api.Error("invalid test date: %v", err))
+			controller.WriteJSON(w, http.StatusUnprocessableEntity, api.Error("invalid test date: %v", err))
```
mikehelmick, line 102: change error text to symptom onset date - I missed that when we changed the date field

pkg/controller/issueapi/issue.go
```diff
@@ -123,10 +139,19 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		controller.WriteJSON(w, http.StatusOK, api.Error("error generating verification, wait a moment and try again"))
+		controller.WriteJSON(w, http.StatusInternalServerError, api.Error("error generating verification, wait a moment and try again"))
 		return
 	}
 
+	if smsProvider != nil {
+		message := fmt.Sprintf(smsTemplate, code, int(expiration.Minutes()))
+		if err := smsProvider.SendSMS(ctx, request.Phone, message); err != nil {
+			ic.logger.Errorf("otp.SendSMS: %v", err)
+			controller.WriteJSON(w, http.StatusInternalServerError, api.Error("failed to send sms: %v", err))
```
mikehelmick, line 150: Do we want to delete the code in that case?

pkg/controller/issueapi/issue.go
```diff
@@ -156,5 +181,8 @@ func (ic *IssueAPI) getAuthorizationFromContext(r *http.Request) (*database.Auth
 		return nil, nil, fmt.Errorf("unable to identify authorized requestor")
 	}
 	return nil, user, nil
-
 }
+
+const smsTemplate = `Your exposure notifications verification code is %s. ` +
```
mikehelmick, line 186: this is fine for now, we'll probably need PHA customization here

and we'll need a URI to the app at some point.

pkg/controller/issueapi/issue.go
```diff
@@ -68,11 +68,26 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Verify SMS configuration
+	smsProvider, err := ic.db.GetSMSProvider("") // TODO(sethvargo): pass in realm
```
mikehelmick, line 72: would be good to put a write through cache in front of these (can be future PR)

pkg/controller/issueapi/issue.go
```diff
@@ -68,11 +68,26 @@ func (ic *IssueAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		ic.logger.Errorf("otp.GetSMSProvider: %v", err)
+		controller.WriteJSON(w, http.StatusInternalServerError, api.Error("failed to get sms provider"))
+		return
+	}
+
+	// Phone number is required if SMS is configured
+	if smsProvider != nil && request.Phone == "" {
```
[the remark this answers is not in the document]
sethvargo: Done.

pkg/controller/issueapi/issue.go
sethvargo: I think we want to save the code and present it on the screen for the clinician to read. WDYT?

pkg/controller/issueapi/issue.go
sethvargo: Yea I figured we'd iterate on it.

cmd/server/assets/home.html
```diff
@@ -60,6 +60,21 @@ <h1>Issue Diagnosis Verification Code</h1>
...
 
+      {{ if .smsEnabled }}
+      <div class="row form-group">
+        <label for="symptomDate" class="col-sm-3">Patient phone</label>
+        <div class="col-sm-9">
+          <div class="datepicker date input-group p-0 shadow-sm">
+            <input type="tel" id="phone" name="phone" class="form-control" />
+          </div>
+          <small class="form-text text-muted">
+            This must be a phone number capable of receiving SMS text messages.
```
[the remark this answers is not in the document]
sethvargo: So, Twilio is "smart" in that by default it uses the country code associated with the account if one isn't provided. I'm thinking we leave this as-is for now and maybe iterate on it later? I originally went down a path of some client-side javascript stuff, but the twilio api error responses are pretty good.

pkg/controller/issueapi/issue.go
sethvargo: Done

pkg/controller/issueapi/issue.go
sethvargo: Done

pkg/controller/issueapi/issue.go
sethvargo: Done

mikehelmick: yeah, I think it's fine, but this only sends the error back

Issue #108, opened 2020-07-13
Improve formation of chaff requests


mikehelmick:
Use https://github.com/mikehelmick/go-chaff to track real responses and return heuristically valid responses.

mikehelmick:
Also - we should remove the /api/cover path and have this handled by the main API calls.

PR #136, opened 2020-07-27
Use chaff for handling fake requests

sethvargo:
Fixes GH-108


cmd/apiserver/main.go
```diff
@@ -129,3 +135,7 @@ func apiKeyFunc() httplimit.KeyFunc {
 		return ipKeyFunc(r)
 	}
 }
+
+func handleChaff(tracker *chaff.Tracker, next http.Handler) http.Handler {
```
sethvargo, line 139: I'm not a huge fan of one-line helpers, but the handler becomes really difficult to follow without it.

pkg/api/api.go
```diff
@@ -96,16 +96,3 @@ type VerificationCertificateResponse struct {
 	Certificate string `json:"certificate"`
 	Error       string `json:"error"`
 }
-
-// CoverRequest is a client->server request for simulating traffic to the server.
```
sethvargo, line 100: Do we need to keep this in and mark as deprecated instead?

README.md
```diff
@@ -130,58 +130,57 @@ keys and API keys have one of two levels of access: `DEVICE` or `ADMIN`.
...
-  ```
+The following APIs exist for the API server (`cmd/apiserver`). All APIs are JSON
+over HTTPS, only use `POST`, and require that the API key be passed in the HTTP
+header `X-API-Key`.
+
+In addition to "real" requests, the server also accept chaff (fake) requests.
+These can be used to obfuscate real traffic from a network observer or server
+operator. To initiate a chaff request, set the `X-Chaff` header on your request.
+The server will respond with a fake response that your client **MUST NOT**
+process. Client's should sporadically issue chaff requests.
```
mikehelmick, line 141: also - note that the client should send a request body that is of the appropriate size, but that that request body will not be processed at all.

README.md
```diff
@@ -130,58 +130,57 @@ keys and API keys have one of two levels of access: `DEVICE` or `ADMIN`.
...
+
+    **VerifyCodeRequest:**
+
+    ```json
+    {
+      "code": "<the code>"
+    }
+    ```
+
+    **VerifyCodeResponse:**
```
mikehelmick, line 154: let's add a padding field here for the chaff requests, so the client knows to ignore it.

pkg/api/api.go
mikehelmick: as far as I know , this isn't in use any anyone yet.

README.md
sethvargo: Wouldn't the client have known it sent a chaff request? I think I'm missing why we'd need something in the response.

README.md
sethvargo: Done

README.md
mikehelmick: I mean, one would hope. We do need to pad out the responses so they're the right size, but it's also find to document that they're not JSON responses. Either way.

PR #140, opened 2020-07-28
Stop using gorilla/context

sethvargo:
It's unmaintained and leaks memory


go.sum
```diff
@@ -9,45 +9,34 @@ cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxK
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
```
sethvargo, line 9: I'm not gonna pretend to understand why removing context created such a large diff on the sum

pkg/controller/user/deleteuser.go
```diff
@@ -43,22 +41,12 @@ func NewDeleteController(ctx context.Context, config *config.ServerConfig, db *d
 func (udc *userDeleteController) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer http.Redirect(w, r, "/users", http.StatusSeeOther)
 
-	user, err := controller.GetUser(w, r)
```
sethvargo, line 46: This was never used

pkg/controller/user/saveuser.go
```diff
@@ -47,38 +46,26 @@ func NewSaveController(ctx context.Context, config *config.ServerConfig, db *dat
 }
 
 func (usc *userSaveController) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	defer http.Redirect(w, r, "/users", http.StatusSeeOther)
-
-	user, err := controller.GetUser(w, r)
```
sethvargo, line 52: This was never used

pkg/controller/user/saveuser.go
```diff
@@ -47,38 +46,26 @@ func NewSaveController(ctx context.Context, config *config.ServerConfig, db *dat
...
-	usc.logger.Infof("FLASH: %+v", flash)
-
-	m := html.GetTemplateMap(r)
-	m["user"] = user
-
 	var form formData
 	if err := controller.BindForm(w, r, &form); err != nil {
-		usc.logger.Errorf("error parsing form: %v", err)
 		flash.Error("Failed to process form: %v", err)
-		controller.WriteJSON(w, http.StatusBadRequest, nil)
```
sethvargo, line 68: Unclear why we were returning JSON?

mikehelmick: yeah - redirect is right here, it's a human used page.

PR #142, opened 2020-07-29
Followup PR comments

sethvargo:
Followup to GH-105

### DIFF
--- a/pkg/controller/user.go
+++ b/pkg/controller/user.go
@@ -18,7 +18,6 @@ import (
 	"context"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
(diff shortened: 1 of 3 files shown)


pkg/controller/user.go
```diff
@@ -67,7 +66,6 @@ func WithUser(ctx context.Context, u *database.User) context.Context {
 // not a user object, the result will be nil.
 func UserFromContext(ctx context.Context) *database.User {
 	v := ctx.Value(contextKeyUser)
-	logging.FromContext(ctx).Infof("GETTING USER FROM CONTEXT, %+v", v)
```
mikehelmick, line 70: oops - that's how I discovered the context keys were colliding

PR #148, opened 2020-07-29
update docs to mention realms, add-users script fixes

crwilcox:


README.md
```diff
@@ -108,8 +108,14 @@ export DEV_MODE=1
 # Migrate DB
 ./scripts/dev dbmigrate
 
+
+# create a realm and note the return number
+# successfully created realm default (1)
+go run ./cmd/add-realm --name "default"
```
mikehelmick, line 114: so dbmigrate should create an initial realm called "Default"  with ID of 1

I think this command can be dropped in favor of a comment that says as much

cmd/add-users/main.go
```diff
@@ -33,8 +33,13 @@ func main() {
 	disabledFlag := flag.Bool("disabled", false, "true if user should be disabled")
 	realmID := flag.Int64("realm", -1, "realm to add the user to")
 	realmAdminID := flag.Int64("admin-realm", -1, "realm to add the user to")
```
mikehelmick, line 35: let's make this a bool flag

Issue #143, opened 2020-07-29
Realm admin / config


mikehelmick:
- Implement UI for creating / updating realms
- Implement realm specific settings UI

Necessary for customization - and settings for the realm.
Needed for #138 #117 #134 

/assign

mikehelmick:
The test types to report should be configurable.

With a default of confirmed results only.

sethvargo:
Realm-specific settings UI is complete

PR #150, opened 2020-07-29
Add basic realm admin page.

mikehelmick:
only lets you change the realm name
next will be SMS settings things

Towards #143


sethvargo:
Should this be /realm/settings instead?

mikehelmick:
> Should this be /realm/settings instead?

updated - wasn't sure how that would collide w/ the /realm subroute, but it's all good.

PR #151, opened 2020-07-30
Unify HTML and JSON rendering logic

sethvargo:
There's a fun edge case where template/json rendering can fail partially through. When that happens, the experience is quite awful. This PR unifies common logic into a shared renderer capable of rendering HTML and JSON performantly and as safely as possible. 

Part of this involved plumbing the renderer through a lot of stuff. In doing so, it became clear that we should have one controller per _package_, not per _file_, so this unifies controllers up to their package levels. Since each file is no longer its own `http.Handler`, there are functions on the controller that return `http.Handler`s when called. This will make things easier to test in the future. There's a lot of whitespace changes, so [`?w=1`](https://github.com/google/exposure-notifications-verification-server/pull/151?w=1) is likely a friend here (sorry).

I tried to use `git mv`, but the diff is still larger than I'd like 😦 .


cmd/adminapi/main.go
```diff
@@ -87,15 +88,22 @@ func realMain(ctx context.Context) error {
 	}
 	r.Use(httplimiter.Handle)
 
+	// Create the renderer
+	h, err := render.New(ctx, "", config.DevMode)
```
sethvargo, line 92: DevMode exists on all the configs now. If enabled on an API server, it will print debug messages if JSON marshalling or rendering fails.

pkg/controller/apikey/create.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/flash"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+)
+
+func (c *Controller) HandleCreate() http.Handler {
+	type FormData struct {
+		Name string               `form:"name,required"`
```
sethvargo, line 27: These values are now marked as required to prevent blank API keys

pkg/controller/user/create.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/flash"
+)
+
+func (c *Controller) HandleCreate() http.Handler {
+	type FormData struct {
+		Email    string `form:"email,required"`
+		Name     string `form:"name,required"`
+		Admin    bool   `form:"admin"`
+		Disabled bool   `form:"disabled"`
```
mikehelmick, line 29: I dropped this from the page as it didn't make sense to create a user in disabled state.

PR #152, opened 2020-07-30
Add counts of codes generated to users and api key pages

crwilcox:


crwilcox:
image

sethvargo:
This LGTM sans 2 things:

1. `COUNT` queries are actually really expensive because they require a full table lock. It's probably fine for now, but it could become a source of contention if there's a lot of viewing of the admin page.

1. I'd like to hold off on merging this until GH-151 since the rebase of this will be far easier than the rebase of that 😄 

/hold

crwilcox:
@sethvargo sounds good to me. I can hold this np. It should be easy-ish to merge conflicts. 

Great point on counts though. being accurate to the 1 isn't actually that important for this anyway. The goal is to detect someone that is an order of magnitude differen, not 4 vs 5 :)

mikehelmick:
you can rebase now

pkg/controller/apikey/apikey.go
```diff
@@ -65,7 +65,19 @@ func (lc *apikeyListController) ServeHTTP(w http.ResponseWriter, r *http.Request
 	if _, err := realm.GetAuthorizedApps(lc.db, true); err != nil {
 		flash.ErrorNow("Error loading API Keys: %v", err)
 	}
+
+	creationCounts := make(map[uint]int64)
+	for _, app := range realm.AuthorizedApps {
+		count, err := lc.db.CountVerificationCodesByAuthorizedApp(app.ID)
```
mikehelmick, line 71: I don't think we want these to be realtime counts and it limits our count to essentially what's live.

I would like to see this more like a separate table that increments these counts on a per day, per user|app, per realm basis

that way we can track and show history overtime and not have it lost when primary data is cleaned up.

sethvargo: and that also solves the performance concern I highlighted above 😄

pkg/controller/apikey/index.go
```diff
@@ -48,9 +48,21 @@ func (c *Controller) HandleIndex() http.Handler {
 		}
 
 		m := controller.TemplateMapFromContext(ctx)
+
+		creationCounts := make(map[uint]int64)
+		for _, app := range realm.AuthorizedApps {
+			count, err := c.db.CountVerificationCodesByAuthorizedApp(app.ID)
+			if err != nil {
+				flash.Error("Error loading app code creation counts: %v", err)
```
odeke-em, line 56: Do we mean to continue after this error or is the fall through to normal case intentional?

pkg/controller/user/index.go
```diff
@@ -49,11 +49,22 @@ func (c *Controller) HandleIndex() http.Handler {
 			admins[au.ID] = true
 		}
 
+		creationCounts := make(map[uint]int64)
+		for _, user := range realm.RealmUsers {
+			count, err := c.db.CountVerificationCodesByUser(user.ID)
+			if err != nil {
+				flash.Error("Error loading user code creation counts: %v", err)
```
odeke-em, line 56: Ditto here about intentional vs accidental fall through.

pkg/controller/apikey/index.go
crwilcox: removed these. I had them while dev-ing since I was getting intermittent errors

pkg/controller/user/index.go
crwilcox: removed these. I had them while dev-ing since I was getting intermittent errors\

crwilcox:
FYI: some todos in this, but they could be done as separate work items
1) Count is in queries, removing that may be good for avoiding table locks
2) There is no cron setup yet, right now the summaries are re-calculated on page load. it is in a method though so if we have a spot to run scheduled jobs these could be moved

pkg/controller/apikey/apikey.go
crwilcox: Should be resolved now. Count isn't used, also results are stored from previous runs. Still don't have this in a cloud scheduler run, but starting in on that now.

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+// AuthorizedAppStats represents statistics related to a user in the database.
+type AuthorizedAppStats struct {
+	gorm.Model
+	Date            time.Time `gorm:"unique_index:idx_date_app_realm"`
+	AuthorizedAppID int64     `gorm:"unique_index:idx_date_app_realm"`
+	AuthorizedApp   AuthorizedApp
+	RealmID         int64 `gorm:"unique_index:idx_date_app_realm"`
+	Realm           Realm
```
mikehelmick, line 30: this is only necessary if we're going to follow the relation link from the app stats record

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,193 @@
...
+
+	"github.com/jinzhu/gorm"
+)
+
+// AuthorizedAppStats represents statistics related to a user in the database.
+type AuthorizedAppStats struct {
+	gorm.Model
+	Date            time.Time `gorm:"unique_index:idx_date_app_realm"`
+	AuthorizedAppID int64     `gorm:"unique_index:idx_date_app_realm"`
+	AuthorizedApp   AuthorizedApp
```
mikehelmick, line 28: only necessary if we're going to follow the relation from this record

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// TableName sets the AuthorizedAppStats table name
+func (AuthorizedAppStats) TableName() string {
+	return "authorized_app_stats"
+}
+
+func (db *Database) GetAuthorizedAppStatsSummary(a *AuthorizedApp, r *Realm) (*AuthorizedAppStatsSummary, error) {
+	t := time.Now().UTC()
+	roundedTime := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
```
mikehelmick, line 49: t.Truncate(24 * time.Hour)

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,193 @@
...
+
+	var summary = &AuthorizedAppStatsSummary{}
+	var dailyStats []*AuthorizedAppStats
+
+	// get the last 30 days of dates and counts for a given user in a realm.
+	if err := db.db.Preload("AuthorizedApp").Preload("Realm").
```
mikehelmick, line 55: I don't think you need to preload the AuthorizedApp and Realm here - 

since they're parameters to this function - seems like those are unnecessary joins

pkg/database/user_stats.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+// UserStats represents statistics related to a user in the database.
+type UserStats struct {
+	gorm.Model
+	Date        time.Time `gorm:"unique_index:idx_date_user_realm"`
+	UserID      int64     `gorm:"unique_index:idx_date_user_realm"`
+	User        User
+	RealmID     int64 `gorm:"unique_index:idx_date_user_realm"`
+	Realm       Realm
```
mikehelmick, line 30: same - I think we just need the RealmID for tracking but don't need to traverse the models from the stats level.

pkg/database/user_stats.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (db *Database) GetUserStatsSummary(u *User, r *Realm) (*UserStatsSummary, error) {
+	t := time.Now().UTC()
+	roundedTime := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
+
+	var summary = &UserStatsSummary{}
+	var dailyStats []*UserStats
+
+	// get the last 30 days of dates and counts for a given user in a realm.
+	if err := db.db.Preload("User").Preload("Realm").
```
mikehelmick, line 55: no preload realm

pkg/database/user_stats.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	// create 24h, 7d, 30d counts
+	return summary, nil
+}
+
+func (db *Database) GetUserStats(u *User, r *Realm, t time.Time) (*UserStats, error) {
+	var userStats UserStats
+	roundedTime := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
```
sethvargo, line 85: Truncate instead

pkg/controller/user/index.go
```diff
@@ -27,6 +27,9 @@ func (c *Controller) HandleIndex() http.Handler {
 		ctx := r.Context()
 		flash := flash.FromContext(w, r)
 
+		// TODO(crwilcox): this should be moved to a cron run.
+		c.db.UpdateUserStats()
```
sethvargo, line 31: Do you plan to address the TODO in this PR or a follow up? If follow up, can we wrap this in a query param for now?

pkg/database/authorized_app_stats.go
Goalbok: Please I want to build  a trivia application and I need your help

Goalbok: This is my email goalbokrewards@gmail.com

sethvargo: Deleted two comments asking for help building an unrelated app.

pkg/database/authorized_app_stats.go
crwilcox: without both fields, no fields were being placed as columns by gorm. Only appid is stored in gorm. But the model has authorizedapp. I can just track id. is that what you mean?

pkg/controller/user/index.go
crwilcox: @sethvargo was split on this. From our daily sync I am going to work in the background cron.

pkg/database/authorized_app_stats.go
crwilcox: removed object relations from app stats and user stats.

pkg/controller/cleanup/cleanup.go
```diff
@@ -113,6 +114,19 @@ func (c *Controller) HandleCleanup() http.Handler {
 			c.logger.Infof("purged %v verification tokens", count)
 		}
 
+		// Gather usage statistics
+		if err := c.db.UpdateUserStats(); err != nil {
```
crwilcox, line 118: I added these as part of cleanup. They are quick runs and not having it on all user page loads should avoid db contention as things scale.

This could be made less frequent, but updating the issued (1d) sort of needs frequent enough updates to be useful.

cmd/server/assets/apikeys.html
```diff
@@ -52,30 +57,36 @@ <h2>API keys</h2>
...
+            <th scope="col">Issued (1d)</th>
+            <th scope="col">Issued (7d)</th>
+            <th scope="col">Issued (30d)</th>
           </tr>
         </thead>
         <tbody>
-        {{range .apps}}
+          {{range .apps}}
           <tr>
-            <td>{{.Name}}</td>
```
sethvargo, line 60: I think some of this got lost in a rebase?

cmd/server/assets/apikeys.html
```diff
@@ -52,30 +57,36 @@ <h2>API keys</h2>
...
+              {{else}}<span class="badge badge-pill badge-success">enabled</span>{{end}}
             </td>
-            <td width="50">
-              {{if .IsAdminType}}<span class="badge badge-pill badge-primary" data-toggle="tooltip" data-placement="top" title="Can be used to issue verification codes">Admin</span>{{end}}
-              {{if .IsDeviceType}}<span class="badge badge-pill badge-secondary" data-toggle="tooltip" data-placement="top" title="For use in mobile apps to verify codes and get certificates">Device</span>{{end}}
+            <td scope="row">
+              {{index $codesGenerated1d .ID}}
             </td>
-            <td width="50">
-              {{if .DeletedAt}}
```
sethvargo, line 71: Also looks like it was lost in a rebase

cmd/server/assets/users.html
```diff
@@ -88,6 +95,15 @@ <h2>Users</h2>
                 </form>
               {{end}}
             </td>
+            <td scope="row">
+              {{index $codesGenerated1d .ID}}
```
sethvargo, line 99: Is there a reason you can't do `$codesGenerated1d.ID`?

pkg/controller/apikey/index.go
```diff
@@ -38,6 +38,23 @@ func (c *Controller) HandleIndex() http.Handler {
...
 
 		m := controller.TemplateMapFromContext(ctx)
+
+		creationCounts1d := make(map[uint]uint64)
+		creationCounts7d := make(map[uint]uint64)
+		creationCounts30d := make(map[uint]uint64)
+		for _, app := range realm.AuthorizedApps {
+			appStatsSummary, err := c.db.GetAuthorizedAppStatsSummary(app, realm)
+			if err != nil {
+				controller.InternalError(w, r, c.h, err)
```
sethvargo, line 48: Need a `return` after this

pkg/controller/user/index.go
```diff
@@ -36,8 +36,26 @@ func (c *Controller) HandleIndex() http.Handler {
 			return
 		}
 
+		creationCounts1d := make(map[uint]uint64)
+		creationCounts7d := make(map[uint]uint64)
+		creationCounts30d := make(map[uint]uint64)
+		for _, user := range realm.RealmUsers {
+			userStatsSummary, err := c.db.GetUserStatsSummary(user, realm)
+			if err != nil {
+				controller.InternalError(w, r, c.h, err)
```
sethvargo, line 45: Return

pkg/controller/user/index.go
```diff
@@ -36,8 +36,26 @@ func (c *Controller) HandleIndex() http.Handler {
 			return
 		}
 
+		creationCounts1d := make(map[uint]uint64)
```
sethvargo, line 39: This is identical to the other ones? Maybe we should make a helper or separate package?

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"time"
+
+	"github.com/jinzhu/gorm"
+)
+
+// AuthorizedAppStats represents statistics related to a user in the database.
+type AuthorizedAppStats struct {
+	gorm.Model
+	Date            time.Time `gorm:"unique_index:idx_date_app_realm"`
```
sethvargo, line 26: I can't find any docs on what this actually does? Is it valid to have multiple fields against the same index?

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,191 @@
...
+	AuthorizedAppID uint      `gorm:"unique_index:idx_date_app_realm"`
+	RealmID         uint      `gorm:"unique_index:idx_date_app_realm"`
+	CodesIssued     uint64
+}
+
+type AuthorizedAppStatsSummary struct {
+	AuthorizedApp  AuthorizedApp
+	Realm          Realm
```
sethvargo, line 34: AA and Realm should probably be pointers

pkg/database/authorized_app_stats.go
```diff
@@ -0,0 +1,191 @@
...
+	return nil
+}
+
+func (db *Database) countVerificationCodesByAuthorizedApp(realmID uint, appID uint, t time.Time) (uint64, error) {
+	if appID <= 0 {
+		return 0, nil
+	}
+
+	var rows []*VerificationCode
+	if err := db.db.Model(&VerificationCode{}).
```
sethvargo, line 183: Using `.Count` would push this into the SQL layer and would probably be faster (since we wouldn't have to allocate all the rows just to count them)

pkg/database/user_stats.go
```diff
@@ -0,0 +1,191 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: Same comments here as on apikey_stats

pkg/database/authorized_app_stats.go
crwilcox: Count would be faster but count requires locking, which was a previous concern raised, I can go back to that certainly.

cmd/server/assets/users.html
crwilcox: This is the id of a user being used as an index into the map.

pkg/database/authorized_app_stats.go
crwilcox: It's a composite unique index to ensure each entry per user/realm/date is unique, since this entry is really a wrapup for the day. I am trying to find docs now, but it is that with a specified name for the index.

cmd/server/assets/apikeys.html
crwilcox: It isn't lost so much as reformatted. Though I see there are some other changes (`scope="row" `seems to have been dropped)

cmd/server/assets/apikeys.html
crwilcox: Should be resolved now. Most of the difference was my editor changing line breaks. But there were some erroneous 'width' and 'type=row' bits that seem to be off from a few merges.

pkg/controller/user/index.go
crwilcox: opened issue to track this since merged before resolved fully: https://github.com/google/exposure-notifications-verification-server/issues/199

pkg/database/authorized_app_stats.go
crwilcox: https://github.com/google/exposure-notifications-verification-server/issues/200 to track

PR #156, opened 2020-07-30
Switch to gorilla sessions and encrypted cookies

sethvargo:
This PR does a few things:

- Adds a way to process base64 envvars as []byte
- Switches to gorilla/sessions and encrypted cookies
- Switches to gorilla/sessions for realm selection
- Fixes some bugs with realm selection when a user is a member of multiple realms
- Pass flash to the template

Known issues:

- If you change the cookie encryption token, everything breaks and you have to clear cookies to continue. I can create a follow-up issue to track this.
- I didn't modify our flash implementation since, I'll do that in a follow-up since this is already getting large.


cmd/server/main.go
```diff
@@ -69,6 +70,14 @@ func realMain(ctx context.Context) error {
 		return fmt.Errorf("failed to process config: %w", err)
 	}
 
+	// Setup sessions
+	sessions := sessions.NewCookieStore(config.CookieKeys.AsBytes()...)
```
sethvargo, line 74: These are the same options as before

pkg/config/server_config.go
```diff
@@ -74,17 +77,6 @@ func NewServerConfig(ctx context.Context) (*ServerConfig, error) {
 	return &config, nil
 }
 
-func (c *ServerConfig) CSRFKey() ([]byte, error) {
```
sethvargo, line 77: Yay - envconfig will do this now

pkg/controller/middleware/auth.go
```diff
@@ -77,84 +82,41 @@ func (h *RequireRealmAdminHandler) Handle(next http.Handler) http.Handler {
...
 
 			flash.FromContext(w, r).Error("Unauthorized")
 			http.Redirect(w, r, "/signout", http.StatusFound)
 			return
 		}
-		next.ServeHTTP(w, r)
-	})
-}
-
-type RequireAdminHandler struct {
```
sethvargo, line 94: This was replaced with realmadmin

pkg/controller/middleware/auth.go
```diff
@@ -165,22 +127,32 @@ func (h *RequireAuthHandler) Handle(next http.Handler) http.Handler {
...
-			cookie, err := r.Cookie("session")
+			// Get the session
+			session, err := h.sessions.Get(r, "session")
 			if err != nil {
-				return fmt.Errorf("failed to get cookie: %w", err)
+				return fmt.Errorf("failed to get session: %w", err)
+			}
+
+			// Verify firebase cookie value is present
+			if session.Values == nil || session.Values["firebaseCookie"] == nil {
```
sethvargo, line 137: This is a somewhat annoying side effect of gorilla sessions. Values can be nil and Values is a `map[interface{}][interface{}]`, so the key can also be nil.

pkg/controller/realm/index.go
```diff
@@ -60,6 +78,7 @@ func (c *Controller) HandleIndex() http.Handler {
 
 		// User must select their realm.
 		m := controller.TemplateMapFromContext(ctx)
+		m["flash"] = flash
```
sethvargo, line 81: Lost many hours to this.

pkg/controller/realm/select.go
```diff
@@ -23,44 +23,60 @@ import (
...
+		}
+
 		user := controller.UserFromContext(ctx)
 		if user == nil {
 			flash.Error("Unauthorized.")
 			http.Redirect(w, r, "/signout", http.StatusSeeOther)
 			return
 		}
 
-		realm := controller.RealmFromContext(ctx)
```
sethvargo, line 40: I'm not entirely sure how this ever worked before? This is the page that sets the realm's context, but it required a context to be set.

cmd/server/main.go
```diff
@@ -69,6 +70,14 @@ func realMain(ctx context.Context) error {
...
+	sessions.Options.Path = "/"
+	sessions.Options.Domain = "" // TODO: make configurable
```
mikehelmick, line 76: will be interesting to configure in a cloud run + custom domains scenario.

cmd/server/main.go
```diff
@@ -162,31 +175,37 @@ func realMain(ctx context.Context) error {
...
+		sub.Use(requireAuth.Handle)
+		sub.Use(requireRealm.Handle)
+		sub.Use(requireAdmin.Handle)
 
 		apikeyController := apikey.New(ctx, config, db, h)
 		sub.Handle("", apikeyController.HandleIndex()).Methods("GET")
 		sub.Handle("/create", apikeyController.HandleCreate()).Methods("POST")
+	}
 
+	// userrs
```
mikehelmick, line 190: users

pkg/config/server_config.go
```diff
@@ -41,10 +40,14 @@ type ServerConfig struct {
...
 	RevokeCheckPeriod     time.Duration `env:"REVOKE_CHECK_DURATION,default=5m"`
 
-	// CSRF Secret Key. Must be 32-bytes. Can be generated with tools/gen-secret
-	// Use the syntax of secret:// to pull the secret from secret manager.
-	// We assume the secret itself is base64 encoded. Use CSRFKey() to transform to bytes.
-	CSRFAuthKey string `env:"CSRF_AUTH_KEY,required"`
+	// CookieKeys is a slice of bytes. The odd values are hash keys to HMAC the
+	// cookies. The even values are block keys to encrypt the cookie. Both keys
+	// should be 64 bytes. The value's should be specified as base64 encoded.
+	CookieKeys Base64ByteSlice `env:"COOKIE_KEYS,required"`
```
mikehelmick, line 46: nice 👍

pkg/controller/middleware/auth.go
mikehelmick: We still need the require admin handler for the ability to create/delete realms - can I haz back?

pkg/controller/realm/select.go
mikehelmick: RealmFromContext returned nil if one wasn't set - so this worked fine

mikehelmick: ah - no this got changed

I had is reading the cookie directly.

cmd/server/main.go
sethvargo: It has to match what the user hits in the browser.

sethvargo:
Fixed typo and rebased

pkg/controller/middleware/auth.go
sethvargo: Since you asked nicely...

pkg/config/server_config.go
sethvargo: We have to pull in the changes after https://github.com/google/exposure-notifications-server/pull/788#event-3608455621 for it to work with secrets thought. It'll work with local keys just fine.

sethvargo: https://github.com/google/exposure-notifications-verification-server/pull/159

cmd/server/main.go
sethvargo: Made configurable

PR #158, opened 2020-07-31
pkg/database: alleviate User CRUD load by a distributed cache

### LINKED ISSUE #30: Cache database user reads
https://github.com/google/exposure-notifications-verification-server/pull/27#discussion_r441908612

odeke-em:
Adds Redis as a pass-through cache for the methods of *Database
for CRUD for users. Write-throughs are performed on Create and Update,
Read-through for Read, and Deletion for Delete.

This change has an escape hatch for disjoint consistency with
that database in that, if a value isn't updated, its value will
expire in 1 day (86,400 seconds)

Also added tests to verify the pass-through by intercepting each
command sent in.

Fixes #30

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -76,6 +76,8 @@ func realMain(ctx context.Context) error {
 	}
 	defer db.Close()
 
+	db.SetRedisPool(config.RedisPool)
+
 	// Setup firebase
 	app, err := firebase.NewApp(ctx, config.FirebaseConfig())
 	if err != nil {
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -18,13 +18,16 @@ package database
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/google/exposure-notifications-server/pkg/cache"
 	"github.com/google/exposure-notifications-server/pkg/secrets"
 	"github.com/jinzhu/gorm"
 
-	// ensure the postgres dialiect is compiled in.
+	// ensure the postgres dialect is compiled in.
 	_ "github.com/jinzhu/gorm/dialects/postgres"
+
+	"github.com/opencensus-integrations/redigo/redis"
 )
 
 // Database is a handle to the database layer for the Exposure Notifications
@@ -38,6 +41,10 @@ type Database struct {
 
 	// secretManager is used to resolve secrets.
 	secretManager secrets.SecretManager
+
+	// Redis configuration.
+	redisMu   sync.RWMutex
+	redisPool *redis.Pool
 }
 
 // Open created a DB connection through gorm.
@@ -65,6 +72,12 @@ func (c *Config) Open(ctx context.Context) (*Database, error) {
 		cacher:        cacher,
 		secretManager: secretManager,
 	}, nil
+}
+
+func (db *Database) SetRedisPool(redisPool *redis.Pool) {
+	db.redisMu.Lock()
+	db.redisPool = redisPool
+	db.redisMu.Unlock()
 }
 
 // Close will close the database connection. Should be deferred right after Open.
--- a/pkg/database/redis_cache_test.go
+++ b/pkg/database/redis_cache_test.go
@@ -1,0 +1,326 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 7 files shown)


cmd/server/main.go
```diff
@@ -77,6 +77,8 @@ func realMain(ctx context.Context) error {
 	}
 	defer db.Close()
 
+        db.SetRedisPool(config.RedisPool)
```
mikehelmick, line 80: run presubmits - formatting is off.

odeke-em: Done, thanks!

pkg/database/database.go
```diff
@@ -18,13 +18,16 @@ package database
...
 
 	"github.com/google/exposure-notifications-server/pkg/cache"
 	"github.com/google/exposure-notifications-server/pkg/secrets"
 	"github.com/jinzhu/gorm"
 
-	// ensure the postgres dialiect is compiled in.
+	// ensure the postgres dialect is compiled in.
 	_ "github.com/jinzhu/gorm/dialects/postgres"
+
+	"github.com/opencensus-integrations/redigo/redis"
```
sethvargo, line 30: Nit: move this above with the rest

pkg/database/database.go
```diff
@@ -38,6 +41,10 @@ type Database struct {
 
 	// secretManager is used to resolve secrets.
 	secretManager secrets.SecretManager
+
+	// Redis configuration.
+	redisMu   sync.RWMutex
+	redisPool *redis.Pool
```
sethvargo, line 47: This is a little bit weird now because we already have a cache that's in-memory (see cache.Cache above), and now we have a remote cache via Redis...

pkg/database/redis_cache_test.go
```diff
@@ -0,0 +1,326 @@
...
+type redisSequence struct {
+	Kind    string
+	Command string
+	Args    []interface{}
+}
+
+func (rcic *redisCommandInterceptorConn) SendContext(ctx context.Context, commandName string, args ...interface{}) error {
+	rcic.mu.Lock()
+	rcic.seqs = append(rcic.seqs, &redisSequence{"SEND", commandName, args})
+	rcic.mu.Unlock()
```
MushuEE, line 45: I am new to Go, so no action needed: Is there a preference between this and using `defer rcic.mu.Unlock()` after the `Lock()`? I see that pattern often but have not developed a heuristic around when to use it. Thanks!

pkg/database/user.go
```diff
@@ -132,11 +144,19 @@ func (db *Database) CreateUser(email string, name string, admin bool, disabled b
 		return nil, err
 	}
 
+	// After saving the user, now cache it.
```
MushuEE, line 147: Is this a TODO or now captured within SaveUser?

pkg/database/user.go
```diff
@@ -156,6 +179,115 @@ func (db *Database) PurgeUsers(maxAge time.Duration) (int64, error) {
...
+	if err != nil {
+		return err
+	}
+
+	conn.SendContext(ctx, "MULTI")
+	conn.SendContext(ctx, "SET", _USERS_TABLE_BY_EMAIL, user.Email, jsonBlob)
+	conn.SendContext(ctx, "EXPIRE", _USERS_TABLE_BY_EMAIL, user.Email, _CACHE_PURGE_TTL_1DAY.Seconds())
+	conn.SendContext(ctx, "SET", _USERS_TABLE_BY_ID, user.ID, jsonBlob)
+	conn.SendContext(ctx, "EXPIRE", _USERS_TABLE_BY_ID, user.ID, _CACHE_PURGE_TTL_1DAY.Seconds())
+	// Use a Time to live (TTL) of 1days so that for example if succeeded but clearing it from the
```
MushuEE, line 214: NIT: `Use a Time to live (TTL) of 24 hours so that, for example, if <save?> succeeded but clearing it from the Redis cache failed, the cached information will at least be purged if not used within within 24 hours.`

pkg/database/redis_cache_test.go
sethvargo: defer is slower (it pushes a function onto the stack). The performance has gotten a lot better in recent releases, but sometimes when you really care about performance, it's best to do this.

sethvargo:
I'm actually gonna close this out. We have a new caching system in place that we should use instead.

Issue #160, opened 2020-08-01
upon upgrade server run services fails due to missing COOKIE_KEYS variable


sherifkozman:
The latest update requires an env COOKIE_KEYS to run the server service but its not created through terraform
Please advice @sethvargo

mikehelmick:
fix is in #163 

The workaround you can do now, is to use the gen-secret tool in the repo. Run that tool twice to generate 2 secrets

```
exposure-notifications-verification-server on 🌱 issue160 [$] via 🐹 v1.14.2 took 12s 
❯ go run ./tools/gen-secret --bytes=32 -id=session-cookie-a -project=PROJECT_ID
2020/07/31 20:26:04 Created secret: projects/PROJECT_NUMBER/secrets/session-cookie-a
2020/07/31 20:26:04 Creating secret version.
2020/07/31 20:26:04 Added secret version: projects/PROJECT_NUMBER/secrets/session-cookie-a/versions/1

exposure-notifications-verification-server on 🌱 issue160 [$] via 🐹 v1.14.2 took 4s 
❯ go run ./tools/gen-secret --bytes=32 -id=session-cookie-b -project=PROJECT_ID
2020/07/31 20:26:15 Created secret: projects/PROJECT_NUMBER/secrets/session-cookie-b
2020/07/31 20:26:15 Creating secret version.
2020/07/31 20:26:15 Added secret version: projects/PROJECT_NUMBER/secrets/session-cookie-b/versions/1
```

and then on the 'server' service, set the `COOKIE_KEYS` env var to

```
secret://projects/PROJECT_NUMBER/secrets/session-cookie-a/versions/1,secret://projects/PROJECT_NUMBER/secrets/session-cookie-b/versions/1
```

Replacing PROJECT_NUMBER w/ your project number.

PR #163, opened 2020-08-01
Issue160

mikehelmick:
Fixes GH-160


terraform/service_apiserver.tf
```diff
@@ -44,18 +44,6 @@ resource "google_secret_manager_secret_iam_member" "apiserver-db" {
   member    = "serviceAccount:${google_service_account.apiserver.email}"
 }
 
-resource "google_secret_manager_secret_iam_member" "apiserver-twilio" {
```
sethvargo, line 47: I believe we still need this. Otherwise the server can't actually access the secret.

terraform/services.tf
```diff
@@ -58,3 +62,7 @@ locals {
     TOKEN_SIGNING_KEY       = trimprefix(data.google_kms_crypto_key_version.token-signer-version.id, "//cloudkms.googleapis.com/v1/")
   }
 }
+
+output "COOKIE_KEYS" {
+  value = local.session_config
+}
```
sethvargo, line 68: Nit: newline

terraform/sessions.tf
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "random_id" "cookie-key-even" {
+  byte_length = 32
+}
+
+resource "google_secret_manager_secret" "cookie-key-even" {
+  provider  = google-beta
+  secret_id = "cookie-key-even"
```
sethvargo, line 21: It's actually HMAC (odd), Encryption (even)

terraform/sessions.tf
```diff
@@ -0,0 +1,59 @@
...
+  depends_on = [
+    google_project_service.services["secretmanager.googleapis.com"],
+  ]
+}
+
+resource "google_secret_manager_secret_version" "cookie-key-odd-version" {
+  provider    = google-beta
+  secret      = google_secret_manager_secret.cookie-key-odd.id
+  secret_data = random_id.cookie-key-odd.b64_std
+}
```
sethvargo, line 59: Nit: newline

terraform/sms.tf
```diff
@@ -1,31 +0,0 @@
-# Copyright 2020 Google LLC
```
sethvargo, line 1: We still need this. It's creating an _empty_ secret and granting the required IAM permissions.

terraform/service_apiserver.tf
mikehelmick: as discussed - this is now per-realm, so we'll use envelope encryption and store creds in DB

terraform/sessions.tf
mikehelmick: renamed - check ordering on services.sf

terraform/services.tf
```diff
@@ -21,6 +21,10 @@ locals {
     CSRF_AUTH_KEY = "secret://${google_secret_manager_secret_version.csrf-token-version.id}"
   }
 
+  session_config = {
+    COOKIE_KEYS = "secret://${google_secret_manager_secret_version.cookie-encryption-key-version.id},secret://${google_secret_manager_secret_version.cookie-hmac-key-version.id}"
```
sethvargo, line 25: Hash is first, encryption is second: https://github.com/gorilla/securecookie#examples

mikehelmick: typical off by one error., fixed

PR #169, opened 2020-08-01
Support custom domain mappings

sethvargo:
Fixes GH-166

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -107,21 +107,21 @@ variable "adminapi_custom_domain" {
   type    = string
   default = ""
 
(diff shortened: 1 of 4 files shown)


terraform/variables.tf
```diff
@@ -103,6 +103,27 @@ variable "redis_cache_size" {
   description = "Size of the Redis instance in GB."
 }
 
+variable "adminapi_custom_domain" {
+  type    = string
+  default = ""
+
+  description = "Custom domain to map for adminapi. This domain must already be verified by Google. If not provided, no domain mapping is created."
```
mikehelmick, line 110: We should also state that a DNS CNAME record pointing to `ghs.googlehosted.com.` should be created in advance.

sethvargo:
@mikehelmick P(re)TAL

PR #171, opened 2020-08-04
Switch to gorilla sessions and flash

sethvargo:
Individual commits show more details, but

- Fixes a security bug (we were using text/template instead of html/template) which makes XSS very real.
- Adds a generic 500 fallback page for when things go very wrong.
- Centralizes common logic (e.g. forced signout, forced realm selection) into controller package.
- Middlewares automatically inject `flash`, `csrfToken`, and `csrfTemplate`, `currentUser`, and `currentRealm`.
- There's exactly one cookie now (down from 3!)

/assign @mikehelmick


pkg/controller/apikey/index.go
```diff
@@ -18,43 +18,29 @@ import (
...
 		}
 
 		m := controller.TemplateMapFromContext(ctx)
-		m["user"] = user
-		m["realm"] = realm
 		m["apps"] = realm.AuthorizedApps
-		m["flash"] = flash
 		m["typeAdmin"] = database.APIUserTypeAdmin
 		m["typeDevice"] = database.APIUserTypeDevice
-		m[csrf.TemplateTag] = csrf.TemplateField(r)
```
sethvargo, line 57: This is a good example of the enhancement - all these fields are automagically set by the middlewares now.

pkg/controller/flash/flash.go
```diff
@@ -12,197 +12,85 @@
 // See the License for the specific language governing permissions and
```
sethvargo, line 12: Oh wow, this diff is gross. They are totally different implementations.

pkg/controller/flash/flash.go
```diff
@@ -12,197 +12,85 @@
...
-	b, err := json.Marshal(m)
-	if err != nil {
-		// This should never happen because JSON only fails to marshal when the
-		// values are unmarshallable (func, chan), or with circular references,
-		// neither of which can happen here.
-		log.Printf("failed to marshal as json: %v", err)
+// get returns the messages in the key, clearing the values stored at the key.
+func (f *Flash) get(key flashKey) []string {
+	if v, ok := f.values[key]; ok {
+		delete(f.values, key)
```
sethvargo, line 91: Main difference in implementation: new flashes persist **until read** instead of **until the next request**.

pkg/controller/index/index.go
```diff
@@ -48,13 +46,9 @@ func New(ctx context.Context, config *config.ServerConfig, h *render.Renderer) *
 func (c *Controller) HandleIndex() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-		token := csrf.Token(r)
 
 		m := controller.TemplateMapFromContext(ctx)
 		m["firebase"] = c.config.Firebase
-		m["csrftoken"] = token
-
-		w.Header().Add("X-CSRF-Token", token)
```
sethvargo, line 57: This doesn't seem to be used?

pkg/render/render.go
```diff
@@ -19,8 +19,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"html/template"
```
sethvargo, line 22: Fun bug

cmd/server/main.go
```diff
@@ -124,44 +126,44 @@ func realMain(ctx context.Context) error {
 	// Install the CSRF protection middleware.
 	csrfController := csrfctl.New(ctx, h)
 	// TODO(mikehelmick) - there are many more configuration options for CSRF protection.
-	r.Use(csrf.Protect(config.CSRFAuthKey,
+	r.Use(middleware.ConfigureCSRF(ctx, config.CSRFAuthKey,
```
icco, line 129: If we're using our own middleware wrapper here, why not set defaults so you only have to pass in config and csrfController?

cmd/server/main.go
```diff
@@ -124,44 +126,44 @@ func realMain(ctx context.Context) error {
...
-	r.Use(middleware.FlashHandler)
+	// Sessions
+	requireSession := middleware.RequireSession(ctx, sessions, h)
+	r.Use(requireSession)
 
 	// Create common middleware
-	requireAuth := middleware.RequireAuth(ctx, auth, db, h, sessions, config.SessionDuration)
-	requireAdmin := middleware.RequireRealmAdmin(ctx, h, sessions)
-	requireRealm := middleware.RequireRealm(ctx, h, sessions)
+	requireAuth := middleware.RequireAuth(ctx, auth, db, h, config.SessionDuration)
```
icco, line 140: Looking below, it looks like all sub controllers requireAuth, should you just install it on the root router?

pkg/controller/apikey/create.go
```diff
@@ -54,12 +51,13 @@ func (c *Controller) HandleCreate() http.Handler {
...
 
 		if _, err := c.db.CreateAuthorizedApp(realm.ID, form.Name, form.Type); err != nil {
+			c.logger.Errorw("failed to create authorized app", "error", err)
 			flash.Error("Failed to create API key: %v", err)
 			http.Redirect(w, r, "/apikeys", http.StatusSeeOther)
 			return
 		}
 
-		flash.Alert("Created API Key for %q", form.Name)
+		flash.Alert("Successfully created API Key for %v", form.Name)
```
icco, line 60: Why v? Quoted seems like it would be better?

pkg/controller/context.go
```diff
@@ -18,65 +18,81 @@ import (
...
 	}
-	return authorizedApp
+	return t
 }
 
-// WithRealm sets the realm in the cotnext.
-func WithRealm(ctx context.Context, r *database.Realm) context.Context {
-	return context.WithValue(ctx, contextKeyRealm, r)
+// WithRealm stores the current realm on the context.
+func WithRealm(ctx context.Context, app *database.Realm) context.Context {
```
icco, line 57: The variable r makes more sense here instead of app?

pkg/controller/context.go
```diff
@@ -99,26 +115,26 @@ func TemplateMapFromContext(ctx context.Context) TemplateMap {
...
 	}
+
 	return m
 }
 
-// WithUser sets the user in the context.
-func WithUser(ctx context.Context, u *database.User) context.Context {
-	return context.WithValue(ctx, contextKeyUser, u)
+// WithUser stores the current user on the context.
+func WithUser(ctx context.Context, app *database.User) context.Context {
```
icco, line 123: The variable u makes more sense here than app?

pkg/render/render.go
icco: It would be a pain, but switch to https://pkg.go.dev/github.com/google/safehtml?tab=doc ?

sethvargo: Hmm - that seems to be more of an HTML builder-ing library.

cmd/server/main.go
sethvargo: index and session controllers don't require auth

pkg/controller/apikey/create.go
sethvargo: Just trying to be consistent with our other flash messages. We can change them all to be q or v, but I figured the diff was already pretty big 😄

pkg/controller/context.go
sethvargo: You may have caught me copy-pasting 😄

pkg/controller/context.go
sethvargo: You may have caught me copy-pasting 😄

cmd/server/main.go
sethvargo: Done

pkg/controller/apikey/create.go
icco: Ah fair.

pkg/render/render.go
icco: So it's the same API as html/template, but it's to prevent XSS (https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/42934.pdf has a bunch of examples). https://docs.google.com/document/d/1oNlv8Itpr_J59JNKJ0Jd4U93MJmZvXHrB0p8izQdGsM/edit also gives a bunch of examples on security vulns it protects against. 

I think it's probably too much work for now, but might be worth a TODO.

pkg/controller/apikey/index.go
mikehelmick: yay

pkg/controller/flash/flash.go
```diff
@@ -12,197 +12,85 @@
...
-// TODO(mikehelmick||sethvargo): Now that we're on gorilla, we could move to build in session / flash support.
+// Package flash implements flash messages.
 package flash
 
 import (
-	"context"
-	"encoding/base64"
-	"encoding/json"
-	"errors"
+	"encoding/gob"
```
mikehelmick, line 19: https://www.youtube.com/watch?v=SNgftTeHRz4

pkg/controller/index/index.go
mikehelmick: should be safe to remove

pkg/controller/middleware/sessions.go
```diff
@@ -0,0 +1,124 @@
...
+				return err
+			}
+
+			bfbw := &beforeFirstByteWriter{
+				w:      w,
+				before: saveSession,
+				logger: logger,
+			}
+
+			logger.Debugw("done")
```
mikehelmick, line 85: fine to have debug log statement, but "done" probably isn't helpful on it's own

sethvargo: It won't be "done", it'll be `middleware.RequireSession: done`. Named loggers are pretty nice: https://github.com/google/exposure-notifications-verification-server/pull/171/files#diff-0ad52782f7ddd2e9fd3f45181ca4eb7fR39

Issue #153, opened 2020-07-30
Remove disabled user on ceate


mikehelmick:
No reason to create a user in a disabled state.

mikehelmick:
Already removed from HTML template, needs to be removed from controller

PR #172, opened 2020-08-04
Drop disabled users

sethvargo:
We can use soft deletes with gorm to "delete" users. Fixes GH-153.

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -343,7 +343,18 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 			},
 		},
 		{
-			ID: "00014-DropUserDisabled",
+			ID: "00014-DropUserPurgeIndex",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Infof("db migrations: dropping user purge index=")
+				sql := "DROP INDEX IF EXISTS users_purge_index"
+				return tx.Exec(sql).Error
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return tx.Model(&User{}).AddIndex("users_purge_index", "updated_at").Error
+			},
+		},
+		{
+			ID: "00015-DropUserDisabled",
 			Migrate: func(tx *gorm.DB) error {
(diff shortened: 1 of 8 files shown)


pkg/database/migrations.go
```diff
@@ -342,6 +342,17 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
 			},
 		},
+		{
+			ID: "00014-DropUserDisabled",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Infof("db migrations: dropping user disabled column")
+				return tx.Model(&User{}).DropColumn("disabled").Error
+			},
+			Rollback: func(tx *gorm.DB) error {
+				sql := "ALTER TABLE users ADD COLUMN disabled bool NOT NULL DEFAULT true"
```
sethvargo, line 352: I could make this more complete, keying off of deleted_at, but it becomes a much more expensive operation.

pkg/database/migrations.go
```diff
@@ -342,6 +342,18 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00014-DropUserDisabled",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Infof("db migrations: dropping user disabled column")
+				sql := "ALTER TABLE users DROP COLUMN IF EXISTS disabled"
```
mikehelmick, line 349: Q: does this drop the column from the users_purge_index?

also, perhaps we don't need that index at all and should drop it here, if it exists.

pkg/database/migrations.go
mikehelmick: this is fine.

pkg/database/migrations.go
sethvargo: > When you remove a column from a table, PostgreSQL will automatically remove all of the indexes and constraints that involved the dropped column.

Seems so. Added a step to remove the index.

sethvargo:
@mikehelmick PT(another)L

PR #173, opened 2020-08-04
Docker updates

icco:
Updates our dockerfile

 - Move to Alpine - current image has less vulnerabilities than ubuntu. Less of an issue since we aren't running on ubuntu, but #shrugs
 - Add a docker ignore to prevent adding files we don't want in our images. Less of an issue because we are using seperate run container, but a best practice.
 - Split out env layers for easier reading
 - Add goproxy env arg to speed build time by grabbing from secure proxy server.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+scripts
+tools
+terraform
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,16 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.14 AS builder
+FROM golang:1.14-alpine AS builder
 
 ARG SERVICE
 
-RUN apt-get -qq update && apt-get -yqq install upx
+# git needed for building Go
+# upx for optimizing binary
+# binutils for strip command
+RUN apk add --no-cache git upx binutils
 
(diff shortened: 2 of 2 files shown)


sethvargo:
> Move to Alpine - current image has less vulnerabilities than ubuntu. Less of an issue since we aren't running on ubuntu, but #shrugs

We're using multi-stage builds. So while the container is built with golang:1.14, the resulting container uses scratch and copies over the compiled binary.

icco:
> > Move to Alpine - current image has less vulnerabilities than ubuntu. Less of an issue since we aren't running on ubuntu, but #shrugs
> 
> We're using multi-stage builds. So while the container is built with golang:1.14, the resulting container uses scratch and copies over the compiled binary.

Yeah, that's what I meant by "not running on ubuntu". But as there is more default software in the ubuntu image, the list of possible CVEs is greater, and could be triggered by the build process. It is a stretch, but considering how easy it is to switch over, no real downsides.

sethvargo:
Yea, I guess I was just bitten early on? Like go-migrate will totally work and compile on alpine, then fail at runtime inside scratch. I know the same is true for staticcheck.

icco:
Interesting, I haven't run into that. But good to know. Will make sure we're doing runtime checks.

PR #174, opened 2020-08-04
Initial cloud build for auto building of all images

icco:
First step of cloud build automation is to set up trigger on commits to `main` to autobuild to gcr. Decided to go this route instead of wrapping the existing build script because I didn't want to build an image with gcloud, git and bash all in it. Downsides to this are these builds happen in serial, but not a big deal.


mikehelmick:
/approve

forgot to turn both keys

Issue #157, opened 2020-07-31
Make issue UI responsive / scaled for mobile phone


mikehelmick:
/kind enhancement

PR #178, opened 2020-08-05
Mobile-friendly changes

sethvargo:
- Standardize headers and capitalization
- Standardize form layouts
- New mobile-friendly UI for codes
- Scroll to code when generated
- New magic `data-` attributes to turn links into form submit buttons
- Automatically redirect to login page after 3s of logging out
- Drop's CSRF serving - it's not necessary and dropping it simplifies getting codes

Fixes GH-157


pkg/database/realm.go
```diff
@@ -113,10 +113,10 @@ func (r *Realm) LoadRealmUsers(db *Database, includeDeleted bool) error {
 	if includeDeleted {
 		scope = db.db.Unscoped()
 	}
-	if err := scope.Model(r).Related(&r.RealmUsers, "RealmUsers").Error; err != nil {
+	if err := scope.Model(r).Preload("Realms").Preload("AdminRealms").Order("email").Related(&r.RealmUsers, "RealmUsers").Error; err != nil {
```
sethvargo, line 116: Apparently the `preload` struct tag doesn't apply on nested fields, so we have to explicitly declare it.

pkg/database/database.go
```diff
@@ -59,6 +59,12 @@ func (c *Config) Open(ctx context.Context) (*Database, error) {
 	if err != nil {
 		return nil, fmt.Errorf("database gorm.Open: %w", err)
 	}
+
+	// Log SQL statements in debug mode.
+	if c.Debug {
```
sethvargo, line 64: This was _very_ helpful in debugging

pkg/controller/session/delete.go
```diff
@@ -30,9 +30,6 @@ func (c *Controller) HandleDelete() http.Handler {
 			return
 		}
 
-		flash := controller.Flash(session)
```
sethvargo, line 33: The page already has this

sethvargo:
localhost_8080_(iPhone 6_7_8)
localhost_8080_apikeys(iPhone 6_7_8) (1)
localhost_8080_apikeys(iPhone 6_7_8)
localhost_8080_home(iPhone 6_7_8) (1)
localhost_8080_home(iPhone 6_7_8)
localhost_8080_realm_settings(iPhone 6_7_8)
localhost_8080_realm(iPhone 6_7_8)
localhost_8080_users(iPhone 6_7_8)

pkg/database/database.go
mikehelmick: nice

pkg/database/realm.go
mikehelmick: not following, why does this preload AdminRealms?

sethvargo: It's preloading the user's adminrealms (which is what we need to do in order for `CanAdminReam` to work. I don't know _why_ it works, but I've verified the SQL is correct

Issue #180, opened 2020-08-05
Add Metrics and Tracing


icco:
### TL;DR

We want end to end tracing and metric support to match what's in https://github.com/google/exposure-notifications-server

sethvargo:
Can we try to do this with middleware as much as possible? There will be exceptions, but I'm not a huge fan of how much non-core logic is in our handlers for the main server.

icco:
yeah, definitely. Anything generic (ie not business specific metrics) will be in middleware.

icco:
https://github.com/google/exposure-notifications-server/pull/846 also adds some stuff

mikehelmick:
What's left here?

icco:
I think we can close this for more specific bugs. https://github.com/google/exposure-notifications-verification-server/issues/266 and https://github.com/google/exposure-notifications-verification-server/issues/267

PR #183, opened 2020-08-05
Add Observability configs to Verification servers

icco:
First step on #180

**Release Note**

```release-note
Adds default request tracing and http metrics to all four servers.
```


cloudbuild.yaml
```diff
@@ -1,6 +1,8 @@
+timeout: 40m
```
sethvargo, line 1: Hmm - these should build in < 1min usually...

sethvargo: I recommend using `waitFor: ['-']` so they go in parallel

pkg/observability/config.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: We might want to consider making the package in the main server public instead of duplicating this.

cloudbuild.yaml
icco: They do, but when serially, they take about 11m, so they cancel because the default timeout is 10m. Didn't know about waitFor! Adding.

pkg/observability/config.go
icco: Yeah, going to do in a followup PR. Added a TODO

mikehelmick:
seems like a good start - do we want to undraft and check this in?

icco:
> seems like a good start - do we want to undraft and check this in?

Yup! It's good to go. Just wanted to test all of the servers, and it looks to be working.

PR #184, opened 2020-08-05
Releasing

mikehelmick:
## Proposed Changes

* Formalize release process

**Release Note**

```release-note
Formalize release process what was used to bootstrap v0.2.0
```


RELEASING.md
```diff
@@ -0,0 +1,47 @@
+# Exposure Notifications Verificion Servers Release Process
```
icco, line 1: Instead of duplicating the docs, should we just link to one so we don't have config drift?

mikehelmick: updated

Issue #165, opened 2020-08-01
Switch to KMS for SMS configs


sethvargo:
Have a main KMS key configured for the server for ALE, use AAD with per-record information.

Issue #117, opened 2020-07-17
SMS sending of verification codes


mikehelmick:
### TL;DR

Add the ability to send verification codes over SMS

https://www.twilio.com/messaging as likely baked in provider. Must be configurable so operator can bring their own credentials.

/kind enhancement
/priority important-soon

mikehelmick:
One other note - best if the phone numbers don't get stored in the database.

sethvargo:
I have some experience with the Twilio API, so I can take this

sherifkozman:
Twilio has a some considerations to be noted if it is used for an impelemntation that is not in the US.
i.e : Creating a Numeric or Alpha Numeric Sender ID would be a requirement in some destinations and verifying that the server would use those values when posting a payload.
https://support.twilio.com/hc/en-us/articles/223133767-International-support-for-Alphanumeric-Sender-ID

sethvargo:
I think that's fine. The server operator would need to take those steps, but I don't think it would affect our code. Can you share more about your concerns?

sherifkozman:
I am only noting that the Sender ID (i.e numeric or alpha numeric) is a value that should be factored in when integrating with twilio to ensure portability of the service outside of the US. So it would be set as an ENV variable for example the same way you add Twilio API Key and the application SID. Twilio mostly doesnt handle overriding that value on behalf of the sender.

mikehelmick:
The SMS config will need to be in the database and configured on a per-local basis. See #104

mikehelmick:
Follow up from previous comment

SMS provider and use should be database driven.
If there is no provider configured, then the dialog shouldn't be shown.

As we go multi-tenant, this is important since it is possible that different "realms" will use different SMS config and some might not use SMS at all.

sethvargo:
No no, bad GitHub. That was only part 1 of 1000.

PR #190, opened 2020-08-06
Add KMS<->DB abstraction

sethvargo:
- Fixes GH-165
- Fixes GH-117

**Release Note**

```release-note
Use Key Manager to encrypt SMS configurations
```


sethvargo:
screenshot-20200805-212541@2x

cmd/server/assets/realm.html
```diff
@@ -18,15 +18,57 @@ <h1>Realm settings</h1>
```
sethvargo, line 18: See screenshot for render

pkg/database/sms_config.go
```diff
@@ -16,58 +16,113 @@ package database
...
+	if r.twilioAuthToken != "" {
+		r.TwilioAuthToken = r.twilioAuthToken
+	}
+	return nil
+}
+
+// AfterFind runs after a record is found. It's used to mutate values (such as
+// auth tokens) before storing them on the struct. Do not call this function,
+// gorm calls it automatically.
+func (r *SMSConfig) AfterFind() error {
```
mikehelmick, line 92: There needs to be some kind of auto upgrade from secret to encrypted in DB for existing installations.

sethvargo: Added

pkg/controller/realmadmin/save.go
```diff
@@ -48,14 +56,74 @@ func (c *Controller) HandleSave() http.Handler {
 			return
 		}
 
+		// SMS config is all or nothing
+		if (form.TwilioAccountSid != "" || form.TwilioAuthToken != "" || form.TwilioFromNumber != "") &&
+			(form.TwilioAccountSid == "" || form.TwilioAuthToken == "" || form.TwilioFromNumber == "") {
+			flash.Error("Error updating realm: all SMS fields are required")
```
mikehelmick, line 62: nit: not quite accurate, either they're all required or must be blank to disable SMS

sethvargo: Done

Issue #176, opened 2020-08-05
Change verification code rate limit to per-realm


mikehelmick:
### TL;DR

Currently rate limiting is per API key, it should be per-realm

/kind enhancement

crwilcox:
Are user rate limits should likely have a look as well?

sethvargo:
Hey @crwilcox - in which server are you thinking?

PR #192, opened 2020-08-06
HMAC API Keys, implement per-realm limiting on API keys

sethvargo:
Fixes GH-176
Fixes GH-198

## Proposed Changes

* Introduce v2 API keys that encode the realm (+signature to prevent tampering), fallback to looking up API keys in the database for backwards compatibility 
* Stop storing plaintext API keys in the database. Instead HMAC the value and add another column which stores the first few bytes of the API key for display purposes
* Add migrations to update existing API keys
* Add complete CRUD operations for managing API keys

**Release Note**

```release-note
Introduce v2 API keys and switch to rate-limiting by realm instead of by key
```


pkg/database/authorized_app.go
```diff
@@ -130,3 +135,56 @@ func (db *Database) FindAuthorizedAppByAPIKey(apiKey string) (*AuthorizedApp, er
...
+	if _, err := rand.Read(buf); err != nil {
+		return "", fmt.Errorf("failed to rand: %w", err)
+	}
+	key = fmt.Sprintf("%s:%s", key, base64.RawStdEncoding.EncodeToString(buf))
+
+	// Create the HMAC of the key and the realm
+	sig := SignAPIKey(key)
+
+	// Put it all together
+	key = fmt.Sprintf("%s:%s", key, base64.RawStdEncoding.EncodeToString(sig))
```
mikehelmick, line 153: should separator be "." so it doesn't get confused w/ HTTP header "key: value"

sethvargo:
Okay @mikehelmick this is ready for review now

pkg/database/migrations.go
```diff
@@ -31,6 +31,8 @@ const initState = "00000-Init"
 func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 	logger := logging.FromContext(ctx)
 	options := gormigrate.DefaultOptions
+	options.UseTransaction = true
```
mikehelmick, line 34: I think this will break some of the older migrations

pkg/database/migrations.go
```diff
@@ -406,6 +409,63 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+				}
+
+				for _, app := range apps {
+					// If the key has a preview, it's v2
+					if app.APIKeyPreview != "" {
+						continue
+					}
+
+					apiKeyPreview := app.APIKey[:6]
+					newAPIKey, err := db.hmacAPIKey(app.APIKey)
```
mikehelmick, line 451: My worry w/ this is making sure the DB migrate tool has the right HMAC config present

pkg/database/migrations.go
sethvargo: I re-ran all of them a few times from scratch - only one was broken and it's fixed later.

pkg/database/migrations.go
sethvargo: It calls `config.Open` which _requires_ the HMAC key to be present. If it's not it'll fail opening the db.

Should we set these in our build scripts? I don't really have an opinion on the best way to get them into the migration step.

mikehelmick: 3 thoughts


1 - getting the secret ref into the build script would be OK
2 - We could deploy DB migration as a private cloud run service and CURL it w/ the right IAM permissions. Then it would be easy to terraform in the right data.
3 - Move this upgrade into the main server and handle it live on server startup. Can use the cleanup lock thing to run it once and set the next cleanup to occur years out. Then, say 2 releases out, we just remove this.

PR #196, opened 2020-08-06
Add a /healthz

icco:
Some progress towards #180

## Proposed Changes

* Adds a /healthz handler that can be hit by probers. if /healthz?service=database we can also find out if the database connection is working.
* This is useful for lightweight health checking

**Release Note**

```release-note
Adds a /healthz to all server binaries for healthchecking.
```

### DIFF
--- a/pkg/controller/healthz.go
+++ b/pkg/controller/healthz.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
-	"github.com/google/exposure-notifications-verification-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
 
 	"golang.org/x/time/rate"
@@ -17,7 +16,6 @@ var (
 )
 
 func HandleHealthz(hctx context.Context, h *render.Renderer, cfg *database.Config) http.Handler {
-	logger := logging.FromContext(hctx).Named("healthz")
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		params := r.URL.Query()
@@ -25,15 +23,13 @@ func HandleHealthz(hctx context.Context, h *render.Renderer, cfg *database.Confi
 			if rl.Allow() {
 				db, err := cfg.Open(ctx)
 				if err != nil {
-					logger.Errorw("database error", "error", err)
-					h.JSON500(w, render.ErrInternal)
+					InternalError(w, r, h, err)
 					return
 				}
 				defer db.Close()
 
 				if err := db.Ping(ctx); err != nil {
-					logger.Errorw("database error", "error", err)
-					h.JSON500(w, render.ErrInternal)
+					InternalError(w, r, h, err)
 					return
 				}
 			}
--- a/pkg/controller/healthz.go
+++ b/pkg/controller/healthz.go
@@ -15,23 +15,23 @@ var (
 	rl = rate.NewLimiter(rate.Every(time.Minute), 1)
 )
 
-func HandleHealthz(h *render.Renderer, cfg *database.Config) http.Handler {
+func HandleHealthz(hctx context.Context, h *render.Renderer, cfg *database.Config) http.Handler {
+	logger := logging.FromContext(hctx).Named("healthz")
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-		logger := logging.FromContext(ctx).Named("healthz")
 		params := r.URL.Query()
 		if s := params.Get("service"); s == "database" {
 			if rl.Allow() {
 				db, err := cfg.Open(ctx)
 				if err != nil {
-					logger.Errorw("database error", err)
+					logger.Errorw("database error", "error", err)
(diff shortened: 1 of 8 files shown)


icco:
whoops.

/retest

pkg/controller/healthz.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/logging"
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+)
+
+func HandleHealthz(h *render.Renderer, cfg *database.Config) http.Handler {
```
mikehelmick, line 11: I'm worried that having this open and unauthenticated and not rate limited makes it easy to burn through available SQL connections (since this creates a new connection each time)

icco: That's reasonable. We can totally put behind the rate limiter, or just remove the db portion. Whichever you prefer.

mikehelmick: random options

- put a mutex in this handler + only actually run the check ever 30/60 seconds
- drop the connection open and just do the ping
- put in IP based rate limiting w/ a pretty low limit
- make this IP whitelist or require an API key?

sethvargo: I'd like to learn more about why we need /healthz endpoints? AFAIK, Cloud Run can't poke them. If we're on K8S, I think the health endpoints should bind to 127.0.0.1 instead of 0.0.0.0 so they are only available in-pod.

icco: Ping will always open a connection https://golang.org/pkg/database/sql/#DB.Ping, but a mutex is easy enough. Added time/rate to do this.

pkg/controller/healthz.go
```diff
@@ -0,0 +1,43 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.FromContext(ctx).Named("healthz")
+		params := r.URL.Query()
+		if s := params.Get("service"); s == "database" {
+			if rl.Allow() {
+				db, err := cfg.Open(ctx)
+				if err != nil {
+					logger.Errorw("database error", err)
```
sethvargo, line 27: Needs to be `logger.Errorw("database error", "error", err)` and below

pkg/controller/healthz.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+
+	"golang.org/x/time/rate"
+)
+
+var (
+	rl = rate.NewLimiter(rate.Every(time.Minute), 1)
+)
+
+func HandleHealthz(h *render.Renderer, cfg *database.Config) http.Handler {
```
sethvargo, line 18: This should accept a context and pull the logger from the context. The logger isn't stored on the request context, it's stored on the handler context.

pkg/controller/healthz.go
```diff
@@ -0,0 +1,43 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := logging.FromContext(ctx).Named("healthz")
+		params := r.URL.Query()
+		if s := params.Get("service"); s == "database" {
+			if rl.Allow() {
+				db, err := cfg.Open(ctx)
+				if err != nil {
+					logger.Errorw("database error", err)
+					h.JSON500(w, err)
```
sethvargo, line 28: This should probably be `h.InternalError`. Otherwise healthz could leak information. Same below.

pkg/controller/healthz.go
icco: @sethvargo this is to allow stackdriver (or any other prober) to hit our service with as little business logic is possible. This isn't for our deploy system to do a check but rather for external black box probing. It's especially needed since the majority of our systems don't have a consistent url that always returns a 200 on unauthed GET.

pkg/controller/healthz.go
icco: Oh that's interesting, and surprising. Not what I would have expected, but changed.

pkg/controller/healthz.go
icco: Done.

pkg/controller/healthz.go
icco: good catch, fixed.

pkg/controller/healthz.go
```diff
@@ -26,14 +26,14 @@ func HandleHealthz(hctx context.Context, h *render.Renderer, cfg *database.Confi
 				db, err := cfg.Open(ctx)
 				if err != nil {
 					logger.Errorw("database error", "error", err)
-					h.JSON500(w, h.InternalError)
+					h.JSON500(w, render.InternalError)
```
sethvargo: Sorry, I meant `controller.InternalError(w, r, h)`

sethvargo: In dev mode, that'll actually render the error. In prod, it'll be a generic 500 page.

sethvargo: @icco ^

icco: Done.

PR #197, opened 2020-08-06
Automate deploy in cloud build

icco:
## Proposed Changes

* Expands root cloudbuild.yaml to build and publish all four service images
* One those are done, it deploys, but does not promote, those to the cloud run services
* Right now these aren't triggered, as cloud build triggers aren't enabled, but thinking on push to main at the very least. Currently also tags PRs if we want to use it on every push.

**Release Note**

```release-note
NONE
```


cloudbuild.yaml
```diff
@@ -9,50 +9,107 @@ steps:
...
+  - 'cleanup:publish'
+  - 'server:publish'
+  name: 'gcr.io/google.com/cloudsdktool/cloud-sdk:alpine'
+  args:
+  - 'bash'
+  - '-exEuo'
+  - 'pipefail'
+  - '-c'
+  - |-
+    gcloud components install --quiet beta;
```
sethvargo, line 98: What do we need from beta?

PR #206, opened 2020-08-08
Switch to built-in base64-bytes helper

sethvargo:
**Release Note**

```release-note
NONE
```


sethvargo:
@mikehelmick can you re-LGTM. I missed removing an import.

PR #207, opened 2020-08-08
Spelling Nits

whaught:
## Proposed Changes

* Just reading through code, fixing typos

**Release Note**

```release-note
Minor spelling fixes

Environment variable 'COLISSION_RETRY_COUNT' renamed 'COLLISION_RETRY_COUNT'
Server config field  'SESSION_DUATION' to 'SESSION_DURATION'
```


mikehelmick:
Can you add release notes that this fixes spelling on 2 environment variables that may need to be adjusted in your environment?

None of those are set by the TF config, so that's good.

whaught:
> Can you add release notes that this fixes spelling on 2 environment variables that may need to be adjusted in your environment?
> 
> None of those are set by the TF config, so that's good.

Good point. Done.

Hoping nobody has overridden these, but also that renaming them now will prevent future confusion.

PR #210, opened 2020-08-09
Switch to main server logger, always set a logger

sethvargo:
Switch to main server logger, always set a logger

**Release Note**

```release-note
Use shared logging package, always configure a logger
```


pkg/controller/middleware/method.go
```diff
@@ -19,7 +19,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/google/exposure-notifications-verification-server/pkg/logging"
+	"github.com/google/exposure-notifications-server/pkg/logging"
```
icco, line 22: nit: consistent whitespace missing.

cmd/add-users/main.go
```diff
@@ -18,67 +18,89 @@ package main
...
+		logger.Fatal(err)
+	}
+}
+
+func realMain(ctx context.Context) error {
+	logger := logging.FromContext(ctx)
+
 	if len(flag.Args()) > 0 {
-		log.Fatal("Received unexpected arguments:", flag.Args())
+		return fmt.Errorf("unexpected arguments: %v", flag.Args())
```
icco, line 60: nit: %w

pkg/controller/middleware/method.go
sethvargo: It's actually consistent now 😄 . It goes:

- Group 0: stdlib package
- Group 1: internal packages
- Group 2: main server packages
- Group 3: external packages

We were pretty inconsistent about it, so I fixed it in all the files that this touches.

icco: yeah, there's no whitespace between group 2 and 3 in this file, while there is in all other files in this commit.

Issue #209, opened 2020-08-09
Rate limiting needs more visibiilty


mikehelmick:
### TL;DR

Observed the main UX server rate limiting all requests on the landing page (no login)

Because of the middleware interceptor, there was no visibility into why.

Ultimately unblocked this by restarting the server.

/assign @sethvargo

sethvargo:
The response should include 4 headers:

- `X-RateLimit-Limit` - the total configured limit
- `X-RateLimit-Remaining` - the number of remaining requests available in the current interval (0 == the next request will be rate limited)
- `X-RateLimit-Reset - the server time at which the rate limit resets
- `Retry-After` (will be the same as X-RateLimit-Reset, but is canonical)

The most plausible course seems to be that the service failed to establish a connection to redis. In that case, Limit and Remaining would both be 0, and Reset would be Unix(0) (Jan 1, 1970).

We could:

1.  Add more logging to tell us exactly what caused the failed connection
1.  Change the default to be allow (instead of deny) on connection failures
1.  Do nothing and gather the full headers if it happens again

@mikehelmick WDYT?

gurayAlsac:
Could there be a mode where rate-limiting is advisory only (doesn't actually block), or at least in (presumably rare) case db connectivity is lost, it fails open?

sethvargo:
I’m hesitant to make it this advisory only, since it seems like clients already aren’t honoring the response headers.

sethvargo:
We can make it fail open, but it opens us up to a class of attack; if someone can DoS Redis, they could effectively bypass ratelimiting, which feels undesireable.

gurayAlsac:
These would be configurations for testing server only.

sethvargo:
Yes, but if we failed open, this issue would have happened silently in the background, and we never would have uncovered that clients weren't parsing and obeying the rate limiting headers.

sethvargo:
Observability work by @icco going on at GH-226

PR #212, opened 2020-08-09
Allow configuring the rate limiter to fail open

sethvargo:
Refs GH-209

I don't want to change the default value, but we can override this in our configurations after we build and deploy a new version if this continues to be an issue.

**Release Note**

```release-note
Allow configuring the rate limiter to succeed on backend connection failures (default unchanged)
```

/assign @gurayAlsac 
/cc @mikehelmick


mikehelmick:
Note that failing open in dev environments isn't necessarily great unless we explicitly test client behaviour for when this is enabled.

Issue #134, opened 2020-07-24
Custom pin failure errors


mikehelmick:
### TL;DR

Allow PHAs to customizer pin validation failure messages. 
Will allow for providing instructions on how to obtain a new pin.

/kind enhancement

mikehelmick:
Also - for pin error codes, this will be put in place

A pin we don't know about at all is "inavlid"
A pin that was valid, but used is "invalid" (hasn't reached expiry time yet)
A pin that is known and expired is "expired"

mikehelmick:
Also - add in a set of universal error codes so that things can be localized on the client

```
{"Error": "english error message only for devs and logs", "ErrorCode": "code_already_used"}
```

mikehelmick:
Here are the proposed `ErrorCode` strings

- Pin we don't know about would be `"ErrorCode": "code_invalid"`
- Pin that was valid but used would be `"ErrorCode": "code_invalid"`
- Pin that is known and expired would be `"ErrorCode": "code_expired"`

Similar treatment will need to be given to verification token validation.

sethvargo:
I'd recommend keys of `"code"` and `"message"`, but otherwise lgtm

mikehelmick:
/close

this is part of client side config, we now have custom error strings "ErrorCode" in return

PR #214, opened 2020-08-10
Unify on BadRequest

sethvargo:
This reduces the number of error codes that clients need to check against to:

- `400` - search for `"errors"` key in JSON response
- `401` - unauthorized, usually with no `"errors"`, but clients can try to read the JSON body to see if there's additional information (it may be empty)
- `404` - routing error (app coding error)
- `405` - wrong HTTP method (app coding error)
- `429` - rate limited, check headers `X-Retry-After` to determine when to retry
- `5xx` - internal server error

**Release Note**

```release-note
Standardize known HTTP response codes on 400, 401, 404, 405, 429, and 5xx.
```

### DIFF
--- a/pkg/render/json.go
+++ b/pkg/render/json.go
@@ -38,6 +38,21 @@ import (
 // The buffers are fetched via a sync.Pool to reduce allocations and improve
 // performance.
 func (r *Renderer) RenderJSON(w http.ResponseWriter, code int, data interface{}) {
+	// Hello there reader! If you've made it here, you're likely wondering why
+	// you're getting an error about response codes. For client-interop, it's very
+	// important that we retain and maintain the allowed list of response codes.
+	// Adding a new response code requires coordination with the client team so
+	// they can update their applications to handle that new response code.
+	if !r.AllowedResponseCode(code) {
+		r.logger.Errorw("unregistered response code", "code", code)
+
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Header().Set("Content-Type", "application/json")
+		msg := escapeJSON(fmt.Sprintf("%d is not a registered response code", code))
+		fmt.Fprintf(w, jsonErrTmpl, msg)
+		return
+	}
+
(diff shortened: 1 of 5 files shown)


sethvargo:
@icco done

mikehelmick:
Related to #134 - we need to get JSON level error codes in as well (step further than this PR)

pkg/render/render.go
```diff
@@ -28,6 +28,19 @@ import (
 	"go.uber.org/zap"
 )
 
+// allowedResponseCodes are the list of allowed response codes. This is
+// primarily here to catch if someone, in the future, accidentially includes a
+// bad status code.
+var allowedResponseCodes = map[int]struct{}{
```
gurayAlsac, line 34: Do these need to be a map of structs? Would a set of ints not be able to represent the allowed codes?

pkg/render/json.go
```diff
@@ -38,6 +38,21 @@ import (
 // The buffers are fetched via a sync.Pool to reduce allocations and improve
 // performance.
 func (r *Renderer) RenderJSON(w http.ResponseWriter, code int, data interface{}) {
+	// Hello there reader! If you've made it here, you're likely wondering why
+	// you're getting an error about response codes. For client-interop, it's very
+	// important that we retain and maintain the allowed list of response codes.
+	// Adding a new response code requires coordination with the client team so
+	// they can update their applications to handle that new response code.
+	if !r.AllowedResponseCode(code) {
```
gurayAlsac, line 46: naming nit: perhaps "AppHandledResponseCode" communicates the concept more directly? We don't want to send these codes back until client reference app handles them.

icco:
/lgtm 

@gurayAlsac's comments are reasonable, but could totally be dealt with in a future PR.

pkg/render/render.go
sethvargo: Go structs are 0 bytes, so a map[int]struct{} is essentially a slice with O(1) lookups

Issue #109, opened 2020-07-13
Verification Code Status


Ferrumofomega:
### TL;DR

Once an authentication code is issued for a positive test in the verification service, we would like to track the expiration timestamp and if the code was used, so that we can enable more workflows (both automated at not). Right now the database stores these data, but does not expose to the user and or via API.

[This feature was requested by a jurisdiction looking to launch immediately]

### Design

**Proposal**

**Alternatives considered**

**Resources**

- [Link to documentation](TODO)

**Additional information**

crwilcox:
@Ferrumofomega I may not be understanding your request completely. My initial concern would be leaking of unused verification keys and potential abuse vectors.

Can you provide an example of the API call you would need to enable your scenario?

mikehelmick:
Can you share more about the workflows you wish to enable?

We do plan to give access to statistical data related to code use.

sherifkozman:
@mikehelmick along with the Admin API calls available , would it be possible for an a call to be implemented to check if an issued verification code was used or not.
The Data already exisits in the database about the status of the issues code , can that be availed so a third party integration would be able to get a status of an issued code?

mikehelmick:
It would still be good for us to understand the use cases and determine how much of that should be built directly into this server versus enabled through APIs.

sherifkozman:
The current use case and request would give contact tracers the ability to keep track of issues verification codes, if they were used or not and when were they used. This gives more context to CTs to keep track of the codes they issues for postive tested cases.
Exposing it through an API enables integrating this with exisitng systems within the HAs

mikehelmick:
We probably need to assign a tracking GUID for this purpose since codes can be recycled.

sherifkozman:
Thanks @mikehelmick

whaught:
Note switched to UUIDs: https://github.com/google/exposure-notifications-verification-server/pull/217

whaught:
@Ferrumofomega can you add some detail about the scenario users are looking for here? Are we expecting to check on the status of a single code or many?

Ferrumofomega:
Happily. Our user (a US state) would like a way (updated as of late, this
was going to be manual) to programmatically check the status of a single
code.


Samuel Zimmerman
calendly.com/zimmerman-samuel
814-279-7110

whaught:
[...]

PR #215, opened 2020-08-10
Emit ID for verification code

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/109

## Proposed Changes

* Include the Database ID for the issued verification code. This will be later used in a new API to check the claimed/unclaimed status of the code.

### Open Question
* Are there any security concerns with using this DB handle / do we need to create an artificial GUID key for this purpose. The DB key is an auto-increment uint which may allow for scraping or counting; However the user must be authenticated to retrieve code status.

**Release Note**

```release-note
Include ID in the issue VerificationCode response
```

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -15,6 +15,8 @@
 package issueapi
 
 import (
+	"encoding/base64"
+	"encoding/binary"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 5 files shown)


icco:
Google API best practices say that we should use a string for all publicly visible IDs. If it's not a significant amount of work, we should use a GUID instead of an incrementing int.

sethvargo:
> Are there any security concerns with using this DB handle / do we need to create an artificial GUID key for this purpose. The DB key is an auto-increment uint which may allow for scraping or counting; However the user must be authenticated to retrieve code status.

I don't think this vector is a concern in this case. If we go with UUIDs, we'd want to do that at the database layer and use that as the primary key instead of having two primary keys ([example](https://www.starkandwayne.com/blog/uuid-primary-keys-in-postgresql/)).

sethvargo:
It's also worth noting that we already use IDs in the URLs for API keys. If we want to switch to UUIDs, we might consider doing that as a wholesale change across all models as a followup?

pkg/controller/issueapi/issue.go
```diff
@@ -143,8 +145,13 @@ func (c *Controller) HandleIssue() http.Handler {
 			}
 		}
 
+		var intAsBytes []byte
+		binary.LittleEndian.PutUint64(intAsBytes, uint64(id))
+		idString := base64.StdEncoding.EncodeToString(intAsBytes)
```
sethvargo, line 150: You'll probably want URLEncoding here instead, since this will be used as part of a URL. However, it's probably better to use hex here (even though it's longer).

sethvargo: Also document this 😄

PR #217, opened 2020-08-11
Use database UUIDs

sethvargo:
Add UUIDs to verification tokens

**Release Note**

```release-note
Returning UUID with verification tokens
```


sethvargo:
@icco FYI - we were seeing some panics with the hex implementation 😦

icco:
Ah sad alrighty.

PR #218, opened 2020-08-11
Update deployment/release with fixes

sethvargo:
I ran into a few issues while trying to deploy:

1. Some mis-documentation about secrets
2. The migrate server didn't resolve secrets
3. The UUID addition would fail because SET DEFAULT and NOT NULL runs in the same transaction

**Release Note**

```release-note
Update migrations to resolve secrets and fix broken migration for UUIDs on verification tokens
```

### DIFF
--- a/builders/migrate.yaml
+++ b/builders/migrate.yaml
@@ -23,31 +23,22 @@ substitutions:
   _DB_ENCRYPTION_KEY:
   _DB_HOST: 'localhost'
   _DB_NAME: 'verification'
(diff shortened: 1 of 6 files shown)


icco:
/lgtm

Thanks for doing this, I was planning on doing it during lunch :D

sethvargo:
I think we'll need to cut a patch release since the migration is broken in 0.3.0. I'll do that in a bit.

PR #219, opened 2020-08-11
CheckCodeStatus API

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/109

## Proposed Changes

* Create a query for codes by UUID
* Create a request/response for new check status API
* Logic for retrieval
    * User/realm must match issuing user/realm
    * expired/no result is notfound
* E2E test asserts code is 'claimed'

TODO: html interface

**Release Note**

```release-note
CheckCodeStatus API allows user to check the status of an OTP code
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -105,7 +105,7 @@ type IssueCodeResponse struct {
 
 // CheckCodeStatusRequest defines the parameters to request the status for a
 // previously issued OTP code. This is called by the Web frontend.
-// API is served at /api/codestatus
+// API is served at /api/checkcodestatus
 type CheckCodeStatusRequest struct {
 	ID string `json:"id"`
 }
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -117,7 +117,6 @@ type CheckCodeStatusResponse struct {
 
 // CodeStatus defines the status of an OTP code.
 type CodeStatus struct {
-	ID                 string `json:"id"`
 	Claimed            bool   `json:"claimed"`
 	ExpiresAt          string `json:"expiresAt"`          // RFC1123 string formatted timestamp, in UTC.
 	ExpiresAtTimestamp int64  `json:"expiresAtTimestamp"` // Unix, seconds since the epoch. Still UTC.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -107,16 +107,11 @@ type IssueCodeResponse struct {
 // previously issued OTP code. This is called by the Web frontend.
 // API is served at /api/codestatus
 type CheckCodeStatusRequest struct {
-	ID []string `json:"id"`
+	ID string `json:"id"`
 }
 
 // CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
 type CheckCodeStatusResponse struct {
-	CodeStatus map[string]CodeStatus `json:"codeStatus"`
-}
-
-// CodeStatus defines the status of an OTP code.
-type CodeStatus struct {
 	Claimed            bool   `json:"claimed"`
 	ExpiresAt          string `json:"expiresAt"`          // RFC1123 string formatted timestamp, in UTC.
 	ExpiresAtTimestamp int64  `json:"expiresAtTimestamp"` // Unix, seconds since the epoch. Still UTC.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -95,8 +95,8 @@ type IssueCodeRequest struct {
 
 // IssueCodeResponse defines the response type for IssueCodeRequest.
 type IssueCodeResponse struct {
-	// ID is a UUID handle which allows the issuer to track status of the issued verification code.
-	ID string `json:"id"`
+	// UUID is a handle which allows the issuer to track status of the issued verification code.
+	UUID string `json:"uuid"`
 
 	// The OTP code which may be exchanged by the user for a signing token.
 	VerificationCode string `json:"code"`
@@ -117,7 +117,8 @@ type IssueCodeResponse struct {
 // previously issued OTP code. This is called by the Web frontend.
 // API is served at /api/checkcodestatus
 type CheckCodeStatusRequest struct {
-	ID string `json:"id"`
+	// UUID is a handle which allows the issuer to track status of the issued verification code.
+	UUID string `json:"uuid"`
 }
 
 // CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -125,10 +125,6 @@ type CheckCodeStatusResponse struct {
 	// Claimed is true if a user has used the OTP code to get a token via the VerifyCode api.
 	Claimed bool `json:"claimed"`
 
-	// ExpiresAt is a RFC1123 formatted string formatted timestamp, in UTC.
-	// After this time the code will no longer be accepted and is eligible for deletion.
-	ExpiresAt string `json:"expiresAt"` // RFC1123 string formatted timestamp, in UTC.
-
 	// ExpiresAtTimestamp represents Unix, seconds since the epoch. Still UTC.
 	// After this time the code will no longer be accepted and is eligible for deletion.
 	ExpiresAtTimestamp int64 `json:"expiresAtTimestamp"`
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -103,6 +103,26 @@ type IssueCodeResponse struct {
 	ErrorCode          string `json:"errorCode,omitempty"`
 }
 
+// CheckCodeStatusRequest defines the parameters to request the status for a
+// previously issued OTP code. This is called by the Web frontend.
+// API is served at /api/codestatus
+type CheckCodeStatusRequest struct {
+	ID []string `json:"id"`
+}
+
+// CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
+type CheckCodeStatusResponse struct {
+	CodeStatus map[string]CodeStatus `json:"codeStatus"`
+}
+
+// CodeStatus defines the status of an OTP code.
+type CodeStatus struct {
+	Claimed            bool   `json:"claimed"`
+	VerificationCode   string `json:"code"`
+	ExpiresAt          string `json:"expiresAt"`          // RFC1123 string formatted timestamp, in UTC.
+	ExpiresAtTimestamp int64  `json:"expiresAtTimestamp"` // Unix, seconds since the epoch. Still UTC.
+}
+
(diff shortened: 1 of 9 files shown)


pkg/database/vercode.go
```diff
@@ -110,6 +110,16 @@ func (db *Database) FindVerificationCode(code string) (*VerificationCode, error)
 	return &vc, nil
 }
 
+// FindVerificationCodeByIDs find a list of verification codes by UUID.
+func (db *Database) FindVerificationCodeByIDs(uuids []string) ([]VerificationCode, error) {
```
sethvargo, line 114: This feels expensive and is also untested. Would we every query by more than one?

pkg/api/api.go
```diff
@@ -103,6 +103,26 @@ type IssueCodeResponse struct {
...
+
+// CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
+type CheckCodeStatusResponse struct {
+	CodeStatus map[string]CodeStatus `json:"codeStatus"`
+}
+
+// CodeStatus defines the status of an OTP code.
+type CodeStatus struct {
+	Claimed            bool   `json:"claimed"`
+	VerificationCode   string `json:"code"`
```
mikehelmick, line 121: not sure if we want to repeat the actual verification code

pkg/api/api.go
```diff
@@ -103,6 +103,26 @@ type IssueCodeResponse struct {
...
+// CheckCodeStatusRequest defines the parameters to request the status for a
+// previously issued OTP code. This is called by the Web frontend.
+// API is served at /api/codestatus
+type CheckCodeStatusRequest struct {
+	ID []string `json:"id"`
+}
+
+// CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
+type CheckCodeStatusResponse struct {
+	CodeStatus map[string]CodeStatus `json:"codeStatus"`
```
mikehelmick, line 115: What about for codes that are not found (i.e. have been garbage collected)?

pkg/api/api.go
```diff
@@ -103,6 +103,26 @@ type IssueCodeResponse struct {
 	ErrorCode          string `json:"errorCode,omitempty"`
 }
 
+// CheckCodeStatusRequest defines the parameters to request the status for a
+// previously issued OTP code. This is called by the Web frontend.
+// API is served at /api/codestatus
+type CheckCodeStatusRequest struct {
+	ID []string `json:"id"`
```
mikehelmick, line 110: might be good to ask the issue filer if they need the ability to do this by Array?

maybe one at a time is good enough?

pkg/api/api.go
whaught: such codes will be omitted in the response

pkg/api/api.go
whaught: one at a time is certainly easier. The issue filer mentioned automation which might involve multiple queries (eg a health provider checking the status for their patients for follow-up). I'll reply in the issue.

pkg/database/vercode.go
whaught: it shouldn't be expensive - it is an index query and should be no more expensive in the single ID case, but it should be cheaper than repeated queries. We will want to cap the length of the request batch size.

I'm following up in the issue to ensure that the user requirement is batch oriented.

pkg/api/api.go
whaught: apologies. I intended to include the ID, not the code. That said what is the risk of repeating the code?

pkg/api/api.go
whaught: The user notes they only need one

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		code, err := c.db.FindVerificationCodeByID(request.ID)
+		if err != nil {
+			logger.Errorw("failed to check otp code status", "error", err)
+			c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to check otp code status, please try again"))
+			return
+		}
+
+		if code.IssuingUser.Email != user.Email {
```
sethvargo, line 67: Is this required? Seems like anyone who is an admin should be able to lookup code status, no?

Also, I believe `user` can be nil from `getAuthorizationFromContext`.

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,98 @@
...
+			if realm == nil {
+				c.h.RenderJSON(w, http.StatusBadRequest, api.Errorf("missing realm"))
+				return
+			}
+		}
+
+		code, err := c.db.FindVerificationCodeByID(request.ID)
+		if err != nil {
+			logger.Errorw("failed to check otp code status", "error", err)
+			c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to check otp code status, please try again"))
```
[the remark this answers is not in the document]
whaught: this function returns an empty instance of VerificationCode if there are 0 rows found.

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		if code.RealmID != realm.ID {
+			logger.Errorw("failed to check otp code status", "error", "realmID does not match")
+			c.h.RenderJSON(w, http.StatusNotFound, api.Errorf("failed to check otp code status"))
+			return
+		}
+
+		if code.UUID != request.ID {
+			logger.Errorw("failed to check otp code status", "error", "code not found")
```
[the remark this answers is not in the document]
whaught: if FindVerificationCodeByID finds no rows, code is an empty instance of VerificationCode. This is the notfound case (but that's obviously not clear here)

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,98 @@
...
+
+		if code.IssuingUser.Email != user.Email {
+			logger.Errorw("failed to check otp code status", "error", "user email does not match issuing user")
+			c.h.RenderJSON(w, http.StatusUnauthorized, api.Errorf("failed to check otp code status: user does not match issuing user"))
+			return
+		}
+
+		if code.IsExpired() {
+			logger.Errorw("failed to check otp code status", "error", "code exists but is expired")
+			c.h.RenderJSON(w, http.StatusNotFound, api.Errorf("failed to check otp code status"))
```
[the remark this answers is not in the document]
whaught: I like that. I was trying to not give away the exact condition.

pkg/controller/issueapi/codestatus.go
whaught: True. Added a nil check. I was going to start with the most restrictive thing (only the issuer can check the status)

pkg/api/api.go
```diff
@@ -103,6 +103,20 @@ type IssueCodeResponse struct {
...
+// API is served at /api/checkcodestatus
+type CheckCodeStatusRequest struct {
+	ID string `json:"id"`
+}
+
+// CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
+type CheckCodeStatusResponse struct {
+	Claimed            bool   `json:"claimed"`
+	ExpiresAt          string `json:"expiresAt"`          // RFC1123 string formatted timestamp, in UTC.
+	ExpiresAtTimestamp int64  `json:"expiresAtTimestamp"` // Unix, seconds since the epoch. Still UTC.
```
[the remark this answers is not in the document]
whaught: These timestamps match the format of the IssueCodeResponse. It would be odd to have them differ?

pkg/api/api.go
```diff
@@ -103,6 +103,20 @@ type IssueCodeResponse struct {
...
+// previously issued OTP code. This is called by the Web frontend.
+// API is served at /api/checkcodestatus
+type CheckCodeStatusRequest struct {
+	ID string `json:"id"`
+}
+
+// CheckCodeStatusResponse defines the response type for CheckCodeStatusRequest.
+type CheckCodeStatusResponse struct {
+	Claimed            bool   `json:"claimed"`
+	ExpiresAt          string `json:"expiresAt"`          // RFC1123 string formatted timestamp, in UTC.
```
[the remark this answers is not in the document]
whaught: Copied from IssueCodeRepsonse. It would be odd for these responses to not match

pkg/api/api.go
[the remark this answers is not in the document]
whaught: Sure let's start with the one. Still not sure why the Issue api has two

pkg/controller/issueapi/codestatus.go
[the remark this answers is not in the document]
whaught: Updated to make this more obvious

pkg/api/api.go
```diff
@@ -103,6 +103,20 @@ type IssueCodeResponse struct {
 	ErrorCode          string `json:"errorCode,omitempty"`
 }
 
+// CheckCodeStatusRequest defines the parameters to request the status for a
+// previously issued OTP code. This is called by the Web frontend.
+// API is served at /api/checkcodestatus
+type CheckCodeStatusRequest struct {
+	ID string `json:"id"`
```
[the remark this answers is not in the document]
whaught: Sure let's make them all UUID

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,101 @@
...
+		if authApp != nil {
+			realm, err = authApp.GetRealm(c.db)
+			if err != nil {
+				c.h.RenderJSON(w, http.StatusUnauthorized, nil)
+				return
+			}
+		} else {
+			// if it's a user logged in, we can pull realm from the context.
+			realm = controller.RealmFromContext(ctx)
+			if realm == nil {
```
sethvargo, line 58: I think we should move this check to ~63. That way it handles a case where `authApp.GetRealm()` happens to return a `nil` realm.

pkg/controller/issueapi/codestatus.go
```diff
@@ -0,0 +1,101 @@
...
+		if code.RealmID != realm.ID {
+			logger.Errorw("failed to check otp code status", "error", "realmID does not match")
+			c.h.RenderJSON(w, http.StatusNotFound, api.Errorf("code does not exist or has expired"))
+			return
+		}
+
+		c.h.RenderJSON(w, http.StatusOK,
+			&api.CheckCodeStatusResponse{
+				Claimed:            code.Claimed,
+				ExpiresAtTimestamp: code.ExpiresAt.Unix(),
```
sethvargo, line 98: .UTC()

pkg/controller/issueapi/codestatus.go
whaught: agreed. Same for issue.go

PR #221, opened 2020-08-11
Fix rate limiting

sethvargo:
- Download go modules in dedicated step for faster local builds
- Update go-limiter to fix a redis TTL issue
- Update adminapi server config to use shared ratelimit config
- Rate limit by X-Forwarded-For, ensure limiting is applied in correct order

**Release Note**

```release-note
Rate limit by x-forwarded-for headers to get the real client IP
The adminapi now uses `RATE_LIMIT_TOKENS` instead of `RATE_LIMIT` to define the limits
IP addresses are hashed before stored in redis for limiting
Each service has its own key namespace in the rate limiter to avoid clobbering
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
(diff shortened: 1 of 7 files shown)


cmd/server/main.go
```diff
@@ -122,18 +123,17 @@ func realMain(ctx context.Context) error {
...
 	}
-	defer store.Close()
+	defer limiterStore.Close()
 
-	httplimiter, err := httplimit.NewMiddleware(store, userEmailKeyFunc())
+	httplimiter, err := httplimit.NewMiddleware(limiterStore, limiterFunc(ctx))
 	if err != nil {
 		return fmt.Errorf("failed to create limiter middleware: %w", err)
 	}
-	r.Use(httplimiter.Handle)
```
sethvargo, line 136: @mikehelmick this is what I was saying about the middleware order... the `requireAuth` middleware sets the user on the request context, so I thought moving this to the end would fix that. However, the middleware seems to run for the parent router, then subrouters, regardless of when .Use is called. As a result, we have to inject this in each sub.

PR #226, opened 2020-08-13
Limit middleware

icco:
Adds metrics for #209

## Proposed Changes

* Creates a common rate limit middleware that logs metrics
* moves common limit functions to said middleware

**Release Note**

```release-note
Adds metrics for request rate limiting 
```


cmd/adminapi/main.go
```diff
@@ -18,18 +18,16 @@ package main
...
 	"os"
-	"strings"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/issueapi"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/middleware"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/ratelimit"
+	"github.com/google/exposure-notifications-verification-server/pkg/ratelimit/limitware"
```
sethvargo, line 30: Nice name

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+// KeyFunc is a function that accepts an http request and returns a string key
+// that uniquely identifies this request for the purpose of rate limiting.
+//
+// KeyFuncs are called on each request, so be mindful of performance and
+// implement caching where possible. If a KeyFunc returns an error, the HTTP
+// handler will return Internal Server Error and will NOT take from the limiter
+// store.
+type KeyFunc func(r *http.Request) (string, error)
+
+// IPKeyFunc returns a function that keys data based on the incoming requests IP
```
sethvargo, line 47: I don't think we need this

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/kv"
+	"go.opentelemetry.io/otel/api/metric"
+)
+
+const (
+	// HeaderRateLimitLimit, HeaderRateLimitRemaining, and HeaderRateLimitReset
+	// are the recommended return header values from IETF on rate limiting. Reset
+	// is in UTC time.
+	HeaderRateLimitLimit     = "X-RateLimit-Limit"
```
sethvargo, line 29: These are actually exported, so you could use them from the package, but it's also okay to just copy them.

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+func NewMiddleware(s limiter.Store, f KeyFunc) (*Middleware, error) {
+	if s == nil {
+		return nil, fmt.Errorf("store cannot be nil")
+	}
+
+	if f == nil {
+		return nil, fmt.Errorf("key function cannot be nil")
+	}
+
+	meter := global.Meter("github.com/sethvargo/go-limiter")
```
sethvargo, line 89: Is this the right meter? Or would we want to use this middleware package name instead?

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+// the store and sets the common rate limiting headers. If the take is
+// successful, the remaining middleware is called. If take is unsuccessful, the
+// middleware chain is halted and the function renders a 429 to the caller with
+// metadata about when it's safe to retry.
+func (m *Middleware) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Call the key function - if this fails, it's an internal server error.
+		key, err := m.keyFunc(r)
+		if err != nil {
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
sethvargo, line 113: Can we emit both a log message and a metric here? We can get the logger from the context on init and store it on the middleware.

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+		limit, remaining, reset, ok := m.store.Take(key)
+		resetTime := time.Unix(0, int64(reset)).UTC().Format(time.RFC1123)
+
+		// Set headers (we do this regardless of whether the request is permitted).
+		w.Header().Set(HeaderRateLimitLimit, strconv.FormatUint(limit, 10))
+		w.Header().Set(HeaderRateLimitRemaining, strconv.FormatUint(remaining, 10))
+		w.Header().Set(HeaderRateLimitReset, resetTime)
+
+		// Record request status
+		m.meter.RecordBatch(
```
sethvargo, line 127: I'm not super familiar with this implementation - does it block? Does it make sense to throw this into a go function so it can run independent and non-blocking of the limiter?

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+			r.Context(),
+			[]kv.KeyValue{kv.Bool("ok", ok)},
```
sethvargo, line 129: We're going to allocate this on each request. Maybe it makes sense to have this slice be initialized as a private var?

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+			m.reqCounter.Measurement(1),
+		)
+
+		// Fail if there were no tokens remaining.
+		if !ok {
+			w.Header().Set(HeaderRetryAfter, resetTime)
```
sethvargo, line 135: We should produce a metric here. Also maybe log.

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+			return
+		}
+
+		// If we got this far, we're allowed to continue, so call the next middleware
+		// in the stack to continue processing.
+		next.ServeHTTP(w, r)
+	})
+}
+
+// APIKeyFunc returns a default key function for ratelimiting on our API key header.
```
sethvargo, line 146: "It falls back to rate limiting by the client ip"

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+			// v1 API keys do not, fallback to the database
+			app, err := db.FindAuthorizedAppByAPIKey(v)
+			if err == nil && app != nil {
+				logger.Debugw("limiting by api key v1 realm", "realm", app.RealmID)
+				dig := sha1.Sum([]byte(fmt.Sprintf("%d", app.RealmID)))
+				return fmt.Sprintf("%s:realm:%x", scope, dig), nil
+			}
+		}
+
+		// Get the remote addr
```
sethvargo, line 171: This logic is duplicated - we should consider moving it into its own function for readability.

pkg/ratelimit/limitware/middleware.go
icco: Done.

pkg/ratelimit/limitware/middleware.go
icco: I moved to the older implementation, because finding the answer to this was unclear to me. This shouldn't block if I'm reading correctly.

pkg/ratelimit/limitware/middleware.go
icco: The above metric records the request success / fail status in the ok label.

pkg/ratelimit/limitware/middleware.go
icco: done.

pkg/ratelimit/limitware/middleware.go
icco: oh fair, deduped.

pkg/ratelimit/limitware/middleware.go
icco: Changed, good call.

icco:
Alrighty, ready for you again @sethvargo, sorry for the delay.

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+
+		// Set headers (we do this regardless of whether the request is permitted).
+		w.Header().Set(httplimit.HeaderRateLimitLimit, strconv.FormatUint(limit, 10))
+		w.Header().Set(httplimit.HeaderRateLimitRemaining, strconv.FormatUint(remaining, 10))
+		w.Header().Set(httplimit.HeaderRateLimitReset, resetTime)
+
+		// Record request status
+		ctx, err := tag.New(r.Context(), tag.Insert(OkTag, fmt.Sprintf("%v", ok)))
+		if err != nil {
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
sethvargo, line 113: Can we log this error or use `controller.InternalError`?

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+			return
+		}
+		stats.Record(ctx, m.reqCounter.M(1))
+
+		// Fail if there were no tokens remaining.
+		if !ok {
+			w.Header().Set(httplimit.HeaderRetryAfter, resetTime)
```
sethvargo, line 120: Should we emit a metric here too? It'd be nice to be able to look at a dashboard and see the total number of currently rate limited requests.

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+		// If we got this far, we're allowed to continue, so call the next middleware
+		// in the stack to continue processing.
+		next.ServeHTTP(w, r)
+	})
+}
+
+// APIKeyFunc returns a default key function for ratelimiting on our API key
+// header. It falls back to rate limiting by the client ip.
+func APIKeyFunc(ctx context.Context, scope string, db *database.Database) httplimit.KeyFunc {
+	logger := logging.FromContext(ctx).Named("ratelimit")
```
sethvargo, line 134: Since we have the scope now... wanna make the named logger `ratelimit.<scope>` or `<scope>.ratelimit`?

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		return ipAddrLimit(r)
+	}
+}
+
+// UserEmailKeyFunc pulls the user out of the request context and uses that to
+// ratelimit. It falls back to rate limiting by the client ip.
+func UserEmailKeyFunc(ctx context.Context, scope string) httplimit.KeyFunc {
+	logger := logging.FromContext(ctx).Named("ratelimit")
```
sethvargo, line 165: Same thing about scope

pkg/ratelimit/limitware/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+			return fmt.Sprintf("%s:user:%x", scope, dig), nil
+		}
+
+		return ipAddrLimit(r)
+	}
+}
+
+// IPAddressKeyFunc uses the client IP to rate limit.
+func IPAddressKeyFunc(ctx context.Context, scope string) httplimit.KeyFunc {
+	logger := logging.FromContext(ctx).Named("ratelimit")
```
sethvargo, line 185: Same thing about scope

sethvargo:
@icco https://github.com/google/exposure-notifications-verification-server/pull/250

Issue #224, opened 2020-08-12
Clarify language around symptom date and test date


mikehelmick:
### TL;DR

In the UI, symptom date should indicate to the user that they should enter the test date if they don’t know the symptom onset date.

/kind enhancement

mikehelmick:
Another design option

Change the UI and API to accept both symptom date and test date and then when minting the verification token select the right one

1) symptom onset if exists
2) test date if provided and #1 doesn’t exist
3) leave blank if neither

I think this would reduce cognative load on case workers insteading of having to interprate a multi-use date field.

sethvargo:
We could also default test date to today's date. Thoughts?

mikehelmick:
we’d need to run that by the medical team. I think the directive was to leave it blank if neither was known

PR #236, opened 2020-08-14
Support symptom date and test date

sethvargo:
Fixes GH-224

**Release Note**

```release-note
Support symptom date and test date in the UI
```

### DIFF
--- a/pkg/controller/home/home.go
+++ b/pkg/controller/home/home.go
@@ -80,7 +80,6 @@ func (c *Controller) HandleHome() http.Handler {
 		m := controller.TemplateMapFromContext(ctx)
 		// Set test date params
 		now := time.Now().UTC()
-		m["todayDate"] = now.Format("2006-01-02")
 		m["maxDate"] = now.Format("2006-01-02")
 		m["minDate"] = now.Add(c.pastDaysDuration).Format("2006-01-02")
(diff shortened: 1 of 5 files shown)


pkg/controller/home/home.go
```diff
@@ -74,6 +74,7 @@ func (c *Controller) HandleHome() http.Handler {
 		m := controller.TemplateMapFromContext(ctx)
 		// Set test date params
 		now := time.Now().UTC()
+		m["todayDate"] = now.Format("2006-01-02")
```
mikehelmick, line 77: do we need both todaysDate and maxDate?

pkg/controller/issueapi/issue.go
```diff
@@ -30,15 +30,34 @@ import (
 func (c *Controller) HandleIssue() http.Handler {
 	logger := c.logger.Named("issueapi.HandleIssue")
 
+	type FormData struct {
```
mikehelmick, line 33: why this change?

I'm concerned about drift in API definition.

mikehelmick: test date should be added to the adminapi version as well

sethvargo: I think I addressed this in 237483a

sethvargo:
@mikehelmick fixed

PR #241, opened 2020-08-14
Support per-realm config of allowed test types

sethvargo:
Fixes GH-223

**Release Note**

```release-note
Support per-realm config of allowed test types
```

/assign @mikehelmick

### DIFF
--- a/cmd/server/assets/home.html
+++ b/cmd/server/assets/home.html
@@ -1,5 +1,6 @@
 {{define "home"}}
 
+{{$currentRealm := .currentRealm}}
 {{$hasSMSConfig := .hasSMSConfig}}
 
 <!doctype html>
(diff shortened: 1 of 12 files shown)


cmd/server/assets/realm.html
```diff
@@ -30,7 +31,30 @@ <h1>Realm settings</h1>
...
+              {{if $realm.ErrorsFor "name"}}
+              <div class="invalid-feedback">
+                {{joinStrings ($realm.ErrorsFor "name") ", "}}
+              </div>
+              {{end}}
+            </div>
+          </div>
+
+          <div class="form-group row">
+            <label class="col-sm-3">Allowed reports:</label>
```
sethvargo, line 44: We use "report type" and "test type" interchangeable in most places. We should probably standardize.

pkg/database/realm.go
```diff
@@ -31,6 +32,11 @@ type Realm struct {
 	// Name is the name of the realm.
 	Name string `gorm:"type:varchar(200);unique_index"`
 
+	// DisableLikelyReports and DisableNegativeReports are realm-specific
+	// configurations to change how codes are issued by the realm.
+	DisableLikelyReports   bool `gorm:"type:boolean; not null; default:false"`
```
sethvargo, line 37: This weird inversion is required because Gorm won't set values for default values, and the default value for a boolean is false.

cmd/server/assets/home.html
```diff
@@ -45,22 +48,29 @@ <h1>Issue diagnosis verification code</h1>
...
           <div class="form-check">
             <input class="form-check-input" type="radio" name="testType" id="testType2" value="likely" />
             <label class="form-check-label" for="testType2">Likely diagnosis</label>
             <small class="form-text text-muted">
               Clinical diagnosis without a test
             </small>
           </div>
+          {{end}}
+
+          {{if not $currentRealm.DisableNegativeReports}}
```
mikehelmick, line 62: this could be nested - it doesn't make sense to have negative without likely given the revision model.

cmd/server/assets/realm.html
```diff
@@ -30,7 +31,30 @@ <h1>Realm settings</h1>
...
+            <div class="col-sm-9">
+              <div class="form-check">
+                <input type="checkbox" id="allowLikelyReports" name="allowLikelyReports" class="form-check-input"{{if not $realm.DisableLikelyReports}} checked{{end}}>
+                <label class="form-check-label" for="allowLikelyReports">
+                  Likely
+                </label>
+              </div>
+              <div class="form-check">
+                <input type="checkbox" id="allowNegativeReports" name="allowNegativeReports" class="form-check-input"{{if not $realm.DisableNegativeReports}} checked{{end}}>
```
mikehelmick, line 53: I think this should be more of a radio

Confirmed,
Confirmed + Likely,
Confirmed + Likely + Negative

sethvargo:
localhost_8080_home
localhost_8080_realm_settings

PR #246, opened 2020-08-15
Allow configuring log debug mode

sethvargo:
This is intentionally not done as part of envconfig since we want to process log levels very early.

**Release Note**

```release-note
Allow configuring log level (default to INFO)
```


sethvargo:
I didn't want the env package to have magical configuration

PR #251, opened 2020-08-17
Move realm into another header

sethvargo:
Things were getting crowded for long names.

**Release Note**

```release-note
Update to latest CSS framework version, make current realm more prominent
```


sethvargo:
screenshot-20200817-141318@2x

PR #252, opened 2020-08-17
Fix redirect loop with valid Firebase user, but non-database user

sethvargo:
**Release Note**

```release-note
Fix redirect loop when logging in with a non-existent user
```


sethvargo:
FWIW, this is happening because the user _does_ exist in firebase, so a firebase cookie is set. But the user doesn't exist in our database, so we log out. But the cookie is set, so the login page tries to redirect to home (and loop).

Issue #141, opened 2020-07-28
Enable 2FA on firebase auth


mikehelmick:
https://cloud.google.com/identity-platform/docs/web/mfa#enrolling_a_second_factor

/kind enhancement

mikehelmick:
looks like we need to upgrade from firebase auth to cloud identity platform

sherifkozman:
Wouldn't cloud identity requires a GSuite domain attached to it ?

mikehelmick:
I don’t believe so

sherifkozman:
https://support.google.com/cloudidentity/answer/7319251?hl=en

mikehelmick:
haven't completed testing, but I successfully enabled identity platform on a non gsuite account.

mikehelmick:
Anything left on this one?

PR #270, opened 2020-08-18
Alternative sign-in page using Identity Platform

whaught:

## Proposed Changes

* Create an alternative login page
    * Will Unblock us from using the Identity Platform login w/ multi-factor
* At the moment it's overly verbose with an alert bar at the top. We'll make it pretty later.
* Will someday replace our index.html login

Note: FirebaseUI does not support multi-factor auth. That's a shame because it does all the html for us and it is pretty.

Screenshot from 2020-08-18 14-00-35

**Release Note**

```release-note
Create a new login page without firebaseUI
```

### DIFF
--- a/cmd/server/assets/login.html
+++ b/cmd/server/assets/login.html
@@ -1,0 +1,107 @@
+{{define "login"}}
+<!doctype html>
+<html lang="en">
+
+<head>
+    {{template "head" .}}
+
+    {{template "firebase" .}}
+</head>
+
+<body>
+    {{template "navbar" .}}
+
+    <main role="main" class="container">
+        {{template "flash" .}}
+        <div class="alert alert-warning" role="alert" id="message"></div>
+
+        <div class="card mb-3">
+            <div class="card-body">
+                <form>
+                    {{ .csrfField }}
+                    <div class="row form-group">
+                        <label for="email" class="col-sm-3">Email</label>
+                        <div class="col-sm-9">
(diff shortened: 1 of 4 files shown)


icco:
Is this the same as IAP? I was just reading https://cloud.google.com/load-balancing/docs/negs/serverless-neg-concepts#limitations_with which says cloud run doesn't support IAP.

whaught:
> Is this the same as IAP? I was just reading https://cloud.google.com/load-balancing/docs/negs/serverless-neg-concepts#limitations_with which says cloud run doesn't support IAP.

This is: https://cloud.google.com/identity-platform. I believe IAP uses/depends on this for external identity providers (we're just using email sign-in for now), but we're not using the proxy part

cmd/server/assets/login.html
```diff
@@ -0,0 +1,107 @@
...
+
+        <div class="card mb-3">
+            <div class="card-body">
+                <form>
+                    {{ .csrfField }}
+                    <div class="row form-group">
+                        <label for="email" class="col-sm-3">Email</label>
+                        <div class="col-sm-9">
+                            <div class="input-group p-0 shadow-sm">
+                                <input type="text" id="email" name="email" class="form-control" />
```
sethvargo, line 26: Nit: `type=email` is more accurate and will show the right keyboard on mobile

whaught: Thanks! I'll incorporate into my next

PR #273, opened 2020-08-18
Nit: remove form element

whaught:
## Proposed Changes

* This is just a div table, no need for the form / csrf token. Login action happens from javascript on click, not a form submission


cmd/server/assets/login.html
```diff
@@ -17,32 +17,29 @@
 
         <div class="card mb-3">
             <div class="card-body">
-                <form>
```
mikehelmick, line 20: why is this not a form?

PR #276, opened 2020-08-19
Move login html & fix session

whaught:
## Proposed Changes

* Moves the login and signout files into a folder
* Make the login form type 'email' instead of text
* Duplicate the session serving path for login


cmd/server/assets/login/login.html
```diff
@@ -74,7 +74,7 @@
                     idToken => {
                         $.ajax({
                             type: 'POST',
-                            url: "/session",
+                            url: "login/session",
```
icco, line 77: Is it ok that this isn't an absolute url?

PR #278, opened 2020-08-19
Fix code display in UI and verification checks against HMAC values.

mikehelmick:
## Proposed Changes

* Return plain text verification codes (short/long) to controller instead of hmac version
* On expiration check both plaintext and hmac version when matching for expiration date

**Release Note**

```release-note
BUG FIX - check expiration correctly after verification code HMAC upgrade.
```

### DIFF
--- a/pkg/database/vercode.go
+++ b/pkg/database/vercode.go
@@ -96,7 +96,7 @@ func (db *Database) IsCodeExpired(v *VerificationCode, code string) (bool, error
 	switch {
 	case v.Code == code || v.Code == hmacedCode:
 		return !v.ExpiresAt.After(now), nil
-	case v.LongCode == code || v.Code == hmacedCode:
+	case v.LongCode == code || v.LongCode == hmacedCode:
 		return !v.LongExpiresAt.After(now), nil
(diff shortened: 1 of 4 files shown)


icco:
Hmmm, not sure why that test is failing...

pkg/database/vercode.go
```diff
@@ -83,15 +83,23 @@ func (v *VerificationCode) FormatSymptomDate() string {
 
 // IsCodeExpired checks to see if the actual code provides is the
 // short or long code and deteriminies if it is expired based on that.
-func (v *VerificationCode) IsCodeExpired(code string) bool {
+func (db *Database) IsCodeExpired(v *VerificationCode, code string) (bool, error) {
+	hmacedCode := code
+	if len(code) < 20 {
```
MushuEE, line 88: Should we classify `20` with a name as a global? I am not really sure what `20` means. Is this < "short"?

mikehelmick: it was the previous max plaintext code length

can pull out to constant - or drop this and hmac the hmac value

Issue #149, opened 2020-07-29
MT: Per realm signing keys


mikehelmick:
### TL;DR

The token singing key can be shared by the whole server, as it is for internal consumption.

The certificate signing key should be set up on a per-realm basis.

Ideally we would just create this when the realm is created -added into tooling.
The "server" microservice needs to have signing key creation capabilities

/kind enhancement

sethvargo:
Heh - I was just looking at stuff and thinking "why is this in-band of the request" 😄

mikehelmick:
related to #93

mikehelmick:
done.

filed on followup: https://github.com/google/exposure-notifications-verification-server/issues/349

/close

PR #281, opened 2020-08-19
rename interface from signer to keys

mikehelmick:
Refactor before #149 

## Proposed Changes

* Rename signer package to keys

**Release Note**

```release-note
Potentially breaking: Renamed pkg/signer to pkg/keys and renamed the interface from KeyManager to Manager
```


sethvargo:
@mikehelmick I think this actually needs a release note.

PR #282, opened 2020-08-19
Require all email addresses be verified

whaught:

## Proposed Changes

* Requires that user have verified email address
    * if not, redirects to verification page

**Release Note**

```release-note
Require all users have verified email address to log-in
```


whaught:
/close

shoot this is the wrong thing

PR #283, opened 2020-08-19
pkg/database: deduplicate stats collection code

jeremyfaller:
Refactor out some stats code that was duplicated, and add a unit test.

Fixes: #199

NB: The issue, as described has become outdated. It references deleted
code. This PR only address the duplicated code as it currently stands.

## Proposed Changes

*
*
*

**Release Note**

```release-note

```

Issue #200, opened 2020-08-07
Consider moving stats collection to use Count instead of row returns


crwilcox:
Raised in PR: https://github.com/google/exposure-notifications-verification-server/pull/152/files/91bd8f41c6a0c664a835dd4536655ed3c12a46cb#diff-c72a84bc2625363d546b05336b35a643R144 

It may be better to leverage `count` for some of these queries, as opposed to gathering rows and doing it server-side. Originally this was done for concerns of table lock for count operations, but we have since reduced the frequency of this calculation. 

https://github.com/google/exposure-notifications-verification-server/pull/152/files/91bd8f41c6a0c664a835dd4536655ed3c12a46cb#diff-c72a84bc2625363d546b05336b35a643R144

PR #287, opened 2020-08-19
Calculate stats on save

sethvargo:
- Fixes GH-200
- Closes GH-283

**Release Note**

```release-note
Calculate app and user statistics on save
```


cmd/server/assets/users/show.html
```diff
@@ -77,6 +81,32 @@ <h1>{{$user.Name}}</h1>
   </main>
 
   {{template "scripts" .}}
+
+  {{if $stats}}
+  <script src="https://www.gstatic.com/charts/loader.js"></script>
```
mikehelmick, line 86: push this JS out to a sub-template?

sethvargo: The only JS that's "common" is the URL for loader.js, and it didn't seem worth adding that layer of indirection.

mikehelmick: scrolling back and forth, they looked more similar to me :)

PR #288, opened 2020-08-19
Remove the existing index page

whaught:

## Proposed Changes

* Removes the existing login page and replaces it with our custom one
* Check for auth for the email verification page, but don't check verified
     * This allows the sign-out button to render properly

**Release Note**

```release-note
Remove the firebaseUI based login page
```


cmd/server/assets/header.html
```diff
@@ -41,7 +41,7 @@
         <ul class="navbar-nav mr-auto">
           {{if .currentRealm}}
           <li class="nav-item">
-            <a class="nav-link" href="/home">Issue code</a>
+            <a class="nav-link" href="/header">Issue code</a>
```
sethvargo, line 44: I think this is still /home?

pkg/controller/middleware/auth.go
```diff
@@ -36,6 +36,21 @@ import (
 // is set in the template map. It fetches a user from the session and stores the
 // full record in the request context.
 func RequireAuth(ctx context.Context, client *auth.Client, db *database.Database, h *render.Renderer, ttl time.Duration) mux.MiddlewareFunc {
+	return requireAuth(ctx, client, db, h, ttl, true /*requireVerified*/)
+}
+
+// RequireAuthNotVerified requires a user to be logged in. Does not require the user to have verified the login email.
+func RequireAuthNotVerified(ctx context.Context, client *auth.Client, db *database.Database, h *render.Renderer, ttl time.Duration) mux.MiddlewareFunc {
+	return requireAuth(ctx, client, db, h, ttl, false /*requireVerified*/)
```
sethvargo, line 44: Middleware-wise, it's better to say this requires `RequireAuth` to run first, then do something like

```golang
user := controller.UserFromContext(ctx)
if user == nil {
  // unauth, probably an error
}

// Logic as below.
client.GetUserByEmail(ctx, user.email)
```

cmd/server/assets/header.html
whaught: oops.

pkg/controller/middleware/auth.go
whaught: I could do that and have something like RequireEmailVerified, but the reason I have not is that I want checking both to be the default (so it can't be forgotten on pages as we add them).

The email verify page is special in that it requires auth, but not verified

sethvargo: Fair. I'm just really worried about having a method with a magical boolean that changes the flow logic.

PR #293, opened 2020-08-20
Move the error flash insertion to a javascript function

whaught:
## Proposed Changes

* Move the error flash insertion to a javascript function

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -18,6 +18,7 @@
   div.realm-header {
     background-color: #0055b1;
   }
+
 </style>
 {{end}}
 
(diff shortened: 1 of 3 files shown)


cmd/server/assets/header.html
```diff
@@ -208,5 +209,28 @@ <h6 class="dropdown-header">Actions</h6>
...
     })
   });
+
+  var $main = $('main');
+
+  function clearExistingAlerts() {
+    $main.find('div.alert').remove();
+  }
+
+  function alert(message, errLevel) {
```
sethvargo, line 219: `alert` is a javascript built-in technically it's window.alert, but it also exists in the default namespace. Maybe name this `flash`?

whaught: done

PR #300, opened 2020-08-21
Update code issuing to match mocks

sethvargo:
**Release Note**

```release-note
Enhancements to code issue page and flow
```

/assign @whaught 
/assign @mikehelmick


cmd/server/assets/login/login.html
```diff
@@ -73,7 +73,6 @@
           return
         }
 
-        flash("Welcome, " + user.email, "success")
```
whaught, line 76: but the welcome flash gives you feedback while the ajax processes!

sethvargo:
screenshot-20200820-224024@2x
screenshot-20200820-224059@2x

cmd/server/assets/login/login.html
sethvargo: I found it jarring. Like - I saw a flash of green and then it moved on. Did I miss something? Was I supposed to do something?

whaught: Fair. Possibly connection-speed dependant. Less is more, lets remove it.

sethvargo: Yea, maybe we can do a setTimeout on it? That way it only appears if you are waiting 1s+ or something?

pkg/api/api.go
```diff
@@ -142,6 +147,10 @@ type CheckCodeStatusResponse struct {
 	// After this time the code will no longer be accepted and is eligible for deletion.
 	ExpiresAtTimestamp int64 `json:"expiresAtTimestamp"`
 
+	// LongExpiresAtTimestamp repesents the time when the long code expires, in
```
whaught, line 150: is the "long code" different from the code? Still a little confused about how this differs from ExpiresAtTimestamp

cmd/server/assets/login/login.html
whaught: Sometimes the redirect is the slow part. The intent was some kind of positive feedback while you wait for that second page.

pkg/api/api.go
sethvargo: ExpiresAt is for the short code (the 8-12 digit one on-screen). Long.* is for the longer deep-link SMS codes that are sent to the user (and never displayed).

whaught: ah. I think I got confused thinking 'Long' mean int64

whaught:
/hold
/lgtm

Good stuff. Feel free to unhold if you're ready to merge.

pkg/api/api.go
```diff
@@ -142,6 +147,10 @@ type CheckCodeStatusResponse struct {
...
+	// UTC seconds since epoch.
+	LongExpiresAtTimestamp int64 `json:"longExpiresAtTimestamp,omitempty"`
```
mikehelmick, line 152: thanks - I missed that!

PR #302, opened 2020-08-21
Page for creating a new user

whaught:
## Proposed Changes

* Allows anyone to create a new User for authentication
    * If that user doesn't also exist in our custom users table, it results in a hilarious redirect loop

**Release Note**

```release-note
Create a custom create auth page
```


cmd/server/assets/login/login.html
```diff
@@ -32,7 +32,8 @@
                 <button type="submit" id="submit" class="btn btn-primary btn-block">Login</button>
               </form>
               <div class="card-body">
-                <a class="card-link" href="legacylogin">New user</a>
+                <a class="card-link" href="login/create">New user</a>
```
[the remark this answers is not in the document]
whaught: I think we should go with this for now so we can kill the firebaseui one. This one logs-in to session correctly and doesn't ask for real-name.

cmd/server/assets/login/create.html
```diff
@@ -0,0 +1,97 @@
...
+
+<body class="bg-light">
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <div class="d-flex vh-100">
+      <div class="d-flex w-100 justify-content-center align-self-center">
+        <div class="col-sm-6">
+          <div class="card shadow-sm">
+            <div class="card-header">Create New User</div>
```
sethvargo, line 19: Nit: "Create new user"

cmd/server/assets/login/create.html
```diff
@@ -0,0 +1,97 @@
...
+
+  <script>
+    $(function () {
+      let $form = $('#loginForm');
+      let $submit = $('#submit');
+      let $email = $('#email');
+      let $password = $('#password');
+      let $retype = $('#retype');
+
+      let auth = firebase.auth()
```
sethvargo, line 58: sc

cmd/server/assets/login/create.html
```diff
@@ -0,0 +1,97 @@
...
+
+      $form.on('submit', function (event) {
+        event.preventDefault();
+
+        let email = $email.val();
+        let pwd = $password.val();
+
+        if (pwd != $retype.val()) {
+          flash("Password and retyped passwords must match", "danger")
```
sethvargo, line 67: sc here and after return

PR #324, opened 2020-08-22
Ignore Terraform override files.

yegle:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@
 *.terraform/
 *.tfvars
 terraform/state.tf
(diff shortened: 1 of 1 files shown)


yegle:
Is there a supported way to migrate the state from using the terraform files directly to using the files as module? Attempting to just copy the tfstate file over doesn't seem to work (some weird authentication issue upon `terraform refresh`).

sethvargo:
You need to use the `terraform state mv` command

yegle:
Awesome! Thanks Seth for the help!

PR #325, opened 2020-08-22
 Add support for `IGNORE_UNCLEAN_REPO` env var in build script.

yegle:
This is to bypass the check of repo cleanliness for local development.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -19,7 +19,7 @@ set -eEuo pipefail
 ROOT="$(cd "$(dirname "$0")/.." &>/dev/null; pwd -P)"
 ALL_SERVICES="apiserver,cleanup,server,adminapi"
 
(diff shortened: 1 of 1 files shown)


yegle:
Sometimes I build the binary under `tools/` and run the binary directly instead of `go run xxx.go`. This leaves an untracked file in the repo.

Not super critical. I'll just switch to use `go run`.

icco:
> Sometimes I build the binary under `tools/` and run the binary directly instead of `go run xxx.go`. This leaves an untracked file in the repo.
> 
> Not super critical. I'll just switch to use `go run`.

You can go run a directory. Also those binaries have predictable names, just add them to the gitignore.

yegle:
OK, I sent #347 which is less controversy. I'm going to close this PR. Thanks!

PR #329, opened 2020-08-23
Switch to JSON encoding for cacher, decouple cacher from db

sethvargo:
Switch to JSON encoding for the cacher. Gob is faster, but misconfigurations turn into runtime errors. It also decouples the cacher from the database.

This also introduces a KeyFunc type on the cacher to mutate keys before they are stored (e.g. to obfuscate or encrypt them), and fixes the SHA calculations on the rate limiter keys.

**Release Note**

```release-note
Switch to JSON encoding for cacher, decouple cacher from database
```

/assign @mikehelmick

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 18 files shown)


mikehelmick:
check test failures

PR #340, opened 2020-08-24
Use local time zone when issuing a code.

### LINKED ISSUE #63: Better handling of test date for different locales
### TL;DR

Currently the test dates are all captured and stored at UTC.

It would be nice to let the epi specify the test date in the local time zone, which can be conveyed back to the user through the app, etc.

jeremyfaller:
Rather than requiring the user to use UTC when requesting a code, use
the client's local timezone.

Fixes #63

## Proposed Changes

* Use client's timezone when requesting a code
* Requires a "reasonably valid" TZ on the client.

**Release Note**

```release-note
action required: client timezone should be valid.
```


pkg/api/api.go
```diff
@@ -102,6 +102,7 @@ type IssueCodeRequest struct {
 	SymptomDate string `json:"symptomDate"` // ISO 8601 formatted date, YYYY-MM-DD
 	TestDate    string `json:"testDate"`
 	TestType    string `json:"testType"`
+	TZOffset    int    `json:"tzOffset"` // offset in minutes of the user's timezone.
```
mikehelmick, line 105: Can we document what happens if this is omitted? i.e. you're speaking UTC.
And also I think direction matters here for documentation. What sign is required.

like we're on the negative side on UTC, but if I add a negative TZ offset, The logic in issue.go will put us further away

mikehelmick:
/hold

blocking seth's lgtm, sorry

pkg/api/api.go
jeremyfaller: I believe we're all set. I've added unit tests to help validate. LMK.

mikehelmick:
check presubmit failure

PR #348, opened 2020-08-25
implement token signing key rotation

### LINKED ISSUE #94: Token key rotation
Implement key rotation for the verification tokens. These tokens/keys are only used within this server, so no external discovery mechanism is needed.

/kind enhancement

mikehelmick:
Fixes #94

## Proposed Changes

* Allow the token signing key to be rotated by the server operator. This is totally independent from the certificate signing keys

**Release Note**

```release-note
Verification server operators can rotate their token signing key. `TOKEN_SIGNING_KEY` and `TOKEN_SIGNING_KEY_ID` are now array based env vars. They must be the same length. The first items in the lists represents the active key/kid and the remaining entries are allowed to validate.
```

### DIFF
--- a/pkg/config/api_server_config.go
+++ b/pkg/config/api_server_config.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/cache"
@@ -42,7 +43,6 @@ type APIServerConfig struct {
 	APIKeyCacheDuration time.Duration `env:"API_KEY_CACHE_DURATION,default=5m"`
 
 	// Verification Token Config
-	// Currently this does not easily support rotation. TODO(mikehelmick) - add support.
 	VerificationTokenDuration time.Duration `env:"VERIFICATION_TOKEN_DURATION,default=24h"`
 
 	// Token signing
@@ -53,6 +53,10 @@ type APIServerConfig struct {
 
 	// Rate limiting configuration
 	RateLimit ratelimit.Config
+
+	// cached allowed public keys
+	allowedTokenPublicKeys map[string]string
+	mu                     sync.RWMutex
 }
 
 // NewAPIServerConfig returns the environment config for the API server.
@@ -63,6 +67,34 @@ func NewAPIServerConfig(ctx context.Context) (*APIServerConfig, error) {
 		return nil, err
 	}
 	return &config, nil
+}
+
+// AllowedTokenPublicKeys returns a map of 'kid' to the KMS KeyID reference.
+// This represents the keys that are allowed to be used to verify tokens,
+// the TokenSigningKey/TokenSigningKeyID.
+func (c *APIServerConfig) AllowedTokenPublicKeys() map[string]string {
+	{
+		c.mu.RLock()
+		if len(c.allowedTokenPublicKeys) != 0 {
+			c.mu.RUnlock()
+			return c.allowedTokenPublicKeys
+		}
+		c.mu.RUnlock()
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// handle race condition that could occur between lock upgrade.
+	if len(c.allowedTokenPublicKeys) != 0 {
+		return c.allowedTokenPublicKeys
+	}
+
+	c.allowedTokenPublicKeys = make(map[string]string)
+
+	for i, kid := range c.TokenSigning.TokenSigningKeyID {
+		c.allowedTokenPublicKeys[kid] = c.TokenSigning.TokenSigningKey[i]
+	}
+	return c.allowedTokenPublicKeys
 }
 
 func (c *APIServerConfig) Validate() error {
@@ -77,6 +109,10 @@ func (c *APIServerConfig) Validate() error {
 		if err := checkPositiveDuration(f.Var, f.Name); err != nil {
 			return err
 		}
+	}
+
+	if err := c.TokenSigning.Validate(); err != nil {
+		return err
 	}
 
 	return nil
--- a/pkg/config/token_signing_config.go
+++ b/pkg/config/token_signing_config.go
@@ -15,6 +15,8 @@
 package config
 
 import (
+	"fmt"
+
 	"github.com/google/exposure-notifications-server/pkg/keys"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/config/api_server_config.go
```diff
@@ -40,17 +41,24 @@ type APIServerConfig struct {
 	APIKeyCacheDuration time.Duration `env:"API_KEY_CACHE_DURATION,default=5m"`
 
 	// Verification Token Config
-	// Currently this does not easily support rotation. TODO(mikehelmick) - add support.
 	VerificationTokenDuration time.Duration `env:"VERIFICATION_TOKEN_DURATION,default=24h"`
 	TokenSigningKey           string        `env:"TOKEN_SIGNING_KEY,required"`
 	TokenSigningKeyID         string        `env:"TOKEN_SIGNING_KEY_ID,default=v1"`
 	TokenIssuer               string        `env:"TOKEN_ISSUER,default=diagnosis-verification-example"`
+	// In order to rotate keys, one sets these values to the 'old' key ID allowing it to still verify
```
sethvargo, line 48: This backs us into a world where there can only ever be two valid keys. While I understand that's our _current_ situation, given the frequency of change in this project, I wonder if it makes sense to have an array instead where the "first" element is the newest key and all other ones are "old"

mikehelmick: then you have to have parallel arrays w/ kid and key ref.

The other option would be to bring this up one level where we take a cryptoKey and allow all non-destroyed versions for validation and the primary for signing.
But - I then we have to support that across all our KMS implementations

Needing more than 2 valid keys means that you're rotating twice in a 24 hour period which hopefully isn't ever necessary.

If we want this to be more flexible, I think we can push it into the database and manage it like the per-realm signing keys.

We do need to have a general 'admin' page in the UI still anyway, to allow for realm creation.

sethvargo: Up to you. I just know the moment we do this, we'll need to support a third...

mikehelmick: will switch to arrays in the config

mikehelmick:
/unhold

rebased + moved to array based implementation

pkg/config/api_server_config.go
```diff
@@ -79,6 +111,10 @@ func (c *APIServerConfig) Validate() error {
 		}
 	}
 
+	if err := c.TokenSigning.Validate(); err != nil {
+		return err
```
sethvargo, line 115:
```suggestion
		return fmt.Errorf("failed to validate signing tokens: %w", err)
```

pkg/config/token_signing_config.go
```diff
@@ -25,7 +27,22 @@ type TokenSigningConfig struct {
 	// configuration.
 	Keys keys.Config `env:",prefix=TOKEN_"`
 
-	TokenSigningKey   string `env:"TOKEN_SIGNING_KEY, required"`
-	TokenSigningKeyID string `env:"TOKEN_SIGNING_KEY_ID, default=v1"`
-	TokenIssuer       string `env:"TOKEN_ISSUER, default=diagnosis-verification-example"`
+	TokenSigningKey   []string `env:"TOKEN_SIGNING_KEY, required"`
+	TokenSigningKeyID []string `env:"TOKEN_SIGNING_KEY_ID, default=v1"`
```
sethvargo, line 31: Key**s**? and ID**S**?

pkg/config/token_signing_config.go
```diff
@@ -25,7 +27,22 @@ type TokenSigningConfig struct {
 	// configuration.
 	Keys keys.Config `env:",prefix=TOKEN_"`
 
-	TokenSigningKey   string `env:"TOKEN_SIGNING_KEY, required"`
-	TokenSigningKeyID string `env:"TOKEN_SIGNING_KEY_ID, default=v1"`
-	TokenIssuer       string `env:"TOKEN_ISSUER, default=diagnosis-verification-example"`
+	TokenSigningKey   []string `env:"TOKEN_SIGNING_KEY, required"`
```
sethvargo, line 30: Probably doc this:

- Slice of keys, first key is active, other keys are still accepted

pkg/config/token_signing_config.go
```diff
@@ -25,7 +27,22 @@ type TokenSigningConfig struct {
...
+	TokenSigningKeyID []string `env:"TOKEN_SIGNING_KEY_ID, default=v1"`
+	TokenIssuer       string   `env:"TOKEN_ISSUER, default=diagnosis-verification-example"`
+}
+
+func (t *TokenSigningConfig) ActiveKey() string {
+	return t.TokenSigningKey[0]
```
sethvargo, line 36: This will panic if TokenSigningKey is empty

pkg/config/token_signing_config.go
```diff
@@ -25,7 +27,22 @@ type TokenSigningConfig struct {
...
+}
+
+func (t *TokenSigningConfig) ActiveKeyID() string {
+	return t.TokenSigningKeyID[0]
```
sethvargo, line 40: Panic

pkg/config/token_signing_config.go
```diff
@@ -25,7 +27,22 @@ type TokenSigningConfig struct {
...
+}
+
+func (t *TokenSigningConfig) Validate() error {
+	if len(t.TokenSigningKey) != len(t.TokenSigningKeyID) {
+		return fmt.Errorf("TOKEN_SIGNING_KEY and TOKEN_SIGNING_KEY_ID must be lists of the same length")
```
sethvargo, line 45:
```go
keys, ids := len(t.TokenSigningKey), len(t.TokenSigningKeyID); keys != ids {
  return fmt.Errorf("TOKEN_SIGNING_KEY (%d) and TOKEN_SIGNING_KEY_ID (%d) must be the same length", keys, ids)
}
```

pkg/controller/certapi/certapi.go
```diff
@@ -78,19 +78,21 @@ func New(ctx context.Context, config *config.APIServerConfig, db *database.Datab
 }
 
 // Parses and validates the token against the configured keyID and public key.
-// If the token si valid the token id (`tid') and subject (`sub`) claims are returned.
-func (c *Controller) validateToken(ctx context.Context, verToken string, publicKey crypto.PublicKey) (string, *database.Subject, error) {
+// A map of valid 'kid' values is supported.
+// If the token is valid the token id (`tid') and subject (`sub`) claims are returned.
+func (c *Controller) validateToken(ctx context.Context, verToken string, publicKeys map[string]crypto.PublicKey) (string, *database.Subject, error) {
```
sethvargo, line 83: This is probably fine for now, but this looks awfully like the publicKeyCache

pkg/controller/certapi/certificate.go
```diff
@@ -58,12 +59,15 @@ func (c *Controller) HandleCertificate() http.Handler {
 		stats.Record(ctx, c.metrics.Attempts.M(1))
 
 		// Get the public key for the token.
-		publicKey, err := c.pubKeyCache.GetPublicKey(ctx, c.config.TokenSigning.TokenSigningKey, c.kms)
-		if err != nil {
-			c.logger.Errorw("failed to get public key", "error", err)
-			stats.Record(ctx, c.metrics.CertificateErrors.M(1))
-			c.h.RenderJSON(w, http.StatusInternalServerError, api.InternalError())
-			return
+		allowedPublicKeys := make(map[string]crypto.PublicKey)
```
sethvargo, line 62:
```suggestion
		allowedPublicKeys := make(map[string]crypto.PublicKey, len(c.config.AllowedTokenPublicKeys()))
```

pkg/config/api_server_config.go
```diff
@@ -65,6 +69,34 @@ func NewAPIServerConfig(ctx context.Context) (*APIServerConfig, error) {
 	return &config, nil
 }
 
+// AllowedTokenPublicKeys returns a map of 'kid' to the KMS KeyID reference.
+// This represents the keys that are allowed to be used to verify tokens,
+// the TokenSigningKey/TokenSigningKeyID.
+func (c *APIServerConfig) AllowedTokenPublicKeys() map[string]string {
+	{
```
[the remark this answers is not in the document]
mikehelmick: visual scoping of the read lock section

pkg/config/api_server_config.go
```diff
@@ -65,6 +69,34 @@ func NewAPIServerConfig(ctx context.Context) (*APIServerConfig, error) {
...
+		c.mu.RLock()
+		if len(c.allowedTokenPublicKeys) != 0 {
```
jeremyfaller, line 78: The RLock is not necessarily released. If, for example, c=nil, you panic, and don't release the lock. More robust is:

func() {
  c.mu.RLock()
  defer c.mu.RUnlock()
  if .....
}()

jeremyfaller:
Shoot, I might have inadvertently caused an early merge. My mistake. I will clean up if Mike's too busy. (Sorry, for the noob mistake.)

pkg/controller/certapi/certapi.go
mikehelmick: it's populated from the public key cache - see certificate.go, line 63

PR #369, opened 2020-08-25
Automatically clear caches on save

sethvargo:
**Release Note**

```release-note
NONE
```


mikehelmick:
LGTM modulo build errors - concept is right

PR #370, opened 2020-08-25
Use floating form labels for login / check status

whaught:
## Proposed Changes

* Use floating labels on forms so the user can continue to see placeholders as they type

Adapted from:
https://getbootstrap.com/docs/4.0/examples/floating-labels/

floatinglabel

### DIFF
--- a/cmd/server/assets/codestatus/index.html
+++ b/cmd/server/assets/codestatus/index.html
@@ -1,5 +1,7 @@
 {{define "code/status"}}
 
+{{$code := .code}}
+
 <!doctype html>
 <html lang="en">
 
@@ -7,7 +9,7 @@
   {{template "head" .}}
 </head>
 
-<body>
+<body class="bg-light">
   {{template "navbar" .}}
 
   <main role="main" class="container">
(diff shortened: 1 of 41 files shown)


cmd/server/assets/codestatus/index.html
```diff
@@ -21,15 +22,16 @@ <h1>Verification code status</h1>
...
           {{ .csrfField }}
-          <div class="row form-group">
-            <label for="uuid" class="col-sm-3">Code identifier</label>
-            <div class="col-sm-9">
+          <div class="form-group">
+            <div class="form-label-group ">
               <input type="text" id="uuid" name="uuid" class="form-control text-monospace{{if $code.ErrorsFor "uuid"}} is-invalid{{end}}" value="{{$code.UUID}}"
                 placeholder="xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
                 autocomplete="off">
+                <label for="uuid">xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx</label>
```
sethvargo, line 34: I think we want the placeholder to be XXXXXX..., but the floating label should be "UUID" or something

cmd/server/assets/home.html
```diff
@@ -402,6 +401,79 @@ <h1>Create verification code</h1>
       // Show reset button
       $buttonReset.removeClass('d-none');
     }
+
```
sethvargo, line 404: I think this is a bad merge?

whaught: sorry this file shouldn't be here

cmd/server/assets/codestatus/index.html
whaught: I think the two should be the same (whatever that is). It starts with placeholder text and smoothly moves up as you type. Ideally it would be UUID and we'd have some AsYouType formatting going on

sethvargo: Ah I see. I missed that it was an animation.

PR #371, opened 2020-08-25
Sort things

sethvargo:
Sort users and API keys

**Release Note**

```release-note
Sort users and API keys
```


sethvargo:
/hold 

Until after release

PR #373, opened 2020-08-26
[WIP] Allow users to invalidate codes (ajax)

whaught:
## Proposed Changes

* Adds a button with ajax to an invalidate endpoint that sets the code to expire now

bigbutt

**Release Note**

```release-note
Allow users to expire a code
```


pkg/controller/codestatus/expire.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (c *Controller) HandleExpire() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var request api.ExpireCodeRequest
+		if err := controller.BindJSON(w, r, &request); err != nil {
+			c.h.RenderJSON(w, http.StatusBadRequest, api.Error(err))
+			return
+		}
+
+		code, errCode, err := c.CheckCodeStatus(r, request.UUID)
```
sethvargo, line 34: Does this verify that the current user has permission to do this to this code?

pkg/api/api.go
```diff
@@ -155,6 +155,27 @@ type CheckCodeStatusResponse struct {
...
+type ExpireCodeResponse struct {
+	// ExpiresAtTimestamp represents Unix, seconds since the epoch. Still UTC.
+	// After this time the code will no longer be accepted and is eligible for deletion.
+	ExpiresAtTimestamp int64 `json:"expiresAtTimestamp"`
+
+	// LongExpiresAtTimestamp repesents the time when the long code expires, in
+	// UTC seconds since epoch.
+	LongExpiresAtTimestamp int64 `json:"longExpiresAtTimestamp,omitempty"`
+
+	Error     string `json:"error"`
```
sethvargo, line 175: We should `omitempty` on this too

cmd/server/assets/codestatus/show.html
```diff
@@ -18,21 +18,24 @@ <h1>Verification code status</h1>
...
+        <div class="list-group-item">
           <h5 class="mb-1">Status</h5>
           <p class="mb-1">{{.status}}</p>
         </div>
-        <div class="list-group">
-          <h5 class="mb-1">Expiry</h5>
+        <div class="list-group-item">
+          <div class="d-flex w-100 justify-content-between">
+            <h5 class="mb-1">Expiry</h5>
+            <button type="button" id="invalidate" class="btn btn-primary btn-sm">Invalidate code now</button>
```
sethvargo, line 37: I kinda think this button should be below the form and definitly have the btn-danger class

pkg/controller/codestatus/expire.go
whaught: Yes CheckCodeStatus does all that heavy-lifting. Specifically it checks that the user or app is the issuing user or app

pkg/api/api.go
```diff
@@ -151,7 +151,28 @@ type CheckCodeStatusResponse struct {
...
 	LongExpiresAtTimestamp int64 `json:"longExpiresAtTimestamp,omitempty"`
 
-	Error     string `json:"error"`
+	Error     string `json:"error,omitempty"`
+	ErrorCode string `json:"errorCode,omitempty"`
+}
+
+// ExpireCodeRequest defines the parameters to request that a code be expired now.
+// This is called by the Web frontend.
+type ExpireCodeRequest struct {
```
mikehelmick, line 160: we should also sever this on the admin api as well

pkg/controller/codestatus/expire.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
+			c.h.RenderJSON(w, errCode, err)
+			return
+		}
+
+		c.db.ExpireCode(code)
```
mikehelmick, line 40: feels like this should return an error - there's a race condition between expiring and claiming or 2 users expiring, etc

pkg/database/vercode.go
```diff
@@ -207,6 +207,13 @@ func (db *Database) SaveVerificationCode(vc *VerificationCode, maxAge time.Durat
 	return db.db.Save(vc).Error
 }
 
+// ExpireCode saves a verification code as expired.
+func (db *Database) ExpireCode(vc *VerificationCode) error {
```
mikehelmick, line 211: this should be in a transaction and to a for update select and make sure it makes sense to actually invalidate the code

pkg/api/api.go
whaught: that will be the next thing

pkg/database/vercode.go
whaught: excellent point. I think I got excited to see this work at all...

pkg/api/api.go
whaught: jk it's a now thing. Added that endpoint and a db unit test

cmd/server/assets/codestatus/index.html
```diff
@@ -21,16 +21,17 @@ <h1>Verification code status</h1>
...
         <form method="POST" action="show" class="floating-form">
           {{ .csrfField }}
           <div class="form-group">
             <div class="form-label-group ">
-              <input type="text" id="uuid" name="uuid" class='form-control text-monospace{{if $code.ErrorsFor "uuid"}} is-invalid{{end}}' value="{{$code.UUID}}"
-                placeholder="UUID" maxlength="36"autocomplete="off">
-                <label for="uuid">UUID</label>
+              <input type="text" id="uuid" name="uuid"
+                class='form-control text-monospace{{if $code.ErrorsFor "uuid"}} is-invalid{{end}}'
+                value="{{$code.UUID}}" placeholder="UUID" maxlength="36" autocomplete="off">
```
sethvargo, line 33: Can we drop the max-length while we're in here. I regularly lose hours of my life to fields that silently cut off my extra characters...

cmd/server/assets/codestatus/show.html
```diff
@@ -18,45 +18,88 @@ <h1>Verification code status</h1>
...
   </main>
 
   {{template "scripts" .}}
   {{template "codescripts" .}}
 
   <script type="text/javascript">
+    let $buttonInvalidate = $('button#invalidate');
     let $codeExpiresAt;
     let expires = {{.expires }};
+    let uuid = {{.uuid }}
```
sethvargo, line 54: sc

cmd/server/assets/codestatus/show.html
```diff
@@ -18,45 +18,88 @@ <h1>Verification code status</h1>
...
+    let codeCountdown
```
sethvargo, line 55: sc

pkg/database/vercode.go
```diff
@@ -195,6 +195,26 @@ func (db *Database) FindVerificationCodeByUUID(uuid string) (*VerificationCode,
 	return &vc, nil
 }
 
+// ExpireCode saves a verification code as expired.
+func (db *Database) ExpireCode(uuid string) (*VerificationCode, error) {
+	var vc VerificationCode
+	err := db.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Where("uuid = ?", uuid).Find(&vc).Error; err != nil {
```
sethvargo, line 202: I think you need a `tx.Set("gorm:query_option", "FOR UPDATE")` here here

cmd/server/assets/codestatus/index.html
whaught: wait but literally a UUID should not be longer. It's weird if you keep typing

cmd/server/assets/codestatus/show.html
whaught: switching between go and javascript is killin me

cmd/server/assets/codestatus/index.html
whaught: to elaborate: this field has as-you-type javascript that molds your input to UUID valid characters, so the length makes sense (you'll lose non-guid input already)

whaught: okay you're right. The javascript already does all we need. This could lose data, eg. pasting with leading whitespace

cmd/server/assets/codestatus/index.html
```diff
@@ -21,16 +21,17 @@ <h1>Verification code status</h1>
       Use an identifier to check the status of a code previously shared with your patient.
     </p>
 
-    <div class="card mb-3">
+    <div class="card mb-3" style="max-width: 30rem">
```
sethvargo, line 24: Why max this?

cmd/server/assets/codestatus/show.html
```diff
@@ -18,45 +18,88 @@ <h1>Verification code status</h1>
...
+        // Disable the submit button so we only issue one code
+        $buttonInvalidate.prop('disabled', true);
+        let data = {
+          uuid: uuid
+        };
+        invalidateCode(data);
+      });
     });
+
+    function invalidateCode(data) {
```
sethvargo, line 75: I'm not convinced all this javascript is necessary. Could we just have the button be a regular form/link and use old-school HTML and forms?

whaught: We could, but I was worried that would be a worse experience with the submit page ultimately redirecting back to here. Without ajax there would be a funny page-refresh (but it does eliminate the timer reset js and such).

cmd/server/assets/codestatus/index.html
whaught: the list looks kind of weird as it gets huge with the whitespace. Originally I had the button floating right, which looked bad if it was too wide. I'd rather center the card div or something.

That said I'm open to aesthetic tweaks

cmd/server/assets/codestatus/show.html
sethvargo: The Internet is pretty fast, and I think the simplicity it gets us is well worth the UX tradeoff.

mikehelmick: agree the form is simpler to maintain - if you want to check this in and change it later, up to ou.

whaught: /hold

Nah let's hold this. I'll look at a form.
Pulled out the API: https://github.com/google/exposure-notifications-verification-server/pull/380 and I'll focus on the UX separately

pkg/database/vercode.go
```diff
@@ -195,6 +195,16 @@ func (db *Database) FindVerificationCodeByUUID(uuid string) (*VerificationCode,
...
 }
 
+// ExpireCode saves a verification code as expired.
+func (db *Database) ExpireCode(uuid string) (time.Time, error) {
+	now := time.Now().UTC()
+	vc := VerificationCode{
+		ExpiresAt:     now,
+		LongExpiresAt: now,
+	}
+	return now, db.db.Model(&vc).Where("uuid = ?", uuid).Debug().Update(&vc).Error
```
mikehelmick, line 205: .Debug?

also - I still think should make sure it makes sense to expire the code in a Tx:
like: https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/database/token.go#L169

the code shouldn't have already been claimed and it shouldn't be already expired.

mikehelmick: you can also check those w/ a single SQL statement

PR #378, opened 2020-08-26
Add terraform module for smoke test terraform configuration

chaodaiG:
This new module contains some random prefix/suffix for GCP resources that can not be deleted by terraform, so that the same GCP project can be used for smoke testing terraform deployment repeatedly

/assign @sethvargo 
/cc @yegle


chaodaiG:
@sethvargo , please take a look

PR #380, opened 2020-08-26
Add code expire endpoint

whaught:
## Proposed Changes

* Add the expire code API to allow users to invalidate a code (set expiry time to now)
    * Spinning the logic out of: https://github.com/google/exposure-notifications-verification-server/pull/373 so I can focus on UX separately

### DIFF
--- a/pkg/database/vercode.go
+++ b/pkg/database/vercode.go
@@ -202,7 +202,7 @@ func (db *Database) ExpireCode(uuid string) (time.Time, error) {
 		ExpiresAt:     now,
 		LongExpiresAt: now,
 	}
-	return now, db.db.Model(&vc).Where("uuid = ?", uuid).Debug().Update(&vc).Error
+	return now, db.db.Model(&vc).Where("uuid = ?", uuid).Update(&vc).Error
 }
 
 // SaveVerificationCode created or updates a verification code in the database.
--- a/pkg/database/vercode.go
+++ b/pkg/database/vercode.go
@@ -196,26 +196,13 @@ func (db *Database) FindVerificationCodeByUUID(uuid string) (*VerificationCode, 
 }
 
 // ExpireCode saves a verification code as expired.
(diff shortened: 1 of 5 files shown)


pkg/database/vercode.go
```diff
@@ -195,6 +195,29 @@ func (db *Database) FindVerificationCodeByUUID(uuid string) (*VerificationCode,
 	return &vc, nil
 }
 
+// ExpireCode saves a verification code as expired.
+func (db *Database) ExpireCode(uuid string) (*VerificationCode, error) {
+	var vc VerificationCode
+	err := db.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.
```
sethvargo, line 202: Looking at this again, why do we attempt to find the record? Why not just:

```go
now := time.Now().UTC()
db.Table("verification_codes").Where("uuid = ?").Update(VerificationCode{
  ExpiresAt: now,
  LongExpiresAt: now,
}).Error
```

That avoids the need to take out a lock.

whaught: See smaller PRs does work. I'm used to spanner where blind-writes are disallowed.

Nice!

pkg/database/vercode.go
```diff
@@ -195,6 +195,16 @@ func (db *Database) FindVerificationCodeByUUID(uuid string) (*VerificationCode,
...
 }
 
+// ExpireCode saves a verification code as expired.
+func (db *Database) ExpireCode(uuid string) (time.Time, error) {
+	now := time.Now().UTC()
+	vc := VerificationCode{
+		ExpiresAt:     now,
+		LongExpiresAt: now,
+	}
+	return now, db.db.Model(&vc).Where("uuid = ?", uuid).Debug().Update(&vc).Error
```
mikehelmick, line 205: same comment I put on the other review.

Needs to check that it makes sense to actually invalidate the code
* hasn't been claimed
* hasn't already expired

PR #385, opened 2020-08-26
Add jwks endpoint to return realm/key JSON encoded keys

### LINKED ISSUE #93: Implement JWKS discovery for verification certificate keys
### TL;DR

In order to facilitate key rotation and automatic discovery by an exposure notifications server, implement a JWKS document endpoint that provides the current public key material.

### Design

**Proposal**

Needs to be a well known - unauthenticated endpoint for retrieval.

the KIDs need to be encoded in such a way that they can convey multi-tenancy of the verification server.

For example, a kid might look like "us-wa-00001" 

See example JWK here -> https://tools.ietf.org/html/rfc7517#section-3

/kind enhancement
/assign

jeremyfaller:
Fixes #93

## Proposed Changes

* Add endpoint for jwks encded keys.
* Fix typo.

**Release Note**

```release-note
NONE

```

### DIFF
--- a/cmd/server/assets/realmkeys.html
+++ b/cmd/server/assets/realmkeys.html
@@ -88,7 +88,8 @@
                 {{$publicKeys := .publicKeys}}
                 {{range $rk := .realmKeys}}
                 <tr>
-                  <td>{{$rk.GetKID}}
+                  <td>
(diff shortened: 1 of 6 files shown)


jeremyfaller:
Actually, hold off on this PR. I forgot to add caching.  BRB.

Issue #381, opened 2020-08-26
Add HSTS middleware


icco:
Related to https://github.com/google/exposure-notifications-verification-server/issues/362, we should add HSTS headers to our responses to make sure browsers are only using HTTPS, because our tools won't work over HTTP.

icco:
Thanks!

PR #387, opened 2020-08-26
Add middleware for setting HSTS headers

sethvargo:
Fixes GH-381

**Release Note**

```release-note
Set HSTS headers in production service
```

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -130,6 +130,12 @@ func realMain(ctx context.Context) error {
 	}
 	rateLimit := httplimiter.Handle
 
+	// Install HSTS headers in production
+	if !config.DevMode {
+		addHSTS := middleware.AddHSTS(ctx)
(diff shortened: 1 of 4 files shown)


cmd/apiserver/main.go
```diff
@@ -130,6 +130,12 @@ func realMain(ctx context.Context) error {
 	}
 	rateLimit := httplimiter.Handle
 
+	// Install HSTS headers in production
+	if !config.DevMode {
```
mikehelmick, line 134: guessing this won't matter as non-browsers will ignore...

mikehelmick: but also no harm

icco: You don't want to set it in dev because localhost w/ hsts doesn't work.

sethvargo: I think Mike was saying that no one will call the API in the browser so it probably doesn't matter

PR #390, opened 2020-08-27
Use a form (not ajax) for invalidating codes

whaught:
## Proposed Changes

* Adds a button to an invalidate endpoint that sets the code to expire now
* Alternative to ajax https://github.com/google/exposure-notifications-verification-server/pull/373 
* @sethvargo may have suggestions to make this look nicer

Screenshot from 2020-08-26 17-50-16

**Release Note**

```release-note
Allow users to expire a code
```

### DIFF
--- a/cmd/server/assets/codestatus/show.html
+++ b/cmd/server/assets/codestatus/show.html
@@ -42,6 +42,10 @@
         <h5 class="mb-1">Expiry</h5>
         <span id="code-expires-at" class="sm text-danger"></span>
       </div>
+      <form id="expireForm" action="/expire" method="POST">
+        <input type="hidden" name="uuid" value="{{.code.UUID}}" />
+        <button type="submit" id="invalidate" class="btn btn-danger btn-sm">Invalidate code now</button>
+      </form>
     </div>
 
     <div>
--- a/cmd/server/assets/codestatus/show.html
+++ b/cmd/server/assets/codestatus/show.html
@@ -42,28 +42,32 @@
         <h5 class="mb-1">Expiry</h5>
         <span id="code-expires-at" class="sm text-danger"></span>
       </div>
-      <form id="expireForm" action="/expire" method="POST">
-        <input type="hidden" name="uuid" value="{{.code.UUID}}" />
-        <button type="submit" id="invalidate" class="btn btn-danger btn-sm">Invalidate code now</button>
-      </form>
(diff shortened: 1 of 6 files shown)


cmd/server/assets/codestatus/show.html
```diff
@@ -42,24 +42,32 @@ <h5 class="mb-1">Status</h5>
         <h5 class="mb-1">Expiry</h5>
         <span id="code-expires-at" class="sm text-danger"></span>
       </div>
+      <div class="card-body">
+        <form action="/code/expire" method="POST">
```
sethvargo, line 46: I wrote some magical javascript that makes this a bit easier. Check out users/index.html for the `delete` action.

pkg/controller/codestatus/show.go
```diff
@@ -25,11 +25,11 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 )
 
-func (c *Controller) HandleShow() http.Handler {
-	type FormData struct {
-		UUID string `form:"uuid"`
-	}
+type FormData struct {
```
sethvargo, line 28: We generally want this embedded inside the handler to prevent leaking

pkg/controller/codestatus/expire.go
```diff
@@ -51,3 +52,46 @@ func (c *Controller) HandleExpire() http.Handler {
 			})
 	})
 }
+
+func (c *Controller) HandleExpirePage() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		var code *database.VerificationCode
```
sethvargo, line 59: This should be `var code database.VerificationCode` and then pass `&code` around. Otherwise you risk a NPE.

pkg/controller/codestatus/expire.go
```diff
@@ -51,3 +52,46 @@ func (c *Controller) HandleExpire() http.Handler {
...
+		if session == nil {
+			controller.MissingSession(w, r, c.h)
+			return
+		}
+		flash := controller.Flash(session)
+
+		retCode := Code{}
+
+		var form FormData
+		if err := controller.BindForm(w, r, &form); err != nil {
```
sethvargo, line 71: Instead of using a form, you can use `mux.Vars` and put the UUID in a URL. Just another option here.

PR #391, opened 2020-08-27
Set 1m idle timeout on Redis connections

sethvargo:
Centralize our Redis configuration and set a 1m idle timeout on all Redis connections

**Release Note**

```release-note
Set a 1m idle timeout on all Redis connections. `REDIS_HOST` and `REDIS_PORT` are now prefixed based on their scope (e.g. `CACHE_REDIS_HOST` and `RATE_LIMIT_REDIS_HOST`). This enables using a different Redis cluster or configuration for rate limiting vs caching.
```

/assign @mikehelmick 
/assign @icco

### DIFF
--- a/pkg/redis/config.go
+++ b/pkg/redis/config.go
@@ -1,0 +1,32 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


pkg/redis/config.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: This needs to be it's own package because otherwise we have a circular dep

PR #393, opened 2020-08-27
Keepalive

mikehelmick:
## Proposed Changes

* Add optional DB keepalive pings

**Release Note**

```release-note
New environment variable.
* `DB_MAX_CONN_IDLE_TIME` a duration, when set will be passed to DB connection pool. Default value of `1m`
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -20,8 +20,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"sync/atomic"
-	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/base64util"
 	"github.com/google/exposure-notifications-server/pkg/keys"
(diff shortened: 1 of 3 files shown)


pkg/database/database.go
```diff
@@ -130,6 +137,10 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
 		return fmt.Errorf("database gorm.Open: %w", err)
 	}
 
+	if c.MaxConnectionIdleTime > 0 {
```
sethvargo, line 140: I haven't done as much research here, but wouldn't `SetConnMaxLifetime` work better (and avoid all this overhead)?

```go
db.DB().SetConnMaxLifetime(1*time.Minute)
```

mikehelmick: made that the default - we can start there and see if it helps

mikehelmick:
just idle closes at first

PR #396, opened 2020-08-27
Add opencensus integration

### LINKED ISSUE #267: Open Census for Postgresql
### TL;DR

Implement https://opencensus.io/integrations/sql/go_sql/

taddari:
Fixes #267

## Proposed Changes

* Added wrapped driver to pass to gorm 
* Registered callbacks for gorm from https://github.com/sagikazarmark/go-gin-gorm-opencensus

**Release Note**

```release-note
After this PR SQL metrics should be available to view in metrics exporter https://opencensus.io/integrations/sql/go_sql/
```


sethvargo:
@taddari try again - added you to the triage permission which should let you control the bot

pkg/database/database.go
```diff
@@ -120,12 +126,37 @@ func (db *Database) Open(ctx context.Context) error {
...
+	return false
+}
+
 // OpenWithCacher creates a database connection with the cacher. This should
 // only be called once.
 func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) error {
 	c := db.config
 
-	rawDB, err := gorm.Open("postgres", c.ConnectionString())
+	driver := ocsql.Wrap(&postgres.Driver{}, ocsql.WithAllTraceOptions())
```
sethvargo, line 145: Two things here:

1. This feels like a lot of code, and I'm not entirely sure why it's all necessary. Can we just register the callback?

1. I'm concerned that this observability might allow a server operator to correlate specific database events (e.g. an insert) which a TEK upload. It's the same problem as the network observer. Since our chaff requests don't hit the database, it would be easy to separate chaff from non-chaff /cc @mikehelmick

taddari: Hi Seth,

About your first point, the callbacks will actually emit only gorm specific metrics. https://github.com/sagikazarmark/go-gin-gorm-opencensus/blob/master/pkg/ocgorm/stats.go. 

For postgres specific metrics, we need to override the driver. Following are the metrics that are emitted by https://github.com/opencensus-integrations/ocsql#recorded-metrics.

For point 2, I think we should disable tracing in that case and that should prevent the server operator to correlate anything.

taddari: I have changed the code to remove traceability. WDYT?

sethvargo: I think we're gonna need to hold off until we hear from the privacy team (I pinged them).

icco: Also worth remembering, we're sampling traces at 0.4%.

taddari: @sethvargo any update from the privacy team?

sethvargo: @laurentsimon ^

icco:
/lgtm
/approve

Fix merge conflict tho.

taddari:
Done the changes.
FYI, increased the time of "TestIssueToken" from 2 to 5 seconds, as 2 seconds seems not to be enough for github server although it is okay in local machine.

icco (approved):
/lgtm

PR #398, opened 2020-08-27
scope signing key read to selected realm

### LINKED ISSUE #397: API key for one realms returns signing configuration for a different realm
We noticed that using an API Key for a specific realm (i.e 3) returns signing confguration of a different realm (i.e7) , specifically the last realm that was created, causing mismatch in our workflow.

@mikehelmick

mikehelmick:
Fixes #397

**Release Note**

```release-note
CRITICAL FIX: Select correct realm's signing keys for multi-tenant signing. 
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -254,6 +254,7 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
 	var signingKey SigningKey
 	if err := db.db.
 		Model(r).
+		Where("realm_id = ?", r.ID).
 		Where("active = ?", true).
 		Order("signing_keys.created_at DESC").
 		First(&signingKey).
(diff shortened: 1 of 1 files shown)


pkg/database/realm.go
```diff
@@ -254,6 +254,7 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
 	var signingKey SigningKey
 	if err := db.db.
 		Model(r).
+		Where("realm_id = ?", r.ID).
```
sethvargo, line 257: I still don't think this is right. `r` is the realm, not the signing key table. So these are conditions against the realms table. You either need to use related or drop to the signing key table instead:

```go
		Table("signing_keys").
		Where("realm_id = ?", r.ID).
		Where("active = ?", true).
		Order("signing_keys.created_at DESC").
```

Issue #400, opened 2020-08-27
services returns error 500 due to connecting to 127.0.0.1:6379 for REDIS


sherifkozman:
Services are returning error 500 as they are trying to connect to a wrong redis host (127.0.0.1:6379) although env for REDIS (RATE_LIMIT_REDIS_HOST) is already set to a value.

@mikehelmick

PR #402, opened 2020-08-27
Do not double-prefix cache

sethvargo:
Fixes GH-400

**Release Note**

```release-note
Fix redis cache environment variable
```

/assign @mikehelmick


sethvargo:
It's not clear from the diff, but cache.Config already prefixes with `CACHE_`

PR #409, opened 2020-08-28
Finish GCLB Terraform setup

icco:
## Proposed Changes

* Waiting on next version of google-beta, which will have https://github.com/hashicorp/terraform-provider-google/issues/7051
* Attempted testing using https://github.com/hashicorp/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#running-a-local-copy-of-the-provider, but couldn't quite get it working, so putting up here for now, and will return to in bit.

**Release Note**

```release-note
Adds three new host variables to terraform and sets up a GCLB with those hosts.
```


terraform/lb.tf
```diff
@@ -11,47 +11,108 @@
...
+
+  ip_protocol           = "TCP"
+  ip_address            = google_compute_global_address.verification-server.address
+  load_balancing_scheme = "EXTERNAL"
+  port_range            = "443"
+  target                = google_compute_target_https_proxy.https.id
+  network_tier          = "PREMIUM"
+}
+
+resource "google_compute_managed_ssl_certificate" "default" {
```
sethvargo, line 110: Hmm so how does this work? Cloud Run already gives us a TLS cert.

/cc @mikehelmick

icco: This is the SSL cert on the GCLB. Cloud Run will still have it's urls available.

From https://cloud.google.com/load-balancing/docs/negs/setting-up-serverless-negs, this is the path of the things:

flow

icco: For example, you can look at server.test.encv.org and look at it's ssl cert, it's generated through this method.

sethvargo: I guess I'm worried about having two certs for the same URL.

icco: Technically that's not a problem, you can have as many certs for a domain as you want. 

With this change tho, we don't need any of the custom cloud run domain stuff tho, so that's one less thing to worry about.

icco: Just as proof, here's all the certs already issued for encv.org: https://transparencyreport.google.com/https/certificates?hl=en&cert_search_auth=&cert_search_cert=&cert_search=include_subdomains:true;domain:encv.org&lu=cert_search

sethvargo: I'm more worried if an app is doing CA pinning...

icco: In what scenario? All of the certs we've issued are through Google's Let's Encrypt thing (GTS) which all expire every 3 months. This and cloud run use the same cert issuing service. We're also introducing new domains if they are doing some sort of weird pinning there. They could pick a CA further up the chain (their choices for all of our certs are GTS or GlobalSign), but those should be even more stable. I guess I just don't see this as an issue if they are doing cert pinning correctly.

sethvargo: Okay fair. I'm probably just being TLScared.

Issue #411, opened 2020-08-28
Include git SHA/tag in builds and logs


sethvargo:
When debugging things, it'd be helpful to know which version of the code a log corresponds to. We can use build tags and the namespaced logger to always include this information.

/assign

mikehelmick:
bonus points if we display that somewhere in the UI

PR #419, opened 2020-08-30
Always build all binaries and migrations, simplify deployment

sethvargo:
In preparation for me documenting this, I wanted to do a refactor of our build/deploy/migrate/promote lifecycle. This PR:

- Uses a cache of the `/go/pkg` directory to make builds beyond the first much faster
- Updates the Terraform to create a bucket to store said cache (and permissions)
- Moves builds to the CI system and copies compiled binaries into the container
- Tags images based on tag-ref instead of user-supplied tag
- **Compiles migrations into a container image**, so migrations are now tagged and versioned with the deployment
- **Gracefully terminates the Cloud SQL Proxy after running migrations!** I believe we were slowly leaking connections because the container just kills
- Injects build ID and build tag into the logs
- Injects build ID and build tag into the HTML as a meta tag (it's available as a template global variable, so we can debate putting it somewhere else)
- Codifies a lot of the local toil in scripts where possible
- Makes it easier for someone to build off of a fork (we previously hardcoded a lot of paths)

Fixes GH-411

**Release Note**

```release-note
Improve deployment scripts
```


pkg/buildinfo/buildinfo.go
```diff
@@ -0,0 +1,25 @@
...
+// Package buildinfo provides high-level build information injected during
+// build.
+package buildinfo
+
+var (
+	// BuildID is the unique build identifier.
+	BuildID string = "unknown"
+
+	// BuildTag is the git tag from which this build was created.
+	BuildTag string = "unknown"
```
mikehelmick, line 24: is it possible to also get the commit sha for when a build is done from head

jeremyfaller: Does `go tool buildid` work for your use case???

pkg/controller/middleware/template.go
```diff
@@ -33,14 +34,10 @@ func PopulateTemplateVariables(ctx context.Context, config *config.ServerConfig)
...
-
-			// Set initial variables if they are not set.
-			if _, ok := m["server"]; !ok {
-				m["server"] = config.ServerName
-			}
-			if _, ok := m["title"]; !ok {
-				m["title"] = config.ServerName
-			}
+			m["server"] = config.ServerName
+			m["title"] = config.ServerName
```
icco, line 38: Don't you not want to overwrite this?

builders/deploy.yaml
```diff
@@ -18,22 +18,90 @@
...
   _TAG:
 
 steps:
-- id: 'deploy'
-  name: 'gcr.io/google.com/cloudsdktool/cloud-sdk:303.0.0-alpine'
+#
+# adminapi
+#
+- id: 'deploy-adminapi'
+  name: 'gcr.io/google.com/cloudsdktool/cloud-sdk:307.0.0-alpine'
```
icco, line 29: Remove the version?

pkg/buildinfo/buildinfo.go
sethvargo: `git describe --dirty` (which I'm using) does what I think you want, some examples:

```text
v0.1.1
v0.1.1-ab28ced
v0.1.1-ab28ced-dirty
```

sethvargo:
> I personally prefer to be more explicit (by just having a Dockerfile per service), at the cost of speed, so that we could just setup cloud build to automatically build every commit, and then just deploy and promote when we want. But given this is optimizing for the local workflow, it seem solid.

@icco I don't think this precludes or eliminates that possibility. In fact, I want to set this up to auto-build on each commit to HEAD, then we just deploy, maybe migrate, and promote.

cloudbuild.yaml
```diff
@@ -1,115 +0,0 @@
-timeout: 40m
```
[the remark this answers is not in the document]
sethvargo: We can point Cloud Build at the build.yaml file (which was my plan here).

pkg/controller/middleware/template.go
sethvargo: It's actually not possible to overwrite them since we've moved this into the middleware. This is an artifact from whenever they were lazy-added, but now we force-add these attributes very early in the middelware.

builders/deploy.yaml
sethvargo: Eh, two things:

1. I've been bitten by subtle changes in gcloud and would prefer to make deploys as reproducible as possible by reducing variables.

2. Cloud Build is going to cache the container image. So "latest" isn't actually latest anyway.

pkg/buildinfo/buildinfo.go
sethvargo: @jeremyfaller we explicitly want the _Cloud Build_ Build ID here (it's a UUID).

builders/deploy.yaml
icco: Fair enough. I always prefer latest if possible because folks always forget to update, and the larger the diff on upgrade, the higher the chance of bug.

icco:
> > I personally prefer to be more explicit (by just having a Dockerfile per service), at the cost of speed, so that we could just setup cloud build to automatically build every commit, and then just deploy and promote when we want. But given this is optimizing for the local workflow, it seem solid.
> 
> @icco I don't think this precludes or eliminates that possibility. In fact, I want to set this up to auto-build on each commit to HEAD, then we just deploy, maybe migrate, and promote.

I talked with Seth and reread and I don't think I truly understood what this is doing. The details that was missing for me is that `/workspace` is shared between all cloud builds, and that we are going to point cloud build at this `build.yaml`.

jeremyfaller:
No, I'm all good. If you want a LG, LMK.

PR #423, opened 2020-08-31
Include DB_CONN in .env

sethvargo:
**Release Note**

```release-note
NONE
```


terraform/main.tf
```diff
@@ -148,11 +148,13 @@ export COOKIE_KEYS="secret://${google_secret_manager_secret_version.cookie-hmac-
 # Note: these configurations assume you're using the Cloud SQL proxy!
 export DB_APIKEY_DATABASE_KEY="secret://${google_secret_manager_secret_version.db-apikey-db-hmac.id}"
 export DB_APIKEY_SIGNATURE_KEY="secret://${google_secret_manager_secret_version.db-apikey-sig-hmac.id}"
+export DB_CONN="${google_sql_database_instance.db-inst.connection_name}"
+export DB_DEBUG="true"
```
mikehelmick, line 152: not sure we want this true by default

icco: I mean, this is just for the users local config, so I think it's good?

sethvargo: Yea, this would be local config (primarily used when running migrations). This is _not_ set on the service.

PR #427, opened 2020-08-31
force_override isn't persisted in the spec

sethvargo:
**Release Note**

```release-note
NONE
```


sethvargo:
Well that's a new error. Docker hub down?

PR #439, opened 2020-09-01
Redis name configurable in terraform

chaodaiG:
/assign @sethvargo

### DIFF
--- a/terraform-e2e/README.md
+++ b/terraform-e2e/README.md
@@ -4,6 +4,8 @@ This directory contains terraform configuration to be used for deploying verific
 
 ## Prerequisite
 
+Terraform version > 1.13.1
+
 Follow steps from [Terrafrom instructions](https://github.com/google/exposure-notifications-verification-server/tree/main/terraform), going through from top until finishing `Instructions` section, then change into `terraform-e2e` directory:
 
 ```shell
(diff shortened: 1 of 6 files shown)


terraform-e2e/README.md
```diff
@@ -26,23 +28,6 @@ terraform import module.en.google_app_engine_application.app ${PROJECT_ID}
...
-
-  on ../terraform/main.tf line 15:
-  15: provider "google" {
-
-The configuration for
-module.en.provider["registry.terraform.io/hashicorp/google"] depends on values
-that cannot be determined until apply.
-```
-
-This can be mitigated by commenting out [project = var.project](https://github.com/google/exposure-notifications-verification-server/blob/86a05f09dabfeddbce73c9d628ec01d3738782ed/terraform/main.tf#L16), as well as [project = var.project](https://github.com/google/exposure-notifications-verification-server/blob/86a05f09dabfeddbce73c9d628ec01d3738782ed/terraform/main.tf#L23). This is similar to the issue described in https://github.com/hashicorp/terraform/issues/25816, and might be fixed in 0.13.1 release of terraform.
```
chaodaiG, line 44: This had been confirmed fixed in terraform 1.13.1

terraform-e2e/README.md
```diff
@@ -4,6 +4,8 @@ This directory contains terraform configuration to be used for deploying verific
 
 ## Prerequisite
 
+Terraform version > 1.13.1
```
sethvargo, line 7: We don't need to document this. If 1.13 is a requirement, we should add this to the terraform configs under `required_version` (we already have 1.13 there, just bump it). Better to capture this as code that docs.

terraform-e2e/README.md
sethvargo: I'd rather leave this in with a note at the top saying "upgrade to the latest Terraform version first"

terraform-e2e/main.tf
```diff
@@ -28,6 +28,7 @@ module "en" {
 
   project = var.project
   database_name = "en-verification-${random_string.suffix.result}"
+  redis_name = "verification-cache-${random_string.suffix.result}"
```
sethvargo, line 31: Please run `terraform fmt` or install an editor plugin that does auto-formatting.

terraform/variables.tf
```diff
@@ -66,6 +66,11 @@ variable "storage_location" {
   default = "US"
 }
 
+variable "redis_name" {
+  type = string
```
sethvargo, line 70: Please run `terraform fmt`.

terraform-e2e/README.md
chaodaiG: Done

terraform-e2e/README.md
chaodaiG: Added in code

terraform-e2e/main.tf
chaodaiG: done

terraform/variables.tf
chaodaiG: done

PR #446, opened 2020-09-01
Add per realm stats

### LINKED ISSUE #410: Realm level stats
### TL;DR

Add charts AND graphs of how many codes a realm has issued per day.

Bonus points for graphing against number of codes that were claimed.

/assign @jeremyfaller 
/kind enhancement

jeremyfaller:
Adds per realm stats, and a simple visualization of the codes issued and
claimed for the past month.

Fixes #410

**Release Note**

```release-note
Adds per-realm stats, and a simple visualization of the last 30 days of issued and claimed codes.
```

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -184,6 +184,7 @@
               <a class="dropdown-item" href="/apikeys">API keys</a>
               <a class="dropdown-item" href="/realm/keys">Signing keys</a>
               <a class="dropdown-item" href="/users">Users</a>
+              <a class="dropdown-item" href="/realm/stats">Statistics</a>
               <a class="dropdown-item" href="/realm/settings">Settings</a>
               <div class="dropdown-divider"></div>
               {{end}}
--- a/pkg/controller/realmstats/index.go
+++ b/pkg/controller/realmstats/index.go
@@ -1,0 +1,55 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package realmstats
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+)
+
+func (c *Controller) HandleIndex() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		realm := controller.RealmFromContext(ctx)
+		if realm == nil {
+			controller.MissingRealm(w, r, c.h)
+			return
+		}
+
+		today := time.Now().Truncate(24 * time.Hour)
+		nago := today.Add(-30 * 24 * time.Hour).Truncate(24 * time.Hour)
+		stats, err := realm.Stats(c.db, nago, today)
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		c.renderShow(ctx, w, realm, stats)
+	})
+}
+
+func (c *Controller) renderShow(ctx context.Context, w http.ResponseWriter, realm *database.Realm, stats []*database.RealmStats) {
+	m := controller.TemplateMapFromContext(ctx)
+	m["realm"] = realm
+	m["stats"] = stats
+
+	// Valid settings for code parameters.
+	c.h.RenderHTML(w, "realmstats", m)
+}
--- a/pkg/controller/realmstats/realmstats.go
+++ b/pkg/controller/realmstats/realmstats.go
@@ -1,0 +1,43 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package realmstats contains web controllers for changing realm stats.
+package realmstats
+
+import (
+	"context"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+
+	"github.com/google/exposure-notifications-server/pkg/logging"
+
+	"go.uber.org/zap"
+)
+
+type Controller struct {
+	db     *database.Database
+	h      *render.Renderer
+	logger *zap.SugaredLogger
+}
+
+func New(ctx context.Context, db *database.Database, h *render.Renderer) *Controller {
+	logger := logging.FromContext(ctx).Named("realmstats")
+
+	return &Controller{
+		db:     db,
+		h:      h,
+		logger: logger,
+	}
+}
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -608,3 +608,27 @@ func (r *Realm) DestroySigningKeyVersion(ctx context.Context, db *Database, id i
 
 	return nil
 }
+
+// Stats returns the usage statistics for this realm. If no stats exist,
+// returns an empty array.
+func (r *Realm) Stats(db *Database, start, stop time.Time) ([]*RealmStats, error) {
+	var stats []*RealmStats
+
+	start = start.Truncate(24 * time.Hour)
+	stop = stop.Truncate(24 * time.Hour)
+
+	if err := db.db.
+		Model(&RealmStats{}).
+		Where("realm_id = ?", r.ID).
+		Where("(date >= ? AND date <= ?)", start, stop).
+		Order("date ASC").
+		Find(&stats).
+		Error; err != nil {
+		if IsNotFound(err) {
+			return stats, nil
+		}
+		return nil, err
+	}
+
+	return stats, nil
+}
--- a/pkg/database/realm_stats.go
+++ b/pkg/database/realm_stats.go
@@ -1,0 +1,32 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"time"
+)
+
+// RealmStats represents statistics related to a user in the database.
+type RealmStats struct {
+	Date         time.Time `gorm:"date; not null"`
+	RealmID      uint      `gorm:"realm_id; not null"`
+	CodesIssued  uint      `gorm:"codes_issued; default: 0"`
+	CodesClaimed uint      `gorm:"codes_claimed; default: 0"`
+}
+
+// TableName sets the RealmStats table name
+func (RealmStats) TableName() string {
+	return "realm_stats"
+}
--- a/pkg/database/token.go
+++ b/pkg/database/token.go
@@ -183,23 +183,16 @@ func (db *Database) VerifyCodeAndIssueToken(realmID uint, verCode string, accept
 			return err
 		}
 
-		if expired, err := db.IsCodeExpired(&vc, verCode); err != nil {
-			db.logger.Debugw("error checking code expiration", "error", err)
-			return ErrVerificationCodeExpired
-		} else if expired {
-			return ErrVerificationCodeExpired
-		}
-
-		if vc.Claimed {
-			return ErrVerificationCodeUsed
-		}
-
 		if _, ok := acceptTypes[vc.TestType]; !ok {
 			return ErrUnsupportedTestType
 		}
 
+		// And mark as claimed.
+		if err := vc.claim(db, tx, verCode, realmID); err != nil {
+			return err
+		}
+
 		// Mark claimed. Transactional update.
-		vc.Claimed = true
 		if err := tx.Save(&vc).Error; err != nil {
 			return err
 		}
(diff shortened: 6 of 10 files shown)


cmd/server/assets/header.html
```diff
@@ -185,6 +185,7 @@ <h6 class="dropdown-header">Manage realm</h6>
               <a class="dropdown-item" href="/realm/keys">Signing keys</a>
               <a class="dropdown-item" href="/users">Users</a>
               <a class="dropdown-item" href="/realm/settings">Settings</a>
+              <a class="dropdown-item" href="/realm/stats">Statistics</a>
```
sethvargo, line 188: We might want to move this above Settings

pkg/controller/realmstats/index.go
```diff
@@ -0,0 +1,57 @@
...
+
+		c.renderShow(ctx, w, realm, stats)
+	})
+}
+
+func (c *Controller) renderShow(ctx context.Context, w http.ResponseWriter, realm *database.Realm, stats []*database.RealmStats) {
+	m := controller.TemplateMapFromContext(ctx)
+	m["realm"] = realm
+	m["stats"] = stats
+	fmt.Printf("%+v\n", stats)
```
sethvargo, line 53: Debug

pkg/controller/realmstats/realmstats.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type Controller struct {
+	db     *database.Database
+	h      *render.Renderer
+	logger *zap.SugaredLogger
+}
+
+func New(ctx context.Context, db *database.Database, h *render.Renderer) *Controller {
+	logger := logging.FromContext(ctx)
```
sethvargo, line 36:
```suggestion
	logger := logging.FromContext(ctx).Named("realmstats")
```

pkg/database/realm.go
```diff
@@ -608,3 +608,27 @@ func (r *Realm) DestroySigningKeyVersion(ctx context.Context, db *Database, id i
 
 	return nil
 }
+
+// GetStats returns the usage statistics for this realm. If no stats exist,
+// returns an empty array.
+func (r *Realm) GetStats(db *Database, start, stop time.Time) ([]*RealmStats, error) {
```
sethvargo, line 614: Per Go guidelines, I think this should just be `Stats()`?

pkg/database/realm_stats.go
```diff
@@ -0,0 +1,32 @@
...
+
+package database
+
+import (
+	"time"
+)
+
+// RealmStats represents statistics related to a user in the database.
+type RealmStats struct {
+	Date         time.Time `gorm:"date"`
```
sethvargo, line 23: Date and RealmID probably need NOT NULL constraints

pkg/database/token.go
```diff
@@ -204,6 +204,19 @@ func (db *Database) VerifyCodeAndIssueToken(realmID uint, verCode string, accept
 			return err
 		}
 
+		// Log the issued code.
+		now := time.Now().Truncate(24 * time.Hour)
```
sethvargo, line 208: I think we might want to do this inside a callback instead of here.

pkg/database/vercode.go
```diff
@@ -97,11 +98,22 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
 		`
 
-		day := time.Now().UTC().Truncate(24 * time.Hour)
 		if err := scope.DB().Exec(sql, day, v.IssuingAppID).Error; err != nil {
 			scope.Log(fmt.Sprintf("failed to update stats: %v", err))
 		}
 	}
+
+	// Update the per-realm stats.
+	sql := `
```
sethvargo, line 107: Yea, if you move this into a callback (see the other stats calculations), we can avoid the duplication.

pkg/database/realm.go
jeremyfaller: Done

pkg/controller/realmstats/index.go
mikehelmick: the printf is still there

pkg/database/token.go
mikehelmick: agreed - I think it's best to not have stats update required in the transaction

jeremyfaller: Done

pkg/database/vercode.go
jeremyfaller: Done.

pkg/controller/realmstats/index.go
jeremyfaller: Sorry, I hadn't done a push. My workflow has been to tweak the comment after I update my code. Apparently, messages are sent when I do that. I'll tweak my workflow to not update comments until the new code's been uploaded. Sorry for the spam.

pkg/database/realm.go
```diff
@@ -608,3 +608,27 @@ func (r *Realm) DestroySigningKeyVersion(ctx context.Context, db *Database, id i
...
+func (r *Realm) Stats(db *Database, start, stop time.Time) ([]*RealmStats, error) {
+	var stats []*RealmStats
+
+	start = start.Truncate(24 * time.Hour)
+	stop = stop.Truncate(24 * time.Hour)
+
+	if err := db.db.
+		Model(&RealmStats{}).
+		Where("realm_id = ?", r.ID).
+		Where("date >= ? AND date <= ?", start, stop).
```
sethvargo, line 623:
```suggestion
		Where("(date >= ? AND date <= ?)", start, stop).
```

I don't trust gorm sometimes...

pkg/database/vercode.go
sethvargo: Did you push? I'm not seeing any changes.

jeremyfaller:
/hold

Hold on this right now. I don't like the error handling I've put in vercode.go. I'm going to add some more testing.

pkg/database/vercode.go
```diff
@@ -97,11 +98,54 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
+			SET codes_issued = realm_stats.codes_issued + 1
+	`
+
+	if err := scope.DB().Exec(sql, day, v.RealmID).Error; err != nil {
+		scope.Log(fmt.Sprintf("failed to update stats: %v", err))
+	}
+}
+
+// BeforeSave is called after the verification code is saved to the DB.
+func (v *VerificationCode) BeforeSave(scope *gorm.Scope) error {
```
sethvargo, line 120: Hmm - doing this in BeforeSave puts us in a spot where we'd increment the stats, but then some validation could prevent the code from actually being claimed.

jeremyfaller: Ok, I'm going to a side value in VerificationCode. This is too clunky.

sethvargo: Okay - what are your thoughts on making a `Claim` function on the verification code (or on the database) that handles all this logic?

jeremyfaller: LMK what you think of this. It's clunky, but with Hooks (and ORM with dirty) this feels the best we can do. I'm happy to move to a Claim API if you think that's best.

pkg/database/vercode.go
```diff
@@ -73,6 +77,8 @@ func (VerificationCode) TableName() string {
 // to update statistics about usage. If the executions fail, an error is logged
 // but the transaction continues. This is called automatically by gorm.
 func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
+	day := time.Now().UTC().Truncate(24 * time.Hour)
```
mikehelmick, line 80: why not use v.CreatedAt?

pkg/database/vercode.go
```diff
@@ -62,6 +62,10 @@ type VerificationCode struct {
 	LongExpiresAt time.Time
 	IssuingUserID uint
 	IssuingAppID  uint
+
+	// When we claim a VerificationCode through the API, we need to trigger
+	// the per-realm stats update.
+	AreClaiming bool
```
mikehelmick, line 68: not sure I understand why this is needed in addition to claimed?

jeremyfaller: Seth had wanted to use Hooks for updating the stats counts, removing the straight DB update someplace else.

If we do that our ORM doesn't know about dirty bits. So, if the VerificationCode has Claimed set, we'll update the stats whenever the hook is called. Now, we could query the DB, and get the old value -- but it's a question of timing. If we do it BeforeSave, we don't know that we're actually going to do the write. If we do it AfterSave, we don't know if Claimed was set BEFORE the update happened. It's a mess.  (This discussion ignores the fact that testing exercises the API in a manner the app never will.)

So, there's two ways out of it -- having this stupid variable, and continuing to use Hooks; or have a Claim function that does this. Hooks are nice because they're auto-updated and you never have to remember to call Claim. The function is nice because it doesn't have this crap floating around.

I am shooting the middle on all approaches here, and as you and Seth have the larger picture, I defer to your judgements (right now!).

pkg/database/vercode.go
jeremyfaller: Oh, better. on it.

pkg/database/vercode.go
mikehelmick: to properly use this as a dirty bit, don't we need to set AreClaiming = false when updating the stats?

otherwise this problem persists

or only set it to true when the stats are updated (actually I like that better, because it enables backfill on stats later if we need to do that)

jeremyfaller: AreClaiming will default to false (go runtime) for all values read from the DB.

Backfill is a nice side effect I hadn't considered.

mikehelmick: except, we're saving it to the database

jeremyfaller: I thought only values listed as "gorm" are saved!!!!

jeremyfaller: argh

sethvargo: Hmmmmmm. After seeing that we need to track an additional field, I don't think this is the best approach. I think adding a `verificationcode.Claim()` or `database.ClaimVerificationCode` would be cleaner. Sorry for the turmoil.

jeremyfaller:
/hold

Holding AGAIN. while I work out what I don't understand about gorm.

pkg/database/vercode.go
mikehelmick: To not have it be part of the database model, you have to explicitly tag as ignore

https://gorm.io/docs/models.html#Fields-Tags

pkg/database/vercode.go
jeremyfaller: Moved to claim. LMK.

pkg/database/token.go
```diff
@@ -183,23 +183,16 @@ func (db *Database) VerifyCodeAndIssueToken(realmID uint, verCode string, accept
...
-		if vc.Claimed {
-			return ErrVerificationCodeUsed
-		}
-
 		if _, ok := acceptTypes[vc.TestType]; !ok {
 			return ErrUnsupportedTestType
 		}
 
+		// And mark as claimed.
+		if err := vc.claim(db, verCode, realmID); err != nil {
```
mikehelmick, line 191: this needs to be in tx not db!!!

mikehelmick (changes requested):

pkg/database/vercode.go
```diff
@@ -97,11 +96,51 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
+
+	// Update the per-realm stats.
+	sql := `
+		INSERT INTO realm_stats(date, realm_id, codes_issued)
+			VALUES ($1, $2, 1)
+		ON CONFLICT (date, realm_id) DO UPDATE
+			SET codes_issued = realm_stats.codes_issued + 1
+	`
+
+	if err := scope.DB().Exec(sql, v.CreatedAt.Truncate(24*time.Hour), v.RealmID).Error; err != nil {
```
sethvargo, line 112: We don't currently truncate the other times - should we be?

pkg/database/vercode.go
```diff
@@ -97,11 +96,51 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
+		scope.Log(fmt.Sprintf("failed to update stats: %v", err))
+	}
+}
+
+// claim checks the VerificationCode, marks as claimed, and updates the
+// per-realm stats. If any of that fails, an error code is returned.
+func (v *VerificationCode) claim(db *Database, tx *gorm.DB, verCode string, realmID uint) error {
+	if expired, err := db.IsCodeExpired(v, verCode); err != nil {
+		db.logger.Debugw("error checking code expiration", "error", err)
```
sethvargo, line 121: I think we should drop this log message (and the dependency on `db`) and return the error instead. That'll make it bubble as an internalerror instead of silently swallowing right now.

pkg/database/vercode.go
jeremyfaller: yeah, prolly. With the suggested change here (to use the date from the code), I think all things want truncated dates.

pkg/database/vercode.go
jeremyfaller: you got it.

pkg/database/vercode.go
```diff
@@ -97,11 +97,50 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
 
-		day := time.Now().UTC().Truncate(24 * time.Hour)
-		if err := scope.DB().Exec(sql, day, v.IssuingAppID).Error; err != nil {
+		if err := scope.DB().Exec(sql, date, v.IssuingAppID).Error; err != nil {
 			scope.Log(fmt.Sprintf("failed to update stats: %v", err))
 		}
 	}
+
+	// Update the per-realm stats.
+	sql := `
```
sethvargo, line 106: We should probably guard this with a sanity:

```go
if v.RealmID != 0 {
```

can be a follow-up

pkg/database/vercode.go
```diff
@@ -97,11 +97,50 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
+	if err := scope.DB().Exec(sql, date, v.RealmID).Error; err != nil {
+		scope.Log(fmt.Sprintf("failed to update stats: %v", err))
+	}
+}
+
+// claim checks the VerificationCode, marks as claimed, and updates the
+// per-realm stats. If any of that fails, an error code is returned.
+func (v *VerificationCode) claim(db *Database, tx *gorm.DB, verCode string, realmID uint) error {
+	if expired, err := db.IsCodeExpired(v, verCode); err != nil {
+		return err
```
sethvargo, line 122: It'd be nice to wrap this error like `fmt.Errorf("failed to claim code: %w")`

sethvargo: Can be a follow-up

pkg/database/vercode.go
```diff
@@ -97,11 +97,50 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
...
+	`
+
+	if err := scope.DB().Exec(sql, date, v.RealmID).Error; err != nil {
+		scope.Log(fmt.Sprintf("failed to update stats: %v", err))
+	}
+}
+
+// claim checks the VerificationCode, marks as claimed, and updates the
+// per-realm stats. If any of that fails, an error code is returned.
+func (v *VerificationCode) claim(db *Database, tx *gorm.DB, verCode string, realmID uint) error {
```
sethvargo, line 120: This API is a little weird to me now, and I'm not sure the comment matches. While this sets `Claimed: true` on the Verification Code, it does not _save_ that code in the database. The caller still needs to call `db.Save(v)`. That feels subtle and could lead to some bugs. Furthermore, we're updating the realm stats _before_ successfully marking the code as claimed. I'm thinking this flow should be more like:

1. Expiration checks
1. Claimed check
1. Mark as claimed on struct
1. Save struct to database (stop if error)
1. Update stats
1. Return

Alternatively, we need to document this better. What do you think?

Issue #429, opened 2020-09-01
Implement rotation for various encryption/signing keys


sethvargo:
/assign

PR #450, opened 2020-09-02
Allow rotating most secrets

sethvargo:
This isn't 100% complete, but it's about 95% toward GH-429. I'd eventually like to write a utility in `tools/` that generates the secrets and puts them in Secret Manager so a human never sees them, but didn't want to muddle that with the actual docs/implementation.

**Release Note**

```release-note
Add support for rotating HMAC and encryption keys
```

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -2,6 +2,7 @@
 
 This page includes helpful tips for configuring things in production:
 
+
 ## Observability (tracing and metrics)
 
 The observability component is responsible for metrics. The following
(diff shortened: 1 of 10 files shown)


docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
...
+
+You can use `openssl` or similar tooling to generate the secret. Note that this
+is **not** preferred since it requires a human to see the plaintext secret.
+
+```sh
+openssl rand -base64 64 | tr -d "\n" # or 32
+```
+
+If you are using a secret manager, put this value in the secret manager and
+provie its _reference_ in the environment. If you are not using a secret
```
icco, line 67: s/provie/provide/

docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
...
+
+**Recommended frequency:** 30 days, on breach
+
+The cookie keys are an array. The items at odd indicies are HMAC keys and the
+items at even indicies are encryption keys. The HMAC key should be 64 bytes and
+the encryption key should be 32. Both keys are supplied to this system in
+base64, for example:
+
+```sh
+export COOKIE_KEYS="ARLaFwAqBGIkm5pLjAveJuahtCnX2NLoAUz2kCZKrScUaUkEaxHSvJLVYb5yAPCc441Cho5n5yp8jdEmy6hyig==,RLjcRZeqc07s6dh3OK4CM1POjHDZHC+usNU1w/XNTjM="
```
icco, line 31: Could. you add a command to turn this from base64? Just to make this as dumb simple as possible to follow in case of a breach.

docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
...
+upgrade them so long as the older key version is still available in your key
+manager.
+
+While unlikely, this may require you to update the `DB_ENCRYPTION_KEY`
+environment variable.
+
+
+### API Key Signature HMAC keys
+
+**Recommended frequency:** 90 days
```
icco, line 108: not on breach?

docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
...
+```sh
+DB_APIKEY_SIGNATURE_KEY="gSEGlr482MSTm0eGRm2VvS86iQin3+/+80ALBkKKBYgu2EJyhGkvi8BULeFQDW/qZp2y3IbKY0MUTqioG7InBZdCkisYjr8UNuA+PONxMSaw/x8m+CXF28qb2WF0OGYQIPgbOdQ7cChG3Ox5AQgWFmNwyr486lTxSM8TE7dfCfU=,oXrnYzt6MXKBB/+zZWTvkUABW8SSCAFv5Mc475sSVPGBqCf1hWvv/VmByFj/5457Ho0AVbDUiCpKnjW2Q8ZlxqRo5dJyRifwvmW2JYcpxe+Ff/d+tb2x+TwlzqEzrKVatEWX4cmMG7ZP6B1oTCw/uZPTDhgB3lerXVIBTxdAaQc="
+```
+
+Note: Removing any of the keys from this list will invalidate API keys signed by
+that version.
+
+
+### API Key Database HMAC keys
+
```
icco, line 133: not on breach?

docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
...
+DB_APIKEY_SIGNATURE_KEY="1do5HM96Bk9WD15BQC3qbW9e3T2V6T0DHn2i1xGJRKX8tZubxuaezivMhO3uJFEye8SITH3UFB+mo9oE0VGPiP/4TOXejfsd1g2L518itJbrK4/qNh6QMk0I04mqNtR55uvyt7G/ObADn2hQDYMVOGg/C6nLiqO+nqQ/UoUcGkA=,tJiUPEi0xS5QbykypVlquWsxQ0DAgxY41w+PkNqpoqzWQyDnEUAWFwIFUUFllqT+m0f2Kqh8EB1zjYgFcGP16O52rHer5sr4x6nsnQ/AiOHDrztJnEqGvutetHhZHLGKY0HxlxkZxcFLCmbgs6pa0vNUodrzOsCYysD7MLCSL5M="
+```
+
+Note: Removing any of the keys from this list will invalidate API keys HMACed by
+that version.
+
+
+### Verification Code Database HMAC Key
+
+**Recommended frequency:** 30 days
```
icco, line 160: not on breach?

docs/production.md
```diff
@@ -11,3 +12,180 @@ configurations are available:
 | ----------------------- | ------------------------------- | -----------
 | OpenCensus Agent        | `OCAGENT`                       | Use OpenCensus.
 | Stackdriver\*           | `STACKDRIVER`                   | Use Stackdriver.
+
+
+## Rotating secrets
```
icco, line 17: I like this a lot, I look forward to the tooling. Maybe add a TODO for that?

pkg/database/authorized_app.go
```diff
@@ -229,16 +227,35 @@ func (db *Database) SaveAuthorizedApp(r *AuthorizedApp) error {
...
+	sig := hmac.New(sha512.New, keys[0])
+	if _, err := sig.Write([]byte(apiKey)); err != nil {
+		return "", err
 	}
 	return base64.RawURLEncoding.EncodeToString(sig.Sum(nil)), nil
 }
 
+// generateAPIKeyHMACs creates a permutation of all possible API keys based on
+// the provided HMACs. It's primarily used to find an API key in the database.
+func (db *Database) generateAPIKeyHMACs(apiKey string) ([]string, error) {
```
icco, line 247: why is one public but not the other?

pkg/database/authorized_app.go
```diff
@@ -229,16 +227,35 @@ func (db *Database) SaveAuthorizedApp(r *AuthorizedApp) error {
...
+	sigs := make([]string, 0, len(db.config.APIKeyDatabaseHMAC))
```
icco, line 248: why not `var sigs []string`

icco (commented):
Mostly just nits and questions

docs/production.md
sethvargo: Not sure I understand. The `openssl` commands output base64, and the envvar accepts base64. Why would we want something that decodes it?

docs/production.md
sethvargo: It doesn't really help on breach. In the case of API keys + breach, there will need to be a business decision about how to move forward. Either break existing clients and do a full key rotation, or tune down the rate limiter to prevent abuse until you can push out new API keys to clients.

docs/production.md
sethvargo: See above

docs/production.md
sethvargo: I was toying with the tooling yesterday. It's actually somewhat annoying because Terraform is managing the secret and the envvars.

pkg/database/authorized_app.go
sethvargo: Because we never want someone outside the package to be able to generate the _full_ HMAC set (which is only used for comparisons). Outside of the package, consumers should only generate the _latest_ HMAC.

docs/production.md
sethvargo: See above

docs/production.md
sethvargo: Clarified.

icco (approved):
/lgtm 

Thanks for clarifying!

PR #451, opened 2020-09-02
Restore the old API key code

sethvargo:
Semi-revert of GH-239. This adds two "error" fields to our responses, but it's the best interim solution I can think of.

/assign @mikehelmick 

**Release Note**

```release-note
Temporarily return `error_code` and `errorCode` in API JSON responses for backwards compatibility. `error_code` will be removed in a future version.
```


sethvargo:
/hold

See chat

mikehelmick:
this is fine - it will be essentially impossible to deprecate.
If we ever create a v2 of the API, we can really remove it.

/lgtm
/unhold

PR #454, opened 2020-09-02
Fix intial container images path in terraform services

sherifkozman:
Fixes #
Container images are not in sync with changes in builder scripts

Container images in terraform scripts are pointing towards an older path which was recently changed in cloud builders.

Old Path : gcr.io/${var.project}/github.com/google/exposure-notifications-verification-server/cmd/SERVICE-NAME:initial

New Path : Old Path : gcr.io/${var.project}/github.com/google/exposure-notifications-verification-server/SERVICE-NAME:initial


sherifkozman:
@googlebot I fixed it

PR #455, opened 2020-09-02
Ensure Debugw always has the right number of fields

sethvargo:
Apparently this turned into a panic instead of a silent failure in a recent version.

**Release Note**

```release-note
Fix migration logging
```

/cc @chaodaiG

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -52,12 +52,12 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 			Migrate: func(tx *gorm.DB) error {
 				// This is "out of order" as it were, but is needed to bootstrap fresh systems.
 				// Also in migration 8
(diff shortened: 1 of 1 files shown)


pkg/database/migrations.go
```diff
@@ -52,12 +52,12 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 			Migrate: func(tx *gorm.DB) error {
 				// This is "out of order" as it were, but is needed to bootstrap fresh systems.
 				// Also in migration 8
-				logger.Debugw("db migrations: creating realms table")
+				logger.Debugw("creating realms table")
```
chaodaiG, line 55: How about the logger in this function, does it need the same change?

sethvargo: no, debugw is `(args, ...vars)`. It requires vars be an even number (and 0 is even)

PR #465, opened 2020-09-02
Move API requests to a sub router

icco:
## Proposed Changes

* Moves /api requests into a subrouter
* Makes /health not require an API key

**Release Note**

```release-note
/health no longer requires an API key to call
```


sethvargo (approved):
/lgtm

PR #468, opened 2020-09-02
Use crypto/rand to generate random suffixes in API keys.

yegle:
Log lines like this indicating our probing frequency is fast enough to
cause duplications:

```
error trying to create a new Admin API Key: pq: duplicate key value
violates unique constraint "realm_apikey_name"
```

```release-note
NONE
```

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -17,12 +17,12 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
-	"math/rand"
+	"math/big"
 	"net/http"
 	"os"
 	"strconv"
-	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-server/pkg/observability"
@@ -63,9 +63,12 @@ func main() {
 	logger.Info("successful shutdown")
 }
 
-func randomString() string {
-	rand.Seed(time.Now().Unix())
-	return fmt.Sprintf("%x", rand.Int63())
+func randomString() (string, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(10000))
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%04x", n), nil
 }
 
 func realMain(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


cmd/e2e-runner/main.go
```diff
@@ -64,7 +64,7 @@ func main() {
 }
 
 func randomString() string {
-	rand.Seed(time.Now().Unix())
+	rand.Seed(time.Now().UnixNano())
```
mikehelmick, line 67: please us crypto/rand instead of rand

yegle:
We don't need CSRNG here as it's just used as a suffix to the generated API key. `rand` should be sufficient for the purpose.

Alternatively we can remove the `randomString()` and just use `time.Now().UnixNano()` as suffix.

icco:
> We don't need CSRNG here as it's just used as a suffix to the generated API key. `rand` should be sufficient for the purpose.
> 
> Alternatively we can remove the `randomString()` and just use `time.Now().UnixNano()` as suffix.

Switch to just appending since this is run by cron.

cmd/e2e-runner/main.go
mikehelmick: it's for consistency w/ the rest of the codebase, we're always using crypto/rand instead of rand package

sethvargo: Yup, change math/rand to crypto/rand in the imports and remove this line altogether.

yegle:
I switched to `crypto/rand` and updated the PR description. PTAL.

cmd/e2e-runner/main.go
```diff
@@ -63,9 +63,12 @@ func main() {
 	logger.Info("successful shutdown")
 }
 
-func randomString() string {
-	rand.Seed(time.Now().Unix())
-	return fmt.Sprintf("%x", rand.Int63())
+func randomString() (string, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(10000))
```
sethvargo, line 67: It's probably better to read some number of bytes (e.g. 16), then base64 or hex-encode the result. We do this in a few test helpers.

icco (approved):
/lgtm

PR #475, opened 2020-09-04
Add padding fields to API

sethvargo:
This adds a padding field to all requests and responses. We don't use it anywhere yet, but it will allow clients to send us padded responses for network obfuscation, and we can gradually start sending our own padding back in responses that the client should ignore.

**Release Note**

```release-note
Add `padding` fields to all API requests and responses
```

/assign @mikehelmick


icco (approved):
/lgtm

PR #476, opened 2020-09-04
Terraform fix for Alerting services

sherifkozman:
Terraform doesn't scan recursive directories so Alerting service will not be deployed.

## Proposed Changes

 - Moved notifications and probes terraform scripts to root directory 
 - Added missing "notification_email" variable to main variables.tf 

**Release Note**

```release-note
action required: If you had previously deployed the alerting rules to a separate module, they will need to be moved to the main module.
```


icco:
So we want this in a separate folder because not everyone wants alerting.

/hold
/assign

sherifkozman:
@icco there is a condition of the alerting variables are not present the resources won't be provisioned so if alerting isn't needed it won't be added

icco:
> @icco there is a condition of the alerting variables are not present the resources won't be provisioned so if alerting isn't needed it won't be added

Where is that condition? I'm failing to find it. If there is a condition that defaults to false, I guess I'm ok moving it. I'd prefer this to be in a separate module, but if you feel strongly it should be in the same module, we can do that.

Relatedly, once https://github.com/google/exposure-notifications-verification-server/pull/493 is submitted, please also make the changes to the readme on how to enable this condition.

sherifkozman:
@icco you are correct the condition was set for lb.tf using the same variables declared for both the GCLB and the notifications.
With #477 there would be no need to move them to the root folder and we can keep them in a seprate module

PR #479, opened 2020-09-05
Fix bash command

zssz:


icco:
I mean this is fine... but why though? Could you follow our pull request template to explain why this change is needed? It's functionally equivalent and just a matter of style afaik.

/hold
/Ok-to-test

sethvargo:
`echo` isn't the same on all systems. My guess they are running a version that doesn't support heredocs.

/unhold
/lgtm

PR #483, opened 2020-09-07
Always list active signing key first

sethvargo:
**Release Note**

```release-note
Always list active signing key first in results
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -289,10 +289,10 @@ func (r *Realm) SMSProvider(db *Database) (sms.Provider, error) {
 	return provider, nil
 }
 
-// GetCurrentSigningKey returns the currently active signing key, the one marked
+// CurrentSigningKey returns the currently active signing key, the one marked
 // active in the database. If there is more than one active, the most recently
 // created one wins. Should not occur due to transactional update.
-func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
+func (r *Realm) CurrentSigningKey(db *Database) (*SigningKey, error) {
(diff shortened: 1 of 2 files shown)


pkg/database/realm.go
```diff
@@ -353,13 +353,14 @@ func (r *Realm) SetActiveSigningKey(db *Database, id uint) (string, error) {
...
-// ordered by created_at desc.
+// ListSigningKeys returns the non-deleted signing keys for a realm ordered by
+// created_at desc. They are sorted such that the "active" key is listed first,
+// followed by other keys sorted in created order.
 func (r *Realm) ListSigningKeys(db *Database) ([]*SigningKey, error) {
 	var keys []*SigningKey
 	if err := db.db.
 		Model(r).
-		Order("signing_keys.created_at DESC").
+		Order("signing_keys.active DESC, signing_keys.created_at DESC").
```
mikehelmick, line 363: I'm not sure I like that change - the active signing key is always shown at the top in the current UI.

This would make it so you create a new key and it shows lower int he list, and then the list reorders after you make it active

sethvargo: Hmm the behavior you describe is the behavior I was going for. While it shows the active key in the UI, the active key is sometimes buried in the list.

Either way, I'm not super passionate about it. If you think it's better as-is, we can leave it.

mikehelmick: I like the time ordered list and I can see quickly if I have "newer" keys that aren't active.

I think the typical rotation flow will only have one above active.

let's leave it as is and see if we can get feedback from some early users.

PR #484, opened 2020-09-07
Fix data races in tests

sethvargo:
While working on some other stuff locally and running the tests, I occasionally saw many data races. I think we've also seen these a few times on prow, so I figured I'd dig in. There were two separate data races:

1. The DefaultOptions in migrations is a global mutable. Despite always setting the value to false (and the default being false), the race detector sees this as a data-write and barfs. I removed it, since it's the default value.

2. ~In the event we failed to connect to the database, we were still registering the opencensus views. Furthermore, we could succeed connecting to the database, but fail to initialize gorm, which would trigger a re-connect. This separates the "connect to the database" and "connect to the database via gorm" steps into separate retries. It also moves the opencenusus and driver initialization into a package-level init.~

    We've temporarily reverted the OC integration.

3. The verification codes tests create verification codes in the struct with a set expiration. Those times are set on load, not "on test run". When the system is under heavy load, there could be 5-10s between when the struct is created and when the test is run. To fix this, we changed the test cases to return a _function_ that returns a verification code, ensuring the time-based tests run right before use, not on load.

**Release Note**

```release-note
Fix data race in database initialization
```

PR #485, opened 2020-09-07
Revert "Add opencensus integration"

sethvargo:
Reverts google/exposure-notifications-verification-server#396

This is causing a number of test failures and data races. I haven't had a chance to fully dig in and diagnose, but we need to revert this for now to get the test setup back in a good status.

/assign @mikehelmick


sethvargo:
Sigh. That's the actually racey test I'm trying to fix in GH-484.

/reteset

mikehelmick (approved):

PR #489, opened 2020-09-08
Remove the callback library causing race issue

### LINKED ISSUE #267: Open Census for Postgresql

taddari:
Fixes #267
Thanks Seth for catching the problem.
## Proposed Changes

* Removed the callback library that was causing race issues.
* This PR just overrides the driver, so there should be no problem.
* opencensus postgresql will show the metrics only. As there is no tracing involved, there is no need to pass the context.

```release-note
Following metrics will be available https://github.com/opencensus-integrations/ocsql#metrics
```


icco:
See also https://github.com/google/exposure-notifications-verification-server/pull/487/files

pkg/database/database.go
```diff
@@ -122,6 +126,17 @@ func (db *Database) Open(ctx context.Context) error {
 	return db.OpenWithCacher(ctx, nil)
 }
 
+const driverName = "ocsql"
```
sethvargo, line 129: I know this is what we had before, but while I was debugging this, I found a few bugs.

The registration should happen via an `init()` function at the package level:

```go
const driverName = "ocsql"

// Some docs
func init() {
  for _, v := range sql.Drivers() {
    if v == driverName {
      return
    }
  }
  ocsql.RegisterAllViews()
  sql.Register(driverName, ocsql.Wrap(&postgres.Driver{})
}
```

pkg/database/database.go
```diff
@@ -132,7 +147,21 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
 	var rawDB *gorm.DB
 	if err := withRetries(ctx, func(ctx context.Context) error {
 		var err error
-		rawDB, err = gorm.Open("postgres", c.ConnectionString())
+		driver := ocsql.Wrap(&postgres.Driver{})
```
sethvargo, line 150: Move this to package init (see my comment above)

pkg/database/database.go
```diff
@@ -132,7 +147,21 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
...
+		if !stringInSlice(driverName, sql.Drivers()) {
+			ocsql.RegisterAllViews()
+			sql.Register(driverName, driver)
+		}
+		dbSQL, err := sql.Open(driverName, c.ConnectionString())
```
sethvargo, line 155: Ah yes, this bug 😄 

This is happening in a retry loop (see L148). It's possible that the sql connection successfully opens, but then gorm fails. This would trigger the retry loop to run again, creating a new sql connection. If gorm continues to fail, we'd have ~14 open sql connections multiplied by N services and could exhaust connections.

We should move the `sql.Open` call into it's own `withRetries` block and wrap it as retryable:

```go
var rawSQL *sql.DB
if err := withRetries(ctx, func(ctx context.Context) error {
  var err error
  rawSQL, err = sql.Open(driverName, c.ConnectionString())
  if err != nil {
    return retry.Retryable(err)
  }
}); err != nil {
  return fmt.Errorf("failed to create sql connection: %w", err)
}
```

pkg/database/database.go
```diff
@@ -132,7 +147,21 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
...
+		if err != nil {
+			return fmt.Errorf("failed to open the SQL database: %v", err)
+		}
+		// enable periodic recording of sql.DBStats
+		db.statsCloser = ocsql.RecordStats(dbSQL, 5*time.Second)
+
+		//Need to give postgres dialect as otherwise gorm starts running
```
sethvargo, line 162: Nit: comments should have space after `//`

pkg/database/database.go
```diff
@@ -122,6 +126,17 @@ func (db *Database) Open(ctx context.Context) error {
 	return db.OpenWithCacher(ctx, nil)
 }
 
+const driverName = "ocsql"
+
+func stringInSlice(a string, list []string) bool {
```
sethvargo, line 131: I think you can drop this function if you go with the `init` flow

pkg/database/database.go
taddari: Done, thanks for the input.

pkg/database/database.go
taddari: Done

pkg/database/database.go
taddari: Done

mikehelmick (approved):

PR #490, opened 2020-09-08
Add random padding to responses

sethvargo:
I'm open to what arbitrary upper bound we put

**Release Note**

```release-note
Include random padding bytes in API responses. Clients should not process these bytes.
```

/hold to make sure it won't break clients

/assign @mikehelmick

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -16,7 +16,12 @@
 // well as between mobile devices and the server.
 package api
 
-import "fmt"
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"math/big"
+)
 
 const (
 	// TestTypeConfirmed is the string that represents a confirmed covid-19 test.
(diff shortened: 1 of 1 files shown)


pkg/api/api.go
```diff
@@ -98,13 +103,35 @@ func (e *ErrorReturn) WithCode(code string) *ErrorReturn {
...
 // to prevent a network observer from building a model based on request or
 // response sizes.
-type Padding struct {
-	Padding string `json:"padding,omitempty"`
+type Padding []byte
+
+// MarshalJSON is a custom JSON marshaler for padding. It generates and returns
+// a random number of base64-encoded bytes.
+func (p Padding) MarshalJSON() ([]byte, error) {
+	bi, err := rand.Int(rand.Reader, big.NewInt(256))
```
mikehelmick, line 111: I think random in a range would be good - 1-2kb maybe

mikehelmick: i.e. the padding is enough to obscure any actual responses

mikehelmick (approved):

PR #493, opened 2020-09-08
Update Terraform README re using alerting rules

### LINKED ISSUE #477: Creation of alerting Policies will fail because monitoring workspace must be manually enabled
### TL;DR
Creation of alerting Policies will fail because monitoring workspace must be manually enabled.

### Details : 
GCP Monitoring Workspace have no support in terraform yet and should be enabled manually first before running the Alerting and notifications terraform scripts.

README should be updated to include such note and manual step of enabling a monitoring workspace.

yegle:
Fixes #477 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -8,11 +8,12 @@ continuous deployment.
 
 ## Requirements
 
-- Terraform 0.13.1 [Installation guide](https://www.terraform.io/downloads.html)
+-   Terraform 0.13.1
+    [Installation guide](https://www.terraform.io/downloads.html)
 
-- firebase-cli. [Installation guide](https://firebase.google.com/docs/cli)
+-   firebase-cli. [Installation guide](https://firebase.google.com/docs/cli)
 
-- gcloud. [Installation guide](https://cloud.google.com/sdk/install)
+-   gcloud. [Installation guide](https://cloud.google.com/sdk/install)
 
     Note: Make sure you **unset** `GOOGLE_APPLICATION_CREDENTIALS` in your
     environment:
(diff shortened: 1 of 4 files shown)


terraform/README.md
```diff
@@ -146,6 +158,12 @@ Terraform module.
     inputs/parameters to the module definition. See the `variables.tf` file for
     the full list of configuration options.
 
+1.  (Optional if alerting is enabled): Manually create StackDriver monitoring
```
sethvargo, line 161: Should be Cloud, not Stackdriver

terraform/alerting/workspace.tf
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "null_resource" "manual-step-to-enable-workspace" {
+  # TODO: remove once
+  # https://github.com/hashicorp/terraform-provider-google/issues/2605 is
+  # fixed.
+  provisioner "local-exec" {
+    command = <<EOF
+    echo -e '\033[0;31m>>>> WARNING WARNING WARNING!'
```
sethvargo, line 21: Let's not rely on ANSI sequences here in case anyone is running Windows

yegle: Interesting. Do we need to worry about that? Will `echo` work in Windows?

sethvargo: Echo works in powershell but not cmd. But in general, colors and special formatting should be avoided unless strictly required.

yegle: TIL!

icco (approved):
/lgtm

PR #497, opened 2020-09-09
Add middleware for debug headers

sethvargo:
If the request sends an x-debug header with any value, the response headers will be populated with the build ID and tag.

**Release Note**

```release-note
Return build information in response headers if `X-Debug` is supplied as a request header
```

/cc @icco

### DIFF
--- a/pkg/controller/middleware/debug.go
+++ b/pkg/controller/middleware/debug.go
@@ -1,0 +1,40 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/controller/middleware/debug.go
```diff
@@ -0,0 +1,40 @@
...
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/buildinfo"
+	"github.com/gorilla/mux"
+)
+
+// ProcessDebug adds additional debugging information to the response if the
+// request included the "X-Debug" header with any value.
+func ProcessDebug(ctx context.Context) mux.MiddlewareFunc {
+	header := "x-debug"
```
mikehelmick, line 28: nit: seems like this should be a const

mikehelmick: also - I'm not sure x-debug is the right header, maybe something like x-build-info ?

sethvargo: Made them constants. The return headers are `x-buildinfo-*`. This is the _request_ header, and I'd like to give us flex to add other debug information if we wanted to in the future.

sethvargo:
@mikehelmick PTAnotherL

mikehelmick (approved):

PR #499, opened 2020-09-09
Acquire a lock before registering callbacks

sethvargo:
This should fix the sporadic data race issues we see in CI, at the expense of increasing the test suite by ~3s.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -38,6 +38,12 @@ import (
 	postgres "github.com/lib/pq"
 )
 
+var (
+	// callbackLock prevents multiple callbacks from being registered
+	// simultaneously because that's a data race in gorm.
+	callbackLock sync.Mutex
+)
+
 // Database is a handle to the database layer for the Exposure Notifications
 // Verification Server.
 type Database struct {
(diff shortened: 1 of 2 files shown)


pkg/database/database.go
```diff
@@ -171,6 +176,11 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
 	// Enable auto-preloading.
 	rawDB = rawDB.Set("gorm:auto_preload", true)
 
+	// Prevent multiple simultaneous callback registrations due to a data race in
+	// gorm.
+	db.callbackLock.Lock()
```
mikehelmick, line 181: Are there multiple goroutines calling OpenWithCacher on the same DB handle?

If that's happening, that seems wrong. 

Instead of locking the callbacks, we should just do a quick lock/unlock see if the *Database has been opened or not.

sethvargo: It's not the same db handle - see my comment in chat. It's the same callback handler. Gorm has a global `DefaultCallback` that is set on all `gorm.Open` calls. Thus the first call to `.Callback()` uses a shared global constant.

This only happens in tests because tests are the only thing that can call Open in parallel.

mikehelmick: saw that - thanks!

mikehelmick: so - should this be a global lock and not per-DB?

sethvargo: Made it global, but I forget how Go works across packages with global vars.

icco (approved):
/lgtm

PR #503, opened 2020-09-09
HMAC instead of hash keys in the cacher

sethvargo:
Fixes GH-429 (most of it is done, this was the last remaining bit)

**Release Note**

```release-note
Use HMAC instead of hashes in cacher keys
```


icco (approved):
/lgtm

Issue #304, opened 2020-08-21
Send email login to sign-up new users


whaught:
### TL;DR
New users should get an email to join, then create a password.

### Design

**Problems**
At the moment users can create a new account without being invited. They exist for auth but have not Users entry (or access to any realm). At the moment this results in a redirect loop to the sign-in page (is cookie clearing working right?)
If the user both creates a login and was invited, everything works.

This maybe a potential abuse vector: folks can spam account creation without being invited
This is also a bad experience

**Solutions**
1) When an admin creates a new user, send the new user an email sign-in link.
2) Prompt the user to create a password for future sign-in
    2a) also register 2nd factor
3) Allow admins to re-send the email
4) Deprecate the (currently ad-hoc) email verification and new user auth creation flows

**Alternatives**
Pre-create the user with a phony password and call the password reset email flow?

PR #509, opened 2020-09-09
Remove new-user link

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/304

## Proposed Changes

* Remove the link to create a new user
    * New users should be invited by an existing admin and will receive an email.
    * I'll remove the page entirely in the future. An oncaller can still help someone directly navigate to login/create if we see any problems with the email-invite flow.


mikehelmick:
Should we leave this as is for at least a full release? Check this in after v0.8 is cut?

whaught:
> Should we leave this as is for at least a full release? Check this in after v0.8 is cut?

We could just leave it for a bit. The idea here was to kill the link, but leave the page up. Then if someone screams one can still manually navigate there. Alternatively we can leave as-is for a release and kill the whole thing

mikehelmick:
> > Should we leave this as is for at least a full release? Check this in after v0.8 is cut?
> 
> We could just leave it for a bit. The idea here was to kill the link, but leave the page up. Then if someone screams one can still manually navigate there. Alternatively we can leave as-is for a release and kill the whole thing

we won't know if someone screams :)

let's 
/hold
for a week

whaught:
/reopen

didn't mean to close this one

shohansp:
Suported

pkg/controller/user/delete.go
```diff
@@ -59,7 +59,7 @@ func (c *Controller) HandleDelete() http.Handler {
 			return
 		}
 
-		flash.Alert("Successfully removed user '%v from realm'", user.Email)
+		flash.Alert("Successfully removed user '%v' from realm", user.Email)
```
sethvargo, line 62: Bad rebase?

sethvargo: Actually, looks like that quote shouldn't be anywhere in this string.

mikehelmick:
this is fine to check in now

I'll let you address Seth's comment first

pkg/controller/user/delete.go
whaught: Sorry this is kind of unrelated, just saw the message at one point while testing and didn't like the weird quotes

mikehelmick (approved):

PR #510, opened 2020-09-10
Build before migrating

sethvargo:
Fixes GH-498

**Release Note**

```release-note
NONE
```


sethvargo:
Same change is _not_ required on the key server since I apparently remembered to add it there.

icco (approved):
/lgtm

Issue #506, opened 2020-09-09
Allow realms to require 2FA


mikehelmick:
### TL;DR

All realms to require 2FA before users can issue codes.

/kind enhancement
/assign @whaught

whaught:
The trouble is we are using the following from our server: https://pkg.go.dev/firebase.google.com/go/v4/auth
However I don't believe any such golang client library exists for Identity Platform https://cloud.google.com/identity-platform/docs/admin/manage-mfa-users#before_you_begin. The REST API is backwards compatible from but we need the added the 'multiFactor' field.

Looks like it can all be done in node.js and they are guiding folks to use Functions. Confusing - still looking around.

whaught:
Sure enough: https://github.com/firebase/firebase-admin-node/blob/f11aed754d30fabb2b936e9b8958669c65750f5f/src/auth.d.ts#L217

Not present in other clients: https://github.com/firebase/firebase-admin-go/blob/cef91acd46f2fc5d0b3408d8154a0005db5bdb0b/auth/user_mgt.go#L71

PR #355, opened 2020-08-25
Cache users, api keys, and realms for 30s

sethvargo:
**Release Note**

```release-note
Add caching layer for users, API keys, and realms
```

PR #511, opened 2020-09-10
Include auth factor count in session

whaught:
Issue #https://github.com/google/exposure-notifications-verification-server/issues/506

## Proposed Changes

* Here we verify the token and retrieve the user from cache as we create the auth session
    * Similar logic was removed with https://github.com/google/exposure-notifications-verification-server/pull/355 so looking to see if this was intended @sethvargo 

Next this is going to store the MFA state of the user if they have other factors. This will be used to require mfa per-realm. This solution isn't ideal but firebase doesn't provide a go client for managing mfa yet.

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -371,6 +371,7 @@
     switch (errLevel) {
       case "warning":
       case "danger":
+        break;
       case "success":
         if (dismissible === undefined) {
           dismissible = true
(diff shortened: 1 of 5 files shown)


whaught:
/hold

To run this by @sethvargo - we removed 'VerifyCookieAndUser' in another PR

sethvargo:
I removed it because it causes double-duty 😄. The first request is a login, which saves the firebase cookie. Then, we redirect to a page that requires auth. Since that page has middleware attached, it runs the gambit of checks and returns an error (or redirects) if the user isn't valid, firebase token is invalid, etc. It's just to reduce code duplication honestly.

whaught:
/close

I have different ideas now

PR #512, opened 2020-09-10
Allow realm to specify if MFA is required, prompt, or optional

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/506

## Proposed Changes

*  Create a new column in Realm for MFA mode
    * For now we have Required, Prompt, and Optional
* Required auth always redirects back to registration page
* Move redirect to /realm page for prompt

TODO:
In the future we should grab the factors from the Firebase User instead of storing on session

### DIFF
--- a/cmd/server/assets/realm.html
+++ b/cmd/server/assets/realm.html
@@ -133,6 +133,17 @@
               </div>
             </div>
           </div>
+
+          <div class="form-group row">
+            <label for="MFAMode" class="col-sm-3">Multi factor auth:</label>
+            <div class="col-sm-9">
+              <select name="MFAMode" id="MFAMode" class="form-control">
+                <option value=1 {{if eq $realm.MFAMode 1}}selected{{end}}>Required</option>
+                <option value=0 {{if eq $realm.MFAMode 0}}selected{{end}}>Prompt after login</option>
+                <option value=2 {{if eq $realm.MFAMode 2}}selected{{end}}>Optional</option>
+              </select>
+            </div>
+          </div>
         </div>
       </div>
 
@@ -323,7 +334,7 @@
           <div class="form-group row">
             <label for="twilio_auth_token" class="col-sm-3">Twilio auth token:</label>
             <div class="input-group col-sm-9">
-              <input type="password" id="twilio_auth_token" name="twilio_auth_token" class="form-control" value="{{if $smsConfig}}{{$smsConfig.TwilioAuthToken}}{{end}}">
+              <input type="password" id="twilio_auth_token" name="twilio_auth_token" class="form-control" autocomplete="new-password" value="{{if $smsConfig}}{{$smsConfig.TwilioAuthToken}}{{end}}">
               <div class="input-group-append">
                 <a class="input-group-text" data-toggle-password="twilio_auth_token">
                   <span class="oi oi-lock-locked" aria-hidden="true"></span>
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/google/exposure-notifications-verification-server/pkg/api"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
-	"github.com/gorilla/sessions"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 )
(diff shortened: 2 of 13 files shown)


pkg/controller/middleware/realm.go
```diff
@@ -89,6 +89,12 @@ func RequireRealm(ctx context.Context, cacher cache.Cacher, db *database.Databas
 				return
 			}
 
+			if realm.MFAMode == database.MFARequired {
+				if factors := controller.FactorCountFromSession(session); factors == 0 {
+					http.Redirect(w, r, "/login/registerphone", http.StatusSeeOther)
```
sethvargo, line 94: Need a `return`

pkg/controller/realm/index.go
```diff
@@ -48,8 +49,20 @@ func (c *Controller) HandleIndex() http.Handler {
...
+				http.Redirect(w, r, "/home", http.StatusFound)
+				return
+			}
+
 			controller.StoreSessionRealm(session, realm)
 			flash.Alert("Logged into verification system for '%s'", realm.Name)
+
+			if realm.MFAMode == database.MFAOptionalPrompt {
+				if factors := controller.FactorCountFromSession(session); factors == 0 {
+					http.Redirect(w, r, "/login/registerphone", http.StatusSeeOther)
```
sethvargo, line 62: Need `return`

sethvargo: It'd also be nice to move this logic into the controller since I've seen it at least twice (e.g. `controller.RedirectToMFA`)

pkg/database/migrations.go
```diff
@@ -858,6 +858,23 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00035-AddMFARequiredToRealms",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Debugw("adding MFA required to realm")
+
+				if err := tx.Exec("ALTER TABLE realms ADD COLUMN IF NOT EXISTS mfa_mode INTEGER DEFAULT 0").Error; err != nil {
```
sethvargo, line 866: Nit: you can just `return tx.Exec().Error` here and in rollback

pkg/database/realm.go
```diff
@@ -71,6 +83,8 @@ type Realm struct {
 	// SMS Content
 	SMSTextTemplate string `gorm:"type:varchar(400); not null; default: 'This is your Exposure Notifications Verification code: ens://v?r=[region]&c=[longcode] Expires in [longexpires] hours'"`
 
+	MFAMode MFAMode `gorm:"type:smallint; not null; default: 0"`
```
sethvargo, line 86: Doc

sethvargo (commented):
This seems good to me. The one thing I'm worried about is the delta time between when a realm admin says "require MFA" and when we enforce it. Right now it's **RIGHT NOW**, and that could cause people to get locked out of their accounts.

pkg/database/realm.go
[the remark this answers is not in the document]
whaught: The three are required, optional, and off. I was considering another where we make it permanently skippable (prompt once) - this leave us open to whatever we want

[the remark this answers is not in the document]
whaught: to @sethvargo 's point we could create a variant that prompts and then requires after a user is a certain age to give a grace period to get it set up

cmd/server/assets/realm.html
```diff
@@ -323,7 +334,7 @@ <h1>Realm settings</h1>
           <div class="form-group row">
             <label for="twilio_auth_token" class="col-sm-3">Twilio auth token:</label>
             <div class="input-group col-sm-9">
-              <input type="password" id="twilio_auth_token" name="twilio_auth_token" class="form-control" value="{{if $smsConfig}}{{$smsConfig.TwilioAuthToken}}{{end}}">
+              <input type="password" id="twilio_auth_token" name="twilio_auth_token" class="form-control" autocomplete="new-password" value="{{if $smsConfig}}{{$smsConfig.TwilioAuthToken}}{{end}}">
```
sethvargo, line 337: What's this do?

pkg/controller/controller.go
```diff
@@ -115,6 +116,15 @@ func MissingUser(w http.ResponseWriter, r *http.Request, h *render.Renderer) {
 	InternalError(w, r, h, errMissingUser)
 }
 
+// RedirectToMFA redirects to MFA if multifactor has not been enabled.
+func RedirectToMFA(session *sessions.Session, w http.ResponseWriter, r *http.Request, h *render.Renderer) bool {
```
sethvargo, line 120: Why return a boolean?

cmd/server/assets/realm.html
whaught: prevents chrome from autofilling my username/password into this field

sethvargo: Shouldn't it be `autocomplete="off"`

whaught: Chrome aggressively ignores autocomplete="off". Don't know why.

https://stackoverflow.com/questions/57367813/2019-chrome-76-approach-to-autocomplete-off

It's been the cause of much internet lament. Looks like folks started using non-standard values like autocomplete="nope" to try to trick it, but now it ignores that too.

https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164

sethvargo: That's the saddest thing.

sethvargo (approved):
/lgtm

PR #513, opened 2020-09-10
Switch to APIKey+IP rate limiting, HMAC values

sethvargo:
This changes rate limiting for API keys to be "per API key realm, per IP" to reduce the chance of someone being able to DOS the system if they were to uncover an API key.

It also switches to HMAC (instead of just hash) the values before they are stored in Redis.

**Release Note**

```release-note
**Major change!** Change rate limiting for API keys to rate limit by "Realm + IP" to reduce the chance of a DOS attack. Re-evaluate your rate limits to ensure they still make sense in this new model.
```

### DIFF
--- a/pkg/ratelimit/limitware/middleware.go
+++ b/pkg/ratelimit/limitware/middleware.go
@@ -239,7 +239,7 @@ func UserIDKeyFunc(ctx context.Context, scope string, hmacKey []byte) httplimit.
 			if err != nil {
 				return "", fmt.Errorf("failed to digest user id: %w", err)
 			}
-			return fmt.Sprintf("%suser:%x", scope, dig), nil
+			return fmt.Sprintf("%suser:%s", scope, dig), nil
 		}
 
 		return ipAddrLimit(r)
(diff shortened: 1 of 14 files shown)


pkg/ratelimit/limitware/middleware.go
```diff
@@ -197,94 +196,88 @@ func (m *Middleware) Handle(next http.Handler) http.Handler {
 }
 
 // APIKeyFunc returns a default key function for ratelimiting on our API key
-// header. It falls back to rate limiting by the client ip.
-func APIKeyFunc(ctx context.Context, scope string, db *database.Database) httplimit.KeyFunc {
+// header. Since APIKeys are assumed to be "public" at some point, they are
+// rate limited by [api-key,ip].
+func APIKeyFunc(ctx context.Context, scope string, hmacKey []byte) httplimit.KeyFunc {
```
mikehelmick, line 201: we still need to put this in a realm level bucket and limit the number of verification checks on the realm over a period of time.

the idea was that we don't one one or a small set of IPs to eat all the quota for the realm.

sethvargo: Done

sethvargo:
@mikehelmick fixed - PTanotherL

icco (approved):
/lgtm

Issue #461, opened 2020-09-02
Use Terraform to manage our alerting rules


yegle:
### TL;DR
Similar to #460, we should also manage our alerting rules using Terraform so that the work can be replicated easily.

### Design

Use https://www.terraform.io/docs/providers/google/r/monitoring_alert_policy.html to manage alerting policies in StackDriver.

sethvargo:
> alerting policies in StackDriver.

alerting policies in the _product formerly known as_ StackDriver.

icco:
There are two alerts we should write:

 - Elevated 5xx
 - Elevated rate limited requests

I've got the 5xx if you want to write the rate limit one @yegle ?

yegle:
Sure I can take the rate limit requests one.

yegle:
Consider this fixed.

PR #515, opened 2020-09-10
5xx Alert

icco:
## Proposed Changes

* Add an elevated 5xx alert for #461
* Upgrades terraform version constraints to catch a bug on google_storage_bucket

**Release Note**

```release-note
Add an elevated 5xx requests to our alerting module.
```

### DIFF
--- a/terraform/alerting/monitoring.tf
+++ b/terraform/alerting/monitoring.tf
@@ -36,7 +36,7 @@ resource "google_monitoring_alert_policy" "five_xx" {
     display_name = "Elevated 5xx on Verification Server"
     condition_threshold {
       duration        = "300s"
-      threshold_value = 0.2
+      threshold_value = 2
       comparison      = "COMPARISON_GT"
       filter          = "metric.type=\"run.googleapis.com/request_count\" resource.type=\"cloud_run_revision\" metric.label.\"response_code_class\"=\"5xx\" resource.label.\"service_name\"!=\"e2e-runner\""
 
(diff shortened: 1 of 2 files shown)


terraform/alerting/monitoring.tf
```diff
@@ -27,3 +27,49 @@ resource "google_monitoring_dashboard" "e2e" {
...
+  project      = var.project
+  display_name = "Elevated 5xx"
+  combiner     = "OR"
+  conditions {
+    display_name = "Elevated 5xx on Verification Server"
+    condition_threshold {
+      duration        = "300s"
+      threshold_value = 2
+      comparison      = "COMPARISON_GT"
+      filter          = "metric.type=\"run.googleapis.com/request_count\" resource.type=\"cloud_run_revision\" metric.label.\"response_code_class\"=\"5xx\" resource.label.\"service_name\"!=\"e2e-runner\""
```
yegle, line 41: Should we also define these as YAML files?

icco: Hmm, I guess we could? hadn't thought about it. How about I do in a follow up PR?

terraform/alerting/monitoring.tf
```diff
@@ -27,3 +27,49 @@ resource "google_monitoring_dashboard" "e2e" {
     null_resource.manual-step-to-enable-workspace
   ]
 }
+
+resource "google_monitoring_alert_policy" "five_xx" {
+  project      = var.project
+  display_name = "Elevated 5xx"
```
yegle, line 33: This should be a string without space to make it easier to link to the playbook.

E.g. we can add "See [docs/${policy.display_name}](...)" to all alerts.

terraform/alerting/monitoring.tf
yegle: SG

Issue #519, opened 2020-09-10
Make a date required for code issue


mikehelmick:
### TL;DR

Add a per-realm setting that _requires_ that at least one date is captured when issuing a verification code. It doesn't matter if it is test date or symptom onset date.

The default should remain false / current behavior.

This setting needs to impacts both UI and API for issuing codes (probably new return code for API).

/kind enhancement

PR #520, opened 2020-09-10
Add realm configuration for requiring dates

sethvargo:
Fixes GH-519

This adds a new realm-specific configuration for requiring a date. The default is false and backfilled as false.

**Release Note**

```release-note
Add realm configuration for requiring a date when generating a verification code. This includes a new `missing_date` error code in the issue API.
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -893,6 +893,26 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00037-AddRealmRequireDate",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Debugw("db migrations: adding require_date to realms")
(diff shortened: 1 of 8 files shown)


pkg/database/migrations.go
```diff
@@ -893,6 +893,26 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
```
sethvargo, line 896: Why not `gorm.AutoMigrate`? Because you can't set a NOT NULL constraint and DEFAULT at the same time when there are existing rows in the database. Postgres creates them with NULL, then backfills the default... but that violates the NOT NULL constraint. So we have to create the column as nullable, wait for the tx to finish, then mark it as not nullable.

whaught (commented):
Nice - this pr is giving me deja-vu from require MFA

/lgtm

PR #522, opened 2020-09-11
Enforce realm MFA  as middleware

### LINKED ISSUE #506: Allow realms to require 2FA

whaught:
Fixes #506

## Proposed Changes

*  Enforces realm MFA with middleware, sets a  session bit so users are never prompted twice.
* Disable skip link for realms where MFA is required

### DIFF
--- a/cmd/server/assets/login/registerphone.html
+++ b/cmd/server/assets/login/registerphone.html
@@ -50,7 +50,7 @@
           </div>
         </form>
 
-        {{ if ne .mfamode 1}}
+        {{if ne .currentRealm.MFAModeString "required"}}
         <a id="skip" class="card-link" href="/home">Skip for now</a>
         {{end}}
       </div>
@@ -62,7 +62,7 @@
   {{template "scripts" .}}
 
   <script>
-        {{ if eq .mfamode 1 }}
+        {{if eq .currentRealm.MFAModeString "required"}}
         flash("This realm requires multi-factor authorization.","warning")
         {{end}}
 
--- a/cmd/server/assets/login/registerphone.html
+++ b/cmd/server/assets/login/registerphone.html
@@ -50,7 +50,9 @@
           </div>
         </form>
 
-        <a id="skip" class="card-link" href="/realm">Skip for now</a>
+        {{ if ne .mfamode 1}}
+        <a id="skip" class="card-link" href="/home">Skip for now</a>
+        {{end}}
       </div>
     </div>
 
@@ -60,6 +62,10 @@
   {{template "scripts" .}}
 
   <script>
+        {{ if eq .mfamode 1 }}
+        flash("This realm requires multi-factor authorization.","warning")
+        {{end}}
+
     window.recaptchaVerifier = new firebase.auth.RecaptchaVerifier(
       'recaptcha-container',
       { 'size': 'invisible' }
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -367,7 +367,6 @@ func realMain(ctx context.Context) error {
 		adminSub.Use(requireAuth)
 		adminSub.Use(requireVerified)
 		adminSub.Use(requireSystemAdmin)
-		adminSub.Use(requireMFA)
 		adminSub.Use(rateLimit)
 
 		adminController := admin.New(ctx, config, db, h)
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -188,6 +188,7 @@ func realMain(ctx context.Context) error {
 	requireAdmin := middleware.RequireRealmAdmin(ctx, h)
 	requireRealm := middleware.RequireRealm(ctx, cacher, db, h)
 	requireSystemAdmin := middleware.RequireAdmin(ctx, h)
+	requireMFA := middleware.RequireMFA(ctx, h)
 	rateLimit := httplimiter.Handle
 
 	{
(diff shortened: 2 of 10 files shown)


cmd/server/main.go
```diff
@@ -227,6 +228,7 @@ func realMain(ctx context.Context) error {
 			sub.Use(rateLimit)
 			sub.Use(requireAuth)
 			sub.Use(requireVerified)
+			sub.Use(requireRealm)
```
mikehelmick, line 231: is this moving MFA after realm selection?

i'm not sure why this is added in this chain

pkg/controller/login/register.go
```diff
@@ -16,17 +16,38 @@
...
+
+		session := controller.SessionFromContext(ctx)
+		if session == nil {
+			err := fmt.Errorf("session does not exist in context")
+			logger.Errorw("failed to get session", "error", err)
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		realm := controller.RealmFromContext(ctx)
```
mikehelmick, line 39: instead of requiring a realm  - why not bounce the user back here after they select a realm that requires MFA and include that in the flash messages?

Then we can send all users here as a suggestion on first login

pkg/controller/middleware/mfa.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+
+	"github.com/google/exposure-notifications-server/pkg/logging"
+
+	"github.com/gorilla/mux"
+	"github.com/gorilla/sessions"
+)
+
+// RequireMFA checks the realm's MFA requirements and enforces them.
+// If no realm is selected, this assumes MFA is required.
```
mikehelmick, line 33: add that this should be installed after the require realm controller

pkg/controller/login/register.go
whaught: because then they can just navigate to /home, this enforces it

cmd/server/main.go
whaught: Currently registerphone doesn't include a selected realm, so we don't know which realm settings to enforce! Now we have the user first select the realm and then gets here

cmd/server/assets/login/registerphone.html
```diff
@@ -50,7 +50,9 @@ <h1>SMS auth registration</h1>
           </div>
         </form>
 
-        <a id="skip" class="card-link" href="/realm">Skip for now</a>
+        {{ if ne .mfamode 1}}
```
sethvargo, line 53: This feels too magical to me. Could we add a method on the realm like `RequiresMFA` or something so this becomes:

```
{{ if $currentRealm.RequireMFA }}
```

"1" just feels magical and isn't very descriptive.

cmd/server/main.go
sethvargo: Makes sense to me

pkg/controller/login/register.go
```diff
@@ -16,17 +16,38 @@
...
 )
 
 func (c *Controller) HandleRegisterPhone() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+		logger := logging.FromContext(ctx).Named("login.HandleRegisterPhone")
+
+		session := controller.SessionFromContext(ctx)
+		if session == nil {
+			err := fmt.Errorf("session does not exist in context")
```
sethvargo, line 33: Please use:

```go
controller.MissingSession(w, r, c.h)
return
```

instead

pkg/controller/middleware/mfa.go
```diff
@@ -0,0 +1,76 @@
...
+func RequireMFA(ctx context.Context, h *render.Renderer) mux.MiddlewareFunc {
+	logger := logging.FromContext(ctx).Named("middleware.RequireMFA")
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+
+			session := controller.SessionFromContext(ctx)
+			if session == nil {
+				err := fmt.Errorf("session does not exist in context")
```
sethvargo, line 44: Please use `controller.MissingSession` instead

pkg/controller/middleware/mfa.go
```diff
@@ -0,0 +1,76 @@
...
+				controller.RedirectToMFA(w, r, h)
+				return
+			}
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func NeedsMFARedirect(session *sessions.Session, realm *database.Realm) bool {
```
sethvargo, line 61: What's the case where `realm` would be nil here? It seems like we should never call with a nil realm or, if we do, that's an error?

sethvargo (commented):
A few questions

pkg/controller/middleware/mfa.go
whaught: It is unlikely we'd ever want to call this without a realm, but I've made it resilient to that. The case might be if you wanted to make the /admin pages require MFA (which includes realm creation/management where none is loaded)

cmd/server/assets/login/registerphone.html
whaught: magic numbers begone

sethvargo:
/lgtm
/hold

@whaught not sure if you have other stuff you wanna add in. If not, you can remove this hold 😄

PR #527, opened 2020-09-11
Backfill realm stats based on user and apikey stats

sethvargo:
This uses the existing stats we've been collecting to backfill them onto the realm stats. It only includes backfill up to 9/15, which is when we'll cut a new release.

**Release Note**

```release-note
Add backfill migration for realm stats
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -913,6 +913,57 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return tx.Exec("ALTER TABLE realms ALTER COLUMN require_date SET NULL").Error
 			},
 		},
+		{
+			ID: "00039-RealmStatsToDate",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Debugw("db migrations: changing realm stats to date")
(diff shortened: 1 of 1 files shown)


mikehelmick (approved):

pkg/database/migrations.go
```diff
@@ -913,6 +913,57 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return tx.Exec("ALTER TABLE realm_stats ALTER COLUMN date TYPE timestamp with time zone").Error
+			},
+		},
+		{
+			ID: "00040-BackfillRealmStats",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Debugw("db migrations: back-filling realm stats")
+				sqls := []string{
```
MushuEE, line 930: I don't think it is worth the over haul now but would it be safer and more performant to use [batch calls?](https://github.com/jackc/pgx/blob/040df1ccef857617d44d14fc87f1e4091f68df77/batch_test.go#L15)

*I just started learning about this so I don't have a good understanding of the true advantages of removing round trip to database between calls.

PR #530, opened 2020-09-11
Separate loading realm and require realm

whaught:
## Proposed Changes

* Currently we lose the selected realm for /admin/realm and /realm this causes the navigation bar to lose all of its options which makes for an unusual experience.
    * Instead always load the selected realm (if one is selected) and only enforce that one be selected for pages that require it

**Release Note**

```release-note
Keep the realm selection during realm admin tasks
```


pkg/controller/middleware/realm.go
```diff
@@ -57,7 +49,7 @@ func RequireRealm(ctx context.Context, cacher cache.Cacher, db *database.Databas
 			realmID := controller.RealmIDFromSession(session)
 			if realmID == 0 {
 				logger.Debugw("realm does not exist in session")
-				controller.MissingRealm(w, r, h)
```
sethvargo, line 60: We should comment this noting why we call `next.ServerHTTP`

sethvargo (approved):
/lgtm

I'd like a comment in LoadCurrentRealm since it's a departure from our current pattern, but that can be a followup

PR #531, opened 2020-09-11
Add some comments about LoadRealm

whaught:
## Proposed Changes

* Added some comments in about LoadRealm.
* Only RequireAdmin or RequireRealm is needed, not both


pkg/controller/middleware/realm.go
```diff
@@ -49,6 +49,8 @@ func LoadCurrentRealm(ctx context.Context, cacher cache.Cacher, db *database.Dat
 			realmID := controller.RealmIDFromSession(session)
 			if realmID == 0 {
 				logger.Debugw("realm does not exist in session")
+				// If no realm on session, continue serving.
+				// If realm is non-optional the caller should RequireRealm orRequireAdmin.
```
mikehelmick: space after or

mikehelmick (approved):

PR #533, opened 2020-09-11
More consistent UI and project structure

sethvargo:
Been trying to clean this up for awhile. I went through all our HTML pages and made their styles consistent with light background, card bodies with shadow, form spacing, etc.

**Release Note**

```release-note
Refreshed various UI elements
```


sethvargo:
Since HTML changes indentation: https://github.com/google/exposure-notifications-verification-server/pull/533/files?w=1

cmd/server/assets/codestatus/show.html
```diff
@@ -18,41 +18,44 @@ <h1>Verification code status</h1>
...
+          <h5 class="mb-1">Test type</h5>
+          <p class="mb-1">{{.code.TestType}}</p>
+        </div>
+        <div class="list-group-item">
+          <h5 class="mb-1">Status</h5>
+          <p class="mb-1">{{.code.Status}}</p>
+        </div>
+        <div class="list-group-item">
+          <h5 class="mb-1">Expiry</h5>
+          <span id="code-expires-at" class="sm text-danger">&nbsp;</span>
```
sethvargo, line 44: Empty elements have no height, so this created a weird FOUT on page load. Giving it a space solves that problem.

cmd/server/assets/login/select-realm.html
```diff
@@ -0,0 +1,75 @@
+{{define "login/select-realm"}}
```
sethvargo, line 1: Sigh. I don't know why `git mv` never works for me 😦

sethvargo: Anyway, this is pretty much a straight move into the login folder.

pkg/controller/login/select.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+)
+
+func (c *Controller) HandleSelectRealm() http.Handler {
```
sethvargo, line 25: Moved this into login package and made it a single function that renders the form or processes the selection based on the request method like we do elsewhere.

whaught (commented):
/lgtm

I'm playing with this locally - if I see any nits we can just follow up, but hard to review directly

Issue #517, opened 2020-09-10
Bulk user creation


mikehelmick:
Allow realms to bulk-create user accounts.

Possibly paste in or upload a csv

/kind enhancement

whaught:
/close

Closing this for now. It's implemented, but I want to wait for feedback from users on the experience before going further. It uses html5 file reader to do the batching which is nice because we both want to create the user and send the email (the latter happens from the client in JS from firebase) which means we can't process all server-side. Doing it in small chunks prevents a weird fail-in-the-middle where the users are created but no invites are sent yet.

That said html5 may not work on super old browsers, for example.

PR #539, opened 2020-09-14
Create empty page for user import.

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/517

## Proposed Changes

* Create empty pages for user import
* Fix up users page card, put in an import button
* Fix a few back-links
users

importplaceholder

### DIFF
--- a/cmd/server/assets/realmstats.html
+++ b/cmd/server/assets/realmstats.html
@@ -5,6 +5,7 @@
 
 <!doctype html>
 <html lang="en">
+
 <head>
   {{template "head" .}}
 </head>
@@ -20,7 +21,7 @@
       The data below shows realm statistics.
     </p>
 
-    <div class="card mb-3 shadow-sm" style="min-width:950px;">
+    <div class="card mb-3 shadow-sm">
       <div class="card-header">Statistics</div>
       <div class="card-body table-responsive">
         {{if $stats}}
@@ -49,7 +50,7 @@
           This data is refreshed every 5 minutes.
         </div>
         {{else}}
-          <p>No codes have been issued in this realm.</p>
+        <p>No codes have been issued in this realm.</p>
         {{end}}
       </div>
     </div>
@@ -61,28 +62,29 @@
   {{if $stats}}
   <script src="https://www.gstatic.com/charts/loader.js"></script>
   <script>
-    google.charts.load('current', {packages: ['line']});
+    google.charts.load('current', { packages: ['line'] });
     google.charts.setOnLoadCallback(drawChart)
 
     function drawChart() {
       var data = google.visualization.arrayToDataTable([
         ['Date', 'Issued', "Claimed"],
-        {{range $stat := $stats}}
-        ['{{$stat.Date.Format "Jan 2"}}', {{$stat.CodesIssued}}, {{$stat.CodesClaimed}}],
-        {{end}}
+        {{ range $stat:= $stats }}
+    ['{{$stat.Date.Format "Jan 2"}}', {{ $stat.CodesIssued }}, {{ $stat.CodesClaimed }}],
+      {{ end }}
       ]);
 
-      var options = {
-        colors: ['#007bff', '#ff7b00'],
-        legend: {position: 'bottom'},
-        tooltip: {trigger: 'focus'},
-      };
+    var options = {
+      colors: ['#007bff', '#ff7b00'],
+      legend: { position: 'bottom' },
+      tooltip: { trigger: 'focus' },
+    };
 
-      var chart = new google.charts.Line(document.getElementById('chart'));
-      chart.draw(data, google.charts.Line.convertOptions(options));
+    var chart = new google.charts.Line(document.getElementById('chart'));
+    chart.draw(data, google.charts.Line.convertOptions(options));
     }
   </script>
   {{end}}
 </body>
+
 </html>
 {{end}}
--- a/cmd/server/assets/realmstats.html
+++ b/cmd/server/assets/realmstats.html
@@ -20,7 +20,7 @@
       The data below shows realm statistics.
     </p>
 
-    <div class="card mb-3 shadow-sm">
+    <div class="card mb-3 shadow-sm" style="min-width:950px;">
       <div class="card-header">Statistics</div>
       <div class="card-body table-responsive">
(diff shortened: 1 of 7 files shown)


cmd/server/assets/users/index.html
```diff
@@ -19,59 +19,62 @@
...
-            <td class="text-center">
-              {{if .CanAdminRealm $currentRealm.ID}}
-                <span class="badge badge-pill badge-primary">Admin</span>
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Users</div>
+      <div class="card-body">
+        <a class="btn btn-primary" href="/users/new">Create a new user</a>
+        <a class="btn btn-primary disabled" href="/users/import">Import users</a>
+        {{if $users}}
+        <div class="table-responsive mt-3">
```
sethvargo, line 30: Part of the reason I avoided this was because the table gets all messed up on mobile. What does this look like on mobile?

whaught: Added in a min-width style to prevent the weird internal horizontal scroll on the card

cmd/server/assets/realmstats.html
```diff
@@ -20,7 +20,7 @@ <h1>Realm stats</h1>
       The data below shows realm statistics.
     </p>
 
-    <div class="card mb-3 shadow-sm">
+    <div class="card mb-3 shadow-sm" style="min-width:950px;">
```
sethvargo, line 23: Can be a follow-up, but it'd be good to make this a class.

cmd/server/assets/users/index.html
```diff
@@ -19,69 +19,74 @@
...
-            <td>
-              <a href="/users/{{.ID}}">{{.Name}}</a>
-            </td>
-            <td>
-              {{.Email}}
-            </td>
-            <td class="text-center">
-              {{if .CanAdminRealm $currentRealm.ID}}
-              <span class="badge badge-pill badge-primary">Admin</span>
+    <div class="card mb-3 shadow-sm" style="min-width:550px;">
```
sethvargo, line 24: Same note about min-width here. Can we use any of these instead? https://getbootstrap.com/docs/4.0/utilities/sizing/

cmd/server/assets/users/index.html
```diff
@@ -19,69 +19,74 @@
...
+      <div class="card-header">Users</div>
+      <div class="card-body">
+        <a class=" btn btn-primary" href="/users/new">Create a new user</a>
```
sethvargo, line 27:
```suggestion
        <a class="btn btn-primary" href="/users/new">Create a new user</a>
```

cmd/server/assets/users/index.html
```diff
@@ -19,69 +19,74 @@
...
+        <a class="btn btn-primary disabled" href="/users/import">Import users</a>
+        {{if $users}}
+        <div class="table-responsive mt-3">
```
sethvargo, line 30: We generally want all margins to be on the _bottom_ for consistency. Can you move this mt to be a mb on the preceeding element?

cmd/server/assets/users/index.html
sethvargo, line 27: If you're feeling fancy:

```html
<span class="oi oi-plus" aria-hidden="true"></span>
```

cmd/server/assets/users/index.html
```diff
@@ -19,69 +19,74 @@
...
-              {{.Email}}
-            </td>
-            <td class="text-center">
-              {{if .CanAdminRealm $currentRealm.ID}}
-              <span class="badge badge-pill badge-primary">Admin</span>
+    <div class="card mb-3 shadow-sm" style="min-width:550px;">
+      <div class="card-header">Users</div>
+      <div class="card-body">
+        <a class=" btn btn-primary" href="/users/new">Create a new user</a>
+        <a class="btn btn-primary disabled" href="/users/import">Import users</a>
```
sethvargo, line 28: If you're feeling fancy:

```html
<span class="oi oi-data-transfer-upload" aria-hidden="true"></span>
```

cmd/server/assets/users/index.html
```diff
@@ -19,69 +19,74 @@
...
-              {{- /* cannot delete yourself */ -}}
-              {{if not (eq .ID $currentUser.ID)}}
-              <a href="/users/{{.ID}}" class="d-block text-danger" data-method="DELETE"
-                data-confirm="Are you sure you want to remove '{{.Name}}'?" data-toggle="tooltip"
-                title="Remove this user">
-                <span class="oi oi-trash" aria-hidden="true"></span>
-              </a>
+            </tbody>
+          </table>
+          {{if .pages}}
```
sethvargo, line 68: We shouldn't do it in this PR, but I wonder if this makes sense to be a function of the renderer instead? Then the HTML just becomes:

```html
{{ paginate $pages }}
```

and the renderer builds this HTML. Then we can make it portable across any pages that need pagination.

cmd/server/main.go
```diff
@@ -311,6 +311,7 @@ func realMain(ctx context.Context) error {
 		userSub.Handle("", userController.HandleIndex()).Queries("offset", "{[0-9]*?}").Methods("GET")
 		userSub.Handle("", userController.HandleCreate()).Methods("POST")
 		userSub.Handle("/new", userController.HandleCreate()).Methods("GET")
+		userSub.Handle("/import", userController.HandleImport()).Methods("GET")
```
sethvargo, line 314: I think you'll want to handle `GET` and `POST` here, then switch inside the function to determine whether you want to render the form or process the input.

cmd/server/assets/realmstats.html
whaught: Actually I think  I've fixed it. we have table-responsive which (on purpose) is for tables that should scroll as the div around them collapses. To your point, I think that is weird (especially on mobile), so I've just removed that and it seems to display properly.

We might want to remove responsive tables elsewhere too

cmd/server/main.go
whaught: I'll add  that  in the PR that implements the  import functionality for-reals

sethvargo (approved):
/lgtm
/hold

Feel free to remove the hold when you want to merge

Issue #538, opened 2020-09-14
User pagination


whaught:
Users of the verification server may have hundreds to thousands of users which may eventually make our /users page unsuable. Consider including pagination / filtering to manager a larger user set.

Context:
https://github.com/google/exposure-notifications-verification-server/issues/517

/kind enhancement

PR #541, opened 2020-09-14
Implement pagination for the Users page

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/538

## Proposed Changes

* Adds pagination when there is a large number of users per realm
    * Still probably not ideal for order ~thousands

pages

**Release Note**

```release-note
Pagination for Users page
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -308,6 +308,7 @@ func realMain(ctx context.Context) error {
 
 		userController := user.New(ctx, cacher, config, db, h)
 		userSub.Handle("", userController.HandleIndex()).Methods("GET")
+		userSub.Handle("", userController.HandleIndex()).Queries("offset", "{[0-9]*?}").Methods("GET")
 		userSub.Handle("", userController.HandleCreate()).Methods("POST")
 		userSub.Handle("/new", userController.HandleCreate()).Methods("GET")
 		userSub.Handle("/{id}/edit", userController.HandleUpdate()).Methods("GET")
--- a/pkg/controller/user/index.go
+++ b/pkg/controller/user/index.go
@@ -24,7 +24,7 @@ import (
 )
 
 // pageSize is const for now, could provide options in UX
-const pageSize = 1
+const pageSize = 25
 
 type Pages struct {
 	Previous int
--- a/pkg/controller/user/index.go
+++ b/pkg/controller/user/index.go
@@ -26,11 +26,16 @@ import (
 // pageSize is const for now, could provide options in UX
 const pageSize = 25
 
+type PageLabel struct {
+	Offset int
+	Label  string
+}
+
 type Pages struct {
 	Previous int
 	Current  int
 	Next     int
-	Offsets  []int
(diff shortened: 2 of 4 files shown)


cmd/server/main.go
```diff
@@ -308,6 +308,7 @@ func realMain(ctx context.Context) error {
 
 		userController := user.New(ctx, cacher, config, db, h)
 		userSub.Handle("", userController.HandleIndex()).Methods("GET")
+		userSub.Handle("", userController.HandleIndex()).Queries("offset", "{[0-9]*?}").Methods("GET")
```
sethvargo, line 311: I think this might be better as:

```go
userSub.Handle("page/{offset:[0-9]+}", usersController.HandleIndex()).Methods("GET")
```

And then pull the pagination from `mux.Vars()`. That will give us nicer urls like `/users/page/2`

mikehelmick:
/lgtm
/approve

Based on screenshot , couple of future improvements
* would be good to have a "return to first" and "go to last" type link
* Different sorts (name or email)
* Search

pkg/controller/user/index.go
```diff
@@ -17,33 +17,107 @@ package user
...
+	Current  int
+	Next     int
+	Offsets  []PageLabel
+}
+
 func (c *Controller) HandleIndex() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+		qvar := r.FormValue("offset")
+		offset, _ := strconv.Atoi(qvar)
```
sethvargo, line 45: Need to handle this error, probably an internal server error.

pkg/controller/user/index.go
```diff
@@ -17,33 +17,107 @@ package user
...
 
 		realm := controller.RealmFromContext(ctx)
 		if realm == nil {
 			controller.MissingRealm(w, r, c.h)
 			return
 		}
 
-		users, err := realm.ListUsers(c.db)
+		count, err := realm.CountUsers(c.db)
```
sethvargo, line 53: Hmm, this is pretty expensive (count is a full table lock). I feel like we should have a `perPage` constant and then just pipe the current count into a SQL OFFSET

pkg/database/realm.go
```diff
@@ -434,10 +434,28 @@ func (r *Realm) FindAuthorizedApp(db *Database, id interface{}) (*AuthorizedApp,
...
+		Error; err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
 // ListUsers returns the list of users on this realm.
-func (r *Realm) ListUsers(db *Database) ([]*User, error) {
+func (r *Realm) ListUsers(db *Database, offset, limit int) ([]*User, error) {
+	if limit > 100 {
```
sethvargo, line 452: 100 feels arbitrary

sethvargo (changes requested):

sethvargo:
Oh looks like @mikehelmick was reviewing at the same time. I left some comments @whaught

PR #544, opened 2020-09-15
Upgrade rate limiter

sethvargo:
This upgrades us to the new limiter that supports multi-tenancy. It doesn't actually _implement_ that multi-tenancy on this server yet.

**Release Note**

```release-note
NONE
```


whaught (commented):
/lgtm

PR #547, opened 2020-09-15
Fix terraform smoke test: initiate before destroy

chaodaiG:
Terraform smoke test failed due to destroy action not able to reuse terrafrom state file, initiate terraform before destroy should fix this problem

Smoke test is failing: https://testgrid.k8s.io/googleoss-en-server#verification-terrafrom-smoke


icco (approved):
/lgtm

my bash is rusty, but I think this is fine.

Issue #534, opened 2020-09-11
Abuse detection and prevention


sethvargo:
### TL;DR
We need a way to prevent a leaked API key or leaked user credential from providing codes to verify too many TEK uploads.

Leaked user credentials is largely mitigated by requiring MFA. For ENX, we should require MFA. Leaked API keys are currently mitigated by rate limiting, but we'd like a more intelligent system.

### Proposal

- [x] Update go-limiter and go-redisstore to allow for bucket-level configuration. Right now, `maxMokens` and `fillRate` are properties of the limiter; we need to make them properties of the key. This will enable individual keys to have a different limit.

- [ ] We'll keep existing mitigations and rate limiters in place. This new limiter will only be checked after the preliminary APIKey/userID/IP based limiter. We'll only attempt to "take" from this new limiter right before a successful code issue. If successful, continue. On failure, return a well-communicated error message and error code (if called via the API).

- [x] Provide a realm-level configuration for allowed growth factor. This is a float value and is used to calculate the maximum number of codes the realm can generate in a 24h period. The total number of allowed daily codes is a polynomial calculation over all known data for the realm multiplied by this growth factor.

- [x] Provide some sane base value. This would be the min and default value for a new realm. Starting at 10 for now.

- [x] Run a new daily job (or use the cleanup job) each day at 12:00 UTC to recalculate, update the buckets, reset the counters, etc.

- [ ] Set up monitoring and alerting when realms are at certain capacity levels (e.g. 90%). These will have to go to the server operators since we don't have a way to send email to realm admins. (**input wanted**)

- [x] Realm admins can bump the quota by either increasing the growth factor or disabling the quota for a period of time.

jeremyfaller:
`MAX(codes_in_the_past_7d) * growthrate` = 0 if no tokens in last 7 days. This is a startup problem, and 1.5 is a problem if we have spikes from things like Sturgis or other large public gatherings in jurisdictions with small outbreaks to begin with (eg Maine weddings). I would suggest >= 10 as a factor if we go with constant*MAX approach.

Is it possible we can get a good approximation of the constant based on the R-factor of the disease's growth in a region? That seems like a good way to give guidance for admins.

sethvargo:
Yea, I'm gonna switch to a polynomial distribution instead.

PR #551, opened 2020-09-16
Add modeling service for abuse prevention

sethvargo:
Part of GH-534

**Release Note**

```release-note
Add modeling service for abuse detection (and prevention in the future)
```

### DIFF
--- a/cmd/modeler/main.go
+++ b/cmd/modeler/main.go
@@ -1,0 +1,114 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This server builds or re-builds the statistical models for predicting the
+// future number of codes a realm with generate for abuse prevention.
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/buildinfo"
+	"github.com/google/exposure-notifications-verification-server/pkg/config"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/modeler"
+	"github.com/google/exposure-notifications-verification-server/pkg/ratelimit"
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+	"github.com/sethvargo/go-signalcontext"
+
+	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/google/exposure-notifications-server/pkg/observability"
+	"github.com/google/exposure-notifications-server/pkg/server"
+
+	"github.com/gorilla/mux"
+)
+
+func main() {
+	ctx, done := signalcontext.OnInterrupt()
+
+	debug, _ := strconv.ParseBool(os.Getenv("LOG_DEBUG"))
+	logger := logging.NewLogger(debug)
+	logger = logger.With("build_id", buildinfo.BuildID)
+	logger = logger.With("build_tag", buildinfo.BuildTag)
+
+	ctx = logging.WithLogger(ctx, logger)
+
+	err := realMain(ctx)
+	done()
+
+	if err != nil {
+		logger.Fatal(err)
+	}
+	logger.Info("successful shutdown")
+}
+
+func realMain(ctx context.Context) error {
+	logger := logging.FromContext(ctx)
+
+	config, err := config.NewModeler(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to process config: %w", err)
+	}
+
+	// Setup monitoring
+	logger.Info("configuring observability exporter")
+	oeConfig := config.ObservabilityExporterConfig()
+	oe, err := observability.NewFromEnv(ctx, oeConfig)
+	if err != nil {
+		return fmt.Errorf("unable to create ObservabilityExporter provider: %w", err)
+	}
+	if err := oe.StartExporter(); err != nil {
+		return fmt.Errorf("error initializing observability exporter: %w", err)
+	}
+	defer oe.Close()
+	logger.Infow("observability exporter", "config", oeConfig)
+
+	// Setup database
+	db, err := config.Database.Load(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to load database config: %w", err)
+	}
+	if err := db.Open(ctx); err != nil {
+		return fmt.Errorf("failed to connect to database: %w", err)
+	}
+	defer db.Close()
+
+	// Create the renderer
+	h, err := render.New(ctx, "", config.DevMode)
+	if err != nil {
+		return fmt.Errorf("failed to create renderer: %w", err)
+	}
+
+	// Create the router
+	r := mux.NewRouter()
+
+	// Rate limiting
+	limiterStore, err := ratelimit.RateLimiterFor(ctx, &config.RateLimit)
+	if err != nil {
+		return fmt.Errorf("failed to create limiter: %w", err)
+	}
+	defer limiterStore.Close(ctx)
+
+	modelerController := modeler.New(ctx, config, db, limiterStore, h)
+	r.Handle("/", modelerController.HandleModel()).Methods("POST")
+
+	srv, err := server.New(config.Port)
+	if err != nil {
+		return fmt.Errorf("failed to create server: %w", err)
+	}
+	logger.Infow("server listening", "port", config.Port)
+	return srv.ServeHTTPHandler(ctx, r)
+}
--- a/cmd/server/assets/realm.html
+++ b/cmd/server/assets/realm.html
@@ -356,10 +356,7 @@
           Abuse prevention
         </div>
         <div class="card-body">
-          <div class="alert alert-warning">
-            <span class="oi oi-beaker"></span> This feature is still under
-            active development.
-          </div>
+          {{template "beta-notice" .}}
 
           <p>
             Abuse prevention uses the historical record of your realm's past
--- a/cmd/server/assets/realm.html
+++ b/cmd/server/assets/realm.html
@@ -364,6 +364,12 @@
             rejecting requests that fall outside of the predicted model.
           </p>
 
+          <p>
+            Without abuse protection, an attacker with a compromised credential
+            could generate many fake codes and then use those codes to
+            subsequently upload many fake keys to the system.
+          </p>
+
           <div class="form-group form-check">
             <input class="form-check-input" type="checkbox" name="abuse_prevention_enabled" id="abuse_prevention_enabled" value="1" {{if $realm.AbusePreventionEnabled}} checked{{end}}>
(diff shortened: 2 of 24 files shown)


pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+	if err := gamma.SolveQR(qr, false, beta); err != nil {
+		return fmt.Errorf("failed to solve QR: %w", err)
+	}
+
+	// Build the curve function.
+	m := gamma.RawMatrix()
+	curve := func(x float64) float64 {
+		var result float64
+		for i := len(m.Data) - 1; i >= 0; i-- {
+			result += m.Data[i] * math.Pow(x, float64(i))
```
jeremyfaller, line 165: As formulated, I believe you can eliminate the Pow call with:

`result = m.Data[i] + x * result`

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+	x := mat64.NewDense(len(a), degree+1, nil)
+	for i := range a {
+		for j, p := 0, 1.; j <= degree; j, p = j+1, p*a[i] {
+			x.Set(i, j, p)
+		}
+	}
+	return x
+}
+
+// digest returns the digest of given string as a hex-encoded string, and any
```
jeremyfaller, line 214: Nit: s/of/of a/

pkg/controller/modeler/modeler_test.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if got, want := realm.AbusePreventionLimit, uint(50); got != want {
+			t.Errorf("expected %v to be %v", got, want)
+		}
+	}
+
+	// Add some more statistics
```
jeremyfaller, line 109: Add a test case to check that the model floors to 10, please.

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// Build the list of Xs and Ys.
+	xs := make([]float64, len(ys))
+	for i := range ys {
+		xs[i] = float64(i)
+	}
+
+	// This is probably overkill, but it enables us to pick a different curve in
+	// the future, if we want.
+	degree := 2
```
jeremyfaller, line 150: Really only a comment as your degree is quite low, and probably not relevant because I believe this just ends up being a linear regression right now, but higher order polynomials could put you in a world where you're sensitive to things like weekends skewing your data, and among other things.

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+	if l := len(ys); l < 14 {
+		logger.Warnw("skipping, not enough data", "points", l)
+		return nil
+	}
+
+	// Exclude the most recent record. Depending on timezones, the "day" might not
+	// be over at 00:00 UTC, and we don't want to generate a partial model.
+	ys = ys[:len(ys)-1]
+
+	// Reverse the list - it came in reversed because we sorted by date DESC, but
```
jeremyfaller, line 136: Wanna just ask for the results ascending above? I am not enough of an SQL hacker to know if there's a query that does what you want there.

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// rebuildModel rebuilds the model for a single model.
+func (c *Controller) rebuildModel(ctx context.Context, id uint64) error {
+	logger := c.logger.Named("rebuildModel").With("id", id)
+	db := c.db.RawDB()
+
+	// Get 21 days of historical data for the realm.
+	var ys []float64
+	if err := db.
```
jeremyfaller, line 115: Won't there be some skew here if there's clock sync issues? Consider the case when this runs, and there's been a small number of datapoints entered with the old limiter value. Eg, the data looks like: {1e6,..., 1e6,10}. You'd end up in a situation where the model would regress on that spurious point. Part of that will be hidden with the "fudge factor" added to the model's result, but more careful would be to use dates that are strictly in the past.

jeremyfaller (commented):
Just some comments, feel free to ignore in the name of expeditiousness.

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,227 @@
...
+	// Save the new value back, bypassing any validation.
+	if err := db.
+		Model(&database.Realm{}).
+		Where("id = ?", id).
+		UpdateColumn("abuse_prevention_limit", uint(next)).
+		Error; err != nil {
+		return fmt.Errorf("failed to save model: %w", err)
+	}
+
+	// Update the limiter to use the new value.
```
chaodaiG, line 190: nit, should this be done in a function called something other than rebuildModel?

pkg/controller/modeler/modeler.go
sethvargo: I spent most of yesterday playing with models. Choosing a hire degree polynomial gives us a much higher r2 value, but it causes significantly skews in the data given how small the set is. 

But yea, when I did an 11th degree polynomial, it perfectly predicted all values within 5 of the curve, but then the "next" value was 14 orders of magnitude higher.

pkg/controller/modeler/modeler.go
sethvargo: We're building the regression model based on the past 21 days of data, so we need the SQL query to be "the last 21 records" (order by date DESC). I don't think there's a SQL way to ORDER BY and take the _last_ elements.

pkg/controller/modeler/modeler.go
sethvargo: See a few lines below. I'm excluding the most recent data point from calculations since we can't guarantee that it's a representative sample for the region as the timezone might not be "done" yet.

pkg/controller/modeler/modeler.go
sethvargo: Why?

pkg/controller/modeler/modeler_test.go
sethvargo: Done

pkg/controller/modeler/modeler.go
sethvargo: Done

pkg/controller/modeler/modeler.go
chaodaiG: `rebuildModel` sounds to me data processing only, and whether to update limiter or not is the decision of the caller. It's not important though, just a nit

pkg/controller/modeler/modeler.go
jeremyfaller: Excellent. Of my comments, this was the most pressing. I consider this resolved.

pkg/controller/modeler/modeler.go
sethvargo: Ah got it

cmd/server/assets/realm.html
```diff
@@ -356,10 +356,7 @@ <h1>Realm settings</h1>
...
         </div>
         <div class="card-body">
-          <div class="alert alert-warning">
-            <span class="oi oi-beaker"></span> This feature is still under
-            active development.
-          </div>
+          {{template "beta-notice" .}}
 
           <p>
             Abuse prevention uses the historical record of your realm's past
```
mikehelmick, line 362: I think it might be better if this say why in addition to what.

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// HandleModel accepts an HTTP trigger and re-generates the models.
+func (c *Controller) HandleModel() http.Handler {
+	logger := c.logger.Named("HandleModel")
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		if err := c.rebuildModels(ctx); err != nil {
```
mikehelmick, line 67: should we rate limit / lock this?

similar to cleanup job?

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,210 @@
...
+// rebuildModels iterates over all models with abuse prevention enabled,
+// calculates the new limits, and updates the new limits.
+func (c *Controller) rebuildModels(ctx context.Context) error {
+	logger := c.logger.Named("rebuildModels")
+	db := c.db.RawDB()
+
+	// Get all realm IDs in a single operation so we can iterate realm-by-realm to
+	// avoid a full table lock during stats calculation.
+	var ids []uint64
+	if err := db.
```
mikehelmick, line 86: don't love that this query is here and not in the DB package.

c.db.AbuseProtectionEnabledRealms() - or a better name

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,210 @@
...
+		if err := c.rebuildModel(ctx, id); err != nil {
+			merr = multierror.Append(merr, fmt.Errorf("failed to update realm %d: %w", id, err))
+		}
+	}
+
+	return merr.ErrorOrNil()
+}
+
+// rebuildModel rebuilds and updates the model for a single model.
+func (c *Controller) rebuildModel(ctx context.Context, id uint64) error {
```
mikehelmick, line 107: same, don't like that DB manipulation is bleeding out of the database package

pkg/controller/modeler/modeler.go
```diff
@@ -0,0 +1,210 @@
...
+		Order("date DESC").
+		Limit(21).
+		Pluck("codes_issued", &ys).
+		Error; err != nil {
+		return fmt.Errorf("failed to get stats: %w", err)
+	}
+
+	// Require some reasonable number of days of history before attempting to
+	// build a model.
+	if l := len(ys); l < 14 {
```
mikehelmick, line 125: What if there are gaps?

like we actually have 14 days of data, but some of those days are 0.

cmd/modeler/main.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/google/exposure-notifications-server/pkg/server"
+
+	"github.com/gorilla/mux"
+)
+
+func main() {
+	ctx, done := signalcontext.OnInterrupt()
+
+	debug, _ := strconv.ParseBool(os.Getenv("LOG_DEBUG"))
+	logger := logging.NewLogger(debug)
```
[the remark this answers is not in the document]
sethvargo: It's named inside the controller when we grab it.

pkg/controller/modeler/modeler.go
sethvargo: Moved to db

pkg/controller/modeler/modeler.go
sethvargo: Moved to db

pkg/controller/modeler/modeler.go
sethvargo: Added

cmd/server/assets/realm.html
sethvargo: Added

pkg/controller/modeler/modeler.go
sethvargo: As implemented, they are skipped. There wouldn't be a corresponding date, so there'd be no corresponding zero. 

The original query I wrote actually took that into account, using `generate_series` and a cross join. However, a significant spike-drop-spike _severely_ throws off the model (e.g. 100, 0, 80), so I'd rather just exclude zeros for now during modeling.

sethvargo:
@mikehelmick updated PTAnotherL

icco:
Merge conflicts.

sethvargo:
Rebased @icco

PR #553, opened 2020-09-17
Logic for batch user import

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/517

## Proposed Changes

* Logic for bulk-user-import
   * Users html5 filereader
         * That might be a no-go for very old browsers, so leaving the Beta flash up
* Displays CSV batches in a table and shows progress as it goes.

bulkimport

**Release Note**

```release-note
Logic for bulk user import
```


cmd/server/assets/users/import.html
```diff
@@ -22,19 +23,134 @@ <h1>Import users</h1>
...
+          <tbody id="csvtablebody"></tbody>
+        </table>
       </div>
     </div>
-    <a class="card-link" href="/users">&larr; All users</a>
+    <a class=" card-link" href="/users">&larr; All users</a>
   </main>
 
   {{template "scripts" .}}
+  <script type="text/javascript">
```
sethvargo, line 62: Hmmm this is a lot of javascript. Is there a strong reason to do this on the client vs the server? It seems like a traditional form submit + background processing would suffice here.

sethvargo: I'm also worried about the "modern" requirements here. When you think about most hospitals and medical facilities, they aren't running the latest and greatest browsers. I wouldn't be surprised if we have some clients using IE on XP still, with no support for the HTML5 file objects.

cmd/server/main.go
```diff
@@ -312,6 +312,7 @@ func realMain(ctx context.Context) error {
 		userSub.Handle("", userController.HandleCreate()).Methods("POST")
 		userSub.Handle("/new", userController.HandleCreate()).Methods("GET")
 		userSub.Handle("/import", userController.HandleImport()).Methods("GET")
+		userSub.Handle("/import/userbatch", userController.HandleImportBatch()).Methods("POST")
```
sethvargo, line 315: This is already under `/users` so having `user` in the route feels weird. Can we just use `/import` with a different verb instead?

```go
userSub.Handle("/import", usersController.HandleImport()).Methods("GET", "POST")
```

pkg/api/api.go
```diff
@@ -140,6 +140,27 @@ type CSRFResponse struct {
 	ErrorCode string `json:"errorCode"`
 }
 
+// UserBatchRequest is a request for bulk creation of users.
+// This is called by the Web frontend.
+// API is served at /users/import/userbatch
+type UserBatchRequest struct {
+	Users []BatchUser `json:"users"`
```
sethvargo, line 147:
```suggestion
	Users []*BatchUser `json:"users"`
```

Throughout

pkg/controller/user/importbatch.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		var request api.UserBatchRequest
+		if err := controller.BindJSON(w, r, &request); err != nil {
+			logger.Errorw("Error decoding request", "error", err)
+			c.h.RenderJSON(w, http.StatusBadRequest, api.Error(err))
+			return
+		}
+
+		newUsers := []api.BatchUser{}
```
sethvargo, line 44:
```suggestion
		newUsers := []*api.BatchUser{}
```

sethvargo: You could also pre-allocate this to the worst-case scenario (all users are new) to avoid the grow and copy.

pkg/controller/user/importbatch.go
```diff
@@ -0,0 +1,86 @@
...
+
+		var batchErr error
```
sethvargo, line 46:
```suggestion
		var batchErr *multierror.Error
```

sethvargo: Then you can call `.ErrorOrNil()` later

pkg/controller/user/importbatch.go
```diff
@@ -0,0 +1,86 @@
...
+			user.Realms = append(user.Realms, realm)
+
+			if err := c.db.SaveUser(user); err != nil {
+				logger.Errorw("Error saving user", "error", err)
+				batchErr = multierror.Append(batchErr, err)
+				continue
+			}
+		}
+
+		if batchErr != nil {
```
sethvargo, line 78: this should be batchError.ErrorOrNil()

pkg/controller/user/importbatch.go
```diff
@@ -0,0 +1,86 @@
...
+			controller.InternalError(w, r, c.h, batchErr)
```
sethvargo, line 79: Need a `return` after this

pkg/controller/user/importbatch.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		c.h.RenderJSON(w, http.StatusOK, api.UserBatchResponse{
```
sethvargo, line 82:
```suggestion
		c.h.RenderJSON(w, http.StatusOK, &api.UserBatchResponse{
```

cmd/server/assets/users/import.html
whaught: The main reason is so that it can batch and update progress as it goes. If we submit the whole file, we still need to return a response with all the users so we can do the password reset in javascript.

sethvargo:
@whaught looks like there's test failures

PR #555, opened 2020-09-17
Generate new user account on the server-side

whaught:
## Proposed Changes

* We currently generate new user accounts on the client, this moves it to the server.
    * Ideally we'd send an email from the server too, but we don't have and email service.
    * Someday I'd love to restrict the public firebase APIKey from creating accounts and only allow the server to do it (and rotate a private/secret key)

### DIFF
--- a/pkg/controller/user/create.go
+++ b/pkg/controller/user/create.go
@@ -21,7 +21,7 @@ import (
 	"firebase.google.com/go/auth"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
-	"github.com/google/exposure-notifications-verification-server/pkg/otp"
+	"github.com/sethvargo/go-password/password"
 )
 
 func (c *Controller) HandleCreate() http.Handler {
@@ -99,7 +99,7 @@ func (c *Controller) HandleCreate() http.Handler {
 		}
 
 		if !alreadyExists {
-			pwd, err := otp.GenerateAlphanumericCode(24)
+			pwd, err := password.Generate(24, 8, 8, false, true)
 			if err != nil {
 				flash.Alert("Failed to generate password for '%v'", form.Email)
 				c.renderNew(ctx, w, user, false)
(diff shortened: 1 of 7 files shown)


pkg/controller/user/create.go
```diff
@@ -96,17 +98,35 @@ func (c *Controller) HandleCreate() http.Handler {
 			return
 		}
 
+		if !alreadyExists {
+			pwd, err := otp.GenerateAlphanumericCode(24)
```
sethvargo, line 102: We should use [go-password](https://github.com/sethvargo/go-password) or [go-diceware](https://github.com/sethvargo/go-diceware) here instead since these are use passwords.

pkg/controller/user/create.go
```diff
@@ -96,17 +98,35 @@ func (c *Controller) HandleCreate() http.Handler {
...
+			if err != nil {
+				flash.Alert("Error generating password for '%v'", form.Email)
```
sethvargo, line 104:
```suggestion
				flash.Alert("Failed to generate password for '%v'", form.Email)
```

pkg/controller/user/create.go
```diff
@@ -96,17 +98,35 @@ func (c *Controller) HandleCreate() http.Handler {
...
+				c.renderNew(ctx, w, user, false)
+				return
+			}
+
+			fbUser := &auth.UserToCreate{}
+			fbUser.Email(user.Email).DisplayName(user.Name).Password(pwd)
+			_, err = c.client.CreateUser(ctx, fbUser)
```
sethvargo, line 111:
```go
if _, err := c.client.CreateUser(ctx, fbUser); err != nil {

}
```

avoids shadowing the error

pkg/controller/user/create.go
```diff
@@ -96,17 +98,35 @@ func (c *Controller) HandleCreate() http.Handler {
...
+			if err != nil {
+				flash.Alert("Error creating user '%v'", form.Email)
```
sethvargo, line 113: I think there's an edge case where where a user exists in Firebase but not in our database. In that case, alreadyExists would be false (thus executing this code), but this code would fail because the user exists in firebase.

This can occur when a user is deleted from their last realm. I believe we auto-delete the user record after that, but we don't delete the firebase user.

sethvargo (commented):
A few nits, but I think there's a legitimate edge case we need to handle whereby a user exists in firebase but not in our database.

whaught:
/hold

testing this still

sethvargo (approved):
/lgtm

Issue #545, opened 2020-09-15
Improve 2FA SMA enrollment


mikehelmick:
### TL;DR

We have user reports that 2FA enrollment is difficult due to the JS dialog box.

Move the confirmation to a separate page.

/kind enhancement

whaught:
the window.prompt for pin isn't ideal (it was the fastest thing to write for now)

whaught:
/assign

I'm stealing this. @sethvargo yell at me if you've already got progress here

sethvargo:
Steal away

PR #556, opened 2020-09-17
OTP entry form instead of window alert

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/545

## Proposed Changes

* Create a custom html pin form for login and registerphone
* Resend SMS link is disabled and enables on a 15s timer.
* Use the close-icon to return to previous
* Some css tweaks

pinform

register

**Release Note**

```release-note
Use custom html for entering SMS pin
```

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -36,9 +36,9 @@
   }
 
   a.disabled {
-  color: lightgray;
-  pointer-events: none;
-}
+    color: lightgray;
+    pointer-events: none;
+  }
 </style>
 {{end}}
 
@@ -234,7 +234,7 @@
 {{end}}
 
 {{define "flash"}}
-<div id="alerts">
+<div id="alerts-container">
   {{range $msg := .flash.Errors}}
   <div class="alert alert-danger" role="alert">
     {{$msg}}
@@ -366,7 +366,7 @@
     });
   });
 
-  var $alerts = $('#alerts');
+  var $alerts = $('#alerts-container');
 
   function clearExistingFlash() {
     $alerts.empty();
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -234,24 +234,26 @@
 {{end}}
 
 {{define "flash"}}
-{{range $msg := .flash.Errors}}
-<div class="alert alert-danger" role="alert">
-  {{$msg}}
+<div id="alerts">
+  {{range $msg := .flash.Errors}}
+  <div class="alert alert-danger" role="alert">
+    {{$msg}}
+  </div>
+  {{end}}
+  {{range $msg := .flash.Warnings}}
+  <div class="alert alert-warning" role="alert">
+    {{$msg}}
+  </div>
+  {{end}}
+  {{range $msg := .flash.Alerts}}
+  <div class="alert alert-success alert-dismissible fade show" role="alert">
+    {{$msg}}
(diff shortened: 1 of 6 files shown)


cmd/server/assets/header.html
```diff
@@ -34,6 +34,11 @@
     cursor: pointer;
     text-decoration: none;
   }
+
+  a.disabled {
+  color: lightgray;
```
sethvargo, line 39: Nit indent

cmd/server/assets/header.html
```diff
@@ -229,25 +234,27 @@ <h6 class="dropdown-header">Actions</h6>
...
-  {{$msg}}
-</div>
-{{end}}
-{{range $msg := .flash.Alerts}}
-<div class="alert alert-success alert-dismissible fade show" role="alert">
-  {{$msg}}
-  <button type="button" class="close" data-dismiss="alert" aria-label="Close">
-    <span aria-hidden="true">&times;</span>
-  </button>
+<div id="alerts">
```
sethvargo, line 237: I can see us accidentially conflicting with this in the future. Can we make this `alerts-container` or something more specific of an ID?

cmd/server/assets/login/_loginscripts.html
```diff
@@ -38,3 +38,28 @@
   }
 </script>
 {{end}}
+
+{{define "pindiv"}}
```
sethvargo, line 42: Maybe "login/pindiv" so it's namespaced?

cmd/server/assets/login/_loginscripts.html
```diff
@@ -38,3 +38,28 @@
...
+  <div class="card-header">
+    SMS Pin
+    <button type="button" class="close" aria-label="Close" id="pinClose">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+  <div class="card-body">
+    <form id="pinForm" class="floating-form" action="/" method="POST">
+      <div class="form-label-group">
+        <input type="text" id="pin" name="pin" class="form-control" placeholder="Pin" autocomplete="one-time-code" required autofocus />
```
sethvargo, line 53: We're using PIN and OTP interchangeably here - should standardize on one if possible.

sethvargo: And, if we go with PIN, we should always capitalize it as "PIN"

cmd/server/assets/login/login.html
```diff
@@ -15,7 +15,8 @@
     <div class="d-flex vh-100">
       <div class="d-flex w-100 justify-content-center align-self-center">
         <div class="col-sm-6">
-          <div class="card shadow-sm">
+
+          <div class="card shadow-sm" id="loginDiv">
```
sethvargo, line 19: CSS ids should be kebab-cased

cmd/server/assets/login/login.html
```diff
@@ -114,6 +120,49 @@
...
+          });
+      });
+
+      $pinClose.on('click', function(event) {
+        $submit.prop('disabled', false);
+        $loginDiv.show();
+        $pinDiv.hide();
+      });
+
+      $resendPin.on('click', function(event) {
```
sethvargo, line 148: Do you need to prevent default here?

whaught: it's a link with no href, not a submit button?

Issue #518, opened 2020-09-10
Add/remove realm admin post-user create


mikehelmick:
currently this is fixed, realm admins should be able to change this

/kind enhancement

PR #561, opened 2020-09-17
Allow system admins to remove themselves from a realm

sethvargo:
Fixes GH-518

**Release Note**

```release-note
Allow system admins to remove themselves from realms
```


whaught (commented):
/lgtm

Issue #573, opened 2020-09-18
password complexity enforcement


mikehelmick:
Add the ability to enforce password complexity

Must be a non-dictionary word
Must be composed of characters from at least three of the following four groups

- Upper case letters (A-Z)
- Lower case letters (a-z)
- Arabic numerals (0-9)
- Non-alphanumeric characters (punctuation symbols)

/kind enhancement
/assign @whaught

PR #579, opened 2020-09-18
Complexity requirements UI

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/573

## Proposed Changes

* Javscript-enforces complexity requirements for password change
   * Template provides complexity reqs from server config

passreq

**Release Note**

```release-note
Added UI for password complexity requirements
```

### DIFF
--- a/cmd/server/assets/login/select-password.html
+++ b/cmd/server/assets/login/select-password.html
@@ -36,7 +36,7 @@
                   <label for="retype">Retype password</label>
                 </div>
 
-                {{if .requirements}}
+                {{if .requirements.HasRequirements}}
                 <p class="card-text ml-4">
                   <small class="form-text text-muted">
                     <span class="row">Password should be:</span>
@@ -92,7 +92,7 @@
       let $password = $('#password');
       let $retype = $('#retype');
 
-      {{if .requirements}}
+      {{if .requirements.HasRequirements}}
       let $lenReq = $('#length-req');
       let $upperReq = $('#upper-req');
       let $lowerReq = $('#lower-req');
@@ -125,7 +125,7 @@
         return;
         }
 
-        {{if .requirements}}
+        {{if .requirements.HasRequirements}}
         let upper = 0
         let lower = 0
         let digit = 0
--- a/cmd/server/assets/login/select-password.html
+++ b/cmd/server/assets/login/select-password.html
@@ -36,6 +36,39 @@
                   <label for="retype">Retype password</label>
                 </div>
 
+                {{if .requirements}}
+                <p class="card-text ml-4">
+                  <small class="form-text text-muted">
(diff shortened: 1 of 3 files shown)


cmd/server/assets/login/select-password.html
```diff
@@ -80,9 +121,91 @@
         let email = $email.val();
         let pwd = $password.val();
         if (pwd != $retype.val()) {
-          flash("Password and retyped passwords must match.", "danger");
+           flash.error("Password and retyped passwords must match.");
+        return;
+        }
+
+        {{if .requirements}}
+        let upper = 0
```
sethvargo, line 129: Semicolons, throughout.

mikehelmick: silly languages that require statement terminators :)

cmd/server/assets/login/select-password.html
```diff
@@ -36,6 +36,39 @@
...
+                    {{end}}
+                    {{if gt .requirements.Special 0}}
+                    <span class="row ml-1" id="special-req">
+                      <span id="icon"></span>Contain {{.requirements.Special}} special character
+                    </span>
+                    {{end}}
+                  </small>
+                </p>
+                {{end}}
+
```
mikehelmick, line 71: double checking - the password never gets sent to our sever, so JS validation is the last stop?

whaught: that's correct. It's not my favorite thing.

cmd/server/assets/login/select-password.html
sethvargo: The worst part is that it's not "required". It's "interpreted differently depending on the runtime engine"

cmd/server/assets/login/select-password.html
sethvargo: I think the password would go to us if we did server-side user creation though, right @whaught? The SDK just isn't ready for Go yet?

mikehelmick: The existing FB admin sdk supports user creation and password setting

mikehelmick (approved):

cmd/server/assets/login/select-password.html
whaught: No it's that the Go SDK doesn't support Multi-Factor-Auth and associated management.

I know we can create a user with a password, idk about changing it with an ooB code - I'll check

whaught: The trouble is the oobCode that gets sent in the email - we have no way to verify that (outside of firebase) because we don't send the email.

If we had an SMTP server we could send the email ourselves with a token we track and verify server-side (then we could embed the validation logic server side too). We'd then need to handle oobCode storage and expiry ourselves too.

Issue #575, opened 2020-09-18
password reset interval


mikehelmick:
Allow realms to require that passwords are reset on a certain interval

This can be enforced on realm selection (rather than on login). 

For example if realm A has a 30d change policy, but the user only ever interacts with realm B without this enforcement, they will not be promoted to change their password.

/kind enhacment

sethvargo:
Feels like this could be bundled up with GH-573 /cc @whaught

sethvargo:
@whaught I think this is done?

PR #584, opened 2020-09-18
Record password reset time

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/575

## Proposed Changes

* Record the time of password change. This will be used to enforce password rotation

**Release Note**

```release-note
Record password reset time
```

### DIFF
--- a/pkg/controller/login/select_password.go
+++ b/pkg/controller/login/select_password.go
@@ -27,6 +27,7 @@ func (c *Controller) HandleSelectPassword() http.Handler {
 
 		m := controller.TemplateMapFromContext(ctx)
 		m["firebase"] = c.config.Firebase
+		m["requirements"] = &c.config.PasswordRequirements
 		c.h.RenderHTML(w, "login/select-password", m)
 	})
 }
--- a/pkg/controller/login/select_password.go
+++ b/pkg/controller/login/select_password.go
@@ -19,9 +19,10 @@ import (
 	"net/http"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/flash"
 )
 
-func (c *Controller) HandleSelectPassword() http.Handler {
+func (c *Controller) HandleSelectNewPassword() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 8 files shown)


pkg/controller/login/select_password.go
```diff
@@ -31,3 +33,36 @@ func (c *Controller) HandleSelectPassword() http.Handler {
 		c.h.RenderHTML(w, "login/select-password", m)
 	})
 }
+
+func (c *Controller) HandleSubmitNewPassword() http.Handler {
+	logger := c.logger.Named("HandleSubmitNewPassword")
```
sethvargo, line 38:
```suggestion
	logger := c.logger.Named("login.HandleSubmitNewPassword")
```

pkg/database/migrations.go
```diff
@@ -1081,6 +1081,16 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00046-AddPasswordLastChangedToUsers",
+			Migrate: func(tx *gorm.DB) error {
+				logger.Debugw("adding email verification required to realm")
+				return tx.Exec("ALTER TABLE users ADD COLUMN IF NOT EXISTS last_password_change DATE DEFAULT CURRENT_DATE").Error
```
sethvargo, line 1088: Does this need to be CURRENT_DATE()? I forget

pkg/database/user.go
```diff
@@ -227,6 +229,21 @@ func (u *User) CreateFirebaseUser(ctx context.Context, fbAuth *auth.Client) (boo
 	return false, nil
 }
 
+// PasswordChanged updates the last password change timestamp of the user.
+func (db *Database) PasswordChanged(email string, t time.Time) error {
+	q := db.db.
+		Model(&User{}).
+		Where("email = ?", email).
+		Update("last_password_change", t.UTC())
```
sethvargo, line 237: I think we should use UpdateColumn here instead (which will bypass validations and whatnot. Generally faster.

pkg/database/migrations.go
whaught: I think it's correct as-is https://www.postgresql.org/docs/8.2/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT

PR #588, opened 2020-09-18
Fix missing assets for redirect service

sethvargo:
/cc @icco 

**Release Note**

```release-note
Fix missing assets on redirect service
```

PR #589, opened 2020-09-18
Fix asset path for redirect

icco:
## Proposed Changes

* Changes asset copy to use variable path

**Release Note**

```release-note
NONE
```


icco:
/close

dedupe against #588

PR #591, opened 2020-09-18
Handle nil realm in logic

sethvargo:
/assign @whaught 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/middleware/emailverified.go
+++ b/pkg/controller/middleware/emailverified.go
@@ -83,8 +83,8 @@ func RequireVerified(ctx context.Context, client *auth.Client, db *database.Data
 }
 
 func NeedsEmailVerification(session *sessions.Session, realm *database.Realm, fbUser *auth.UserRecord) bool {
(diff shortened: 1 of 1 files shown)


pkg/controller/middleware/emailverified.go
```diff
@@ -83,8 +83,8 @@ func RequireVerified(ctx context.Context, client *auth.Client, db *database.Data
 }
 
 func NeedsEmailVerification(session *sessions.Session, realm *database.Realm, fbUser *auth.UserRecord) bool {
-	if (realm == nil || realm.EmailVerifiedMode == database.MFARequired) && !fbUser.EmailVerified {
-		return true
+	if realm == nil || realm.EmailVerifiedMode == database.MFARequired {
```
icco, line 86: shouldn't you check if fbUser is nil?

icco: discussed offline.

icco (approved):
/lgtm

PR #592, opened 2020-09-18
Realm setting for password rotation

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/575

## Proposed Changes

* Adds two new settings
    * Require password rotation at N days
    * Warn users about upcoming rotation at N days before expiry

withhelp

**Release Note**

```release-note
New realm setting for password rotation requirements
```

### DIFF
--- a/cmd/server/assets/realm.html
+++ b/cmd/server/assets/realm.html
@@ -173,7 +173,7 @@
             </div>
           </div>
 
-          <div class="form-group row">
+          <div class="form-group row mb-n3">
             <label for="passwordRotate" class="col-sm-3">Password rotation every:</label>
             <div class="col-sm-2">
               <select class="form-control" name="passwordRotate" id="passwordRotate">
@@ -187,7 +187,7 @@
               <label class="ml-3" for="passwordWarn">days</label>
             </div>
           </div>
-          <div class="row">
+          <div class="row mb-3">
             <div class="col-sm-3"></div>
             <small class="form-text text-muted col">
               The server will require the user to change their password after this number of
@@ -195,7 +195,7 @@
             </small>
           </div>
 
-          <div class="form-group row">
+          <div class="form-group row  mb-n3">
             <label for="passwordRotate" class="col-sm-3">Password rotation warning:</label>
             <div class="col-sm-2">
               <select class="custom-select{{if $realm.ErrorsFor "passwordWarn"}} is-invalid{{end}}" name="passwordWarn" id="passwordWarn">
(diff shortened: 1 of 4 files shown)


pkg/controller/realmadmin/save.go
```diff
@@ -44,13 +46,15 @@ func init() {
...
-		MFAMode           int16             `form:"MFAMode"`
-		EmailVerifiedMode int16             `form:"emailVerifiedMode"`
-		RequireDate       bool              `form:"requireDate"`
-		WelcomeMessage    string            `form:"welcomeMessage"`
+		Name               string            `form:"name"`
+		RegionCode         string            `form:"regionCode"`
+		AllowedTestTypes   database.TestType `form:"allowedTestTypes"`
+		MFAMode            int16             `form:"MFAMode"`
+		EmailVerifiedMode  int16             `form:"emailVerifiedMode"`
+		PasswordRotateDays uint              `form:"pwdRotate"`
```
sethvargo, line 54: I'd prefer if this was "password" instead of "pwd" because "pwd" can also mean "print working directory" in some contexts.

cmd/server/assets/realm.html
```diff
@@ -172,6 +172,32 @@ <h1>Realm settings</h1>
               </select>
             </div>
           </div>
+
+          <div class="form-group row">
+            <label for="pwdRotate" class="col-sm-3">Password rotation every:</label>
```
sethvargo, line 177: Probably need help text or something to explain this.

cmd/server/assets/realm.html
```diff
@@ -172,6 +172,32 @@ <h1>Realm settings</h1>
...
+                {{range $prd := .pwdRotateDays}}
+                <option value="{{$prd}}" {{if (eq $prd $current)}}selected{{end}}>{{if (eq $prd 0)}}Off{{else}}{{$prd}}{{end}}</option>
+                {{end}}
+              </select>
+            </div>
+            <div class="col-sm-1">days</div>
+          </div>
+
+          <div class="form-group row">
+            <label for="pwdWarn" class="col-sm-3">Password rotation warning:</label>
```
sethvargo, line 190: Is this days before expiration or days since creation? What's the anchor point and how is it calculated?

sethvargo (changes requested):

cmd/server/assets/realm.html
whaught: anchor point is days before expiration

sethvargo: It'd be good to have some help text for the user to explain this.

PR #598, opened 2020-09-19
Redirect to show page after user-create

whaught:
## Proposed Changes

* Redirect the user to the show-user page after creation
    * The current experience is potentially confusing as the user winds up back at the new-user page (despite the flash alert). This makes it more clear that the user is created.
* Adds a send-password reset button to the user page so admins can resend the reset if need-be.

**Release Note**

```release-note
Add password-reset button for admins
Change new-user redirect to show-user
```

### DIFF
--- a/pkg/controller/user/create.go
+++ b/pkg/controller/user/create.go
@@ -79,6 +79,7 @@ func (c *Controller) HandleCreate() http.Handler {
 			user.Name = form.Name
 		}
 
+		// Create firebase user first, if this fails we don't want a db.User entry
 		created, err := user.CreateFirebaseUser(ctx, c.client)
 		if err != nil {
 			flash.Alert("Failed to create user: %v", err)
--- a/pkg/controller/user/create.go
+++ b/pkg/controller/user/create.go
@@ -106,6 +106,7 @@ func (c *Controller) HandleCreate() http.Handler {
 		stats, err := c.getStats(ctx, user, realm)
 		if err != nil {
 			controller.InternalError(w, r, c.h, err)
+			return
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/controller/user/create.go
```diff
@@ -61,57 +61,59 @@ func (c *Controller) HandleCreate() http.Handler {
...
+		if created {
+			m := controller.TemplateMapFromContext(ctx)
+			m["created"] = true
+		}
+
+		stats, err := c.getStats(ctx, user, realm)
 		if err != nil {
-			flash.Alert("Failed to create user: %v", err)
-			c.renderNew(ctx, w, user, false)
+			controller.InternalError(w, r, c.h, err)
```
sethvargo, line 108: Need a `return` here

pkg/controller/user/show.go
```diff
@@ -54,25 +54,33 @@ func (c *Controller) HandleShow() http.Handler {
...
-		cacheKey := fmt.Sprintf("stats:user:%d:%d", realm.ID, user.ID)
-		if err := c.cacher.Fetch(ctx, cacheKey, &stats, 5*time.Minute, func() (interface{}, error) {
-			now := time.Now().UTC()
-			past := now.Add(-14 * 24 * time.Hour)
-			return user.Stats(c.db, realm.ID, past, now)
-		}); err != nil {
+		stats, err := c.getStats(ctx, user, realm)
+		if err != nil {
 			controller.InternalError(w, r, c.h, err)
-			return
```
sethvargo, line 66: This `return` is still needed

sethvargo (changes requested):

sethvargo (approved):
/gltm

PR #599, opened 2020-09-19
Create an account settings page

whaught:
## Proposed Changes

* Create a my-account page with the current user's information

TODO: this is where we will link in
1. Better password change (no email)
2. Change phone / unenroll 2nd factor
3. Change email and re-verify?

myaccount

**Release Note**

```release-note
Created a 'my account' page for the current user's settings
```


cmd/server/assets/login/account.html
```diff
@@ -0,0 +1,115 @@
...
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <h6 class="card-title">Name</h6>
+        <div class="card-text mb-3">
+          {{$user.Name}}
+        </div>
+
+        <h6 class="card-title">Email</h6>
+        <div class="card-text  mb-3">
```
sethvargo, line 31:
```suggestion
        <div class="card-text mb-3">
```

cmd/server/assets/login/account.html
```diff
@@ -0,0 +1,115 @@
...
+          {{$user.Email}}
+        </div>
+
+        <h6 class="card-title">System admin</h6>
```
sethvargo, line 35: I don't think we should disclose to average users that system admins exist. Perhaps wrap this entire thing in an `{{if $user.Admin}}`?

sethvargo: And then if you switch the CSS to mt instead of mb, you avoid the edge flowing over on the last element

cmd/server/assets/login/account.html
```diff
@@ -0,0 +1,115 @@
...
+        </li>
+      </ul>
+    </div>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card">
+        <div class="card-header">Member of realms</div>
+        <ul class="list-group list-group-flush">
+          {{range $realm := $user.Realms}}
+          <li class="list-group-item">{{$realm.Name}}</li>
```
sethvargo, line 65: Maybe add a little "admin" pill if they are an admin of that realm?

cmd/server/assets/login/account.html
```diff
@@ -0,0 +1,115 @@
...
+        </div>
+      </div>
+    </div>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Authentication</div>
+      <ul class="list-group list-group-flush">
+        <li class="list-group-item">
+          <div class="card-text" id="email-verified"></div>
+          <a href="/login/verify-email" class="card-link">Verify email</a>
```
sethvargo, line 47: Do we still want to show these if their email is verified?

pkg/controller/login/account.go
```diff
@@ -0,0 +1,32 @@
...
+// Package login defines the controller for the login page.
+package login
+
+import (
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+)
+
+func (c *Controller) HandleAccountSettings() http.Handler {
```
sethvargo, line 24: Thinking out loud... We could expand this to allow realm admins to view properties of another user by passing in an ID or qs param.

whaught: Well we already have the /users/{id} page. This one is intended to show some properties (like password and email status) that are more personal to your auth. I also think its annoying to find myself in the Users index list (which I'm sure will get larger) and it seems like a common task.
There's some overlap though.

One question: do we want "leave realm" and "delete account" options in here

pkg/database/user.go
```diff
@@ -40,6 +40,22 @@ type User struct {
 	LastPasswordChange time.Time
 }
 
+// PasswordAgeString displays the age of the password in friendly text.
+func (u *User) PasswordAgeString() string {
+	ago := time.Now().Sub(u.LastPasswordChange)
```
whaught, line 45:
```suggestion
	ago := time.Since(u.LastPasswordChange)
```

pkg/controller/login/account.go
sethvargo: I think we want them, but we want extra confirmation on them.

sethvargo (approved):
/lgtm

PR #601, opened 2020-09-19
Update UX on realm settings

sethvargo:
This updates the UX on the realm settings page to be tabular. There's currently a lot of settings and they aren't logically grouped - this changes that to group various settings within a tab.

**Release Note**

```release-note
Improve UX on realm settings page
```


sethvargo:
screenshot-20200919-144243@2x
screenshot-20200919-144244@2x
screenshot-20200919-144246@2x
screenshot-20200919-144249@2x
screenshot-20200919-144250@2x
screenshot-20200919-144252@2x

/assign @whaught

whaught (commented):
I love it!

/lgtm

whaught:
The one odd thing here is that the menu goes to settings#general

It causes the page to load the first tab and quickly switch to the second. The user sees a weird flicker every time they load the page. Maybe just make 'General' the first tab?

sethvargo:
I think we can set "active" on that first tab? We have to wait for the dom to load to pick the right tab

PR #603, opened 2020-09-19
Load realm for system admin pages

whaught:
## Proposed Changes

* Load the current realm for admin pages
    * Currently if you navigate to an admin page, the rest of the nav menu is lost which is a weird UX dead-end. This is cruft from when we only had RequireRealm (you can still get to this page if no particular realm is selected)
* Differentiated coloring and navigation for admin pages reinforces that this is a different experience.

newadminwhodis

**Release Note**

```release-note
Differentiated nav bar for System Admin
```

### DIFF
--- a/cmd/server/assets/admin/_nav.html
+++ b/cmd/server/assets/admin/_nav.html
@@ -6,11 +6,7 @@
 
   <nav class="navbar navbar-expand-md navbar-dark bg-primary">
     <div class="container">
-      <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbar" aria-controls="navbar"
-        aria-expanded="false" aria-label="Toggle navigation">
-        <span class="navbar-toggler-icon"></span>
-      </button>
-
+      {{template "navtoggle" .}}
       <div class="collapse navbar-collapse" id="navbar">
         <ul class="navbar-nav mr-auto">
           {{if .currentRealm}}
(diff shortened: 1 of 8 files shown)


sethvargo:
I actually removed this on purpose for the system admin pages. I want to avoid any hint that the user is in a realm context when doing system admin tasks.

sethvargo:
And I think the bar at the top is misleading the user into believe they are operating on realm settings vs system settings.

whaught:
I have changed it to differentiate the Admin pages per your concerns, but getting back to your realm now upfront. Thoughts?

sethvargo (commented):
I like the orange banner. I'm not sure I like the code duplication between the admin bar and the regular nav. I hear you on exiting vim, but remember that the average user won't be interacting with this part of the system.

whaught:
> I like the orange banner. I'm not sure I like the code duplication between the admin bar and the regular nav. I hear you on exiting vim, but remember that the average user won't be interacting with this part of the system.

Sorry - tried to remove most of the duplication (I think it'll be a bit more maintainable decomposed a bit).
Maybe this isn't the most important as it affects few... I just want nice things

cmd/server/assets/admin/_nav.html
```diff
@@ -1,4 +1,27 @@
 {{define "admin/navbar"}}
+<header class="mb-3">
+  <div href="/" class="d-block px-3 py-2 text-center text-bold text-white admin-header">
+    System Admin
```
sethvargo, line 4:
```suggestion
    System admin
```

sethvargo (approved):
/lgtm

Issue #346, opened 2020-08-24
Change signer and public key caches to use redis cache instead of local in-memory


mikehelmick:
/kind enhancement

sethvargo:
We'll need to make sure the cachable types only have public fields. I think all the public key implementations are public struct fields

PR #604, opened 2020-09-19
Use the same cacher for public key and signing caches

sethvargo:
Fixes GH-346

**Release Note**

```release-note
Use the configured cacher for public key and signing key caches
```

/assign @jeremyfaller

### DIFF
--- a/pkg/controller/jwks/jwks.go
+++ b/pkg/controller/jwks/jwks.go
@@ -52,8 +52,25 @@ type Controller struct {
 func (c *Controller) HandleIndex() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+
+		// key is the key in the cacher where the values for this JWK are cached.
+		key := r.URL.String()
+
+		// See if there's a cached value. Note we cannot use Fetch here because our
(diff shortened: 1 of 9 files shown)


pkg/controller/jwks/jwks.go
```diff
@@ -52,8 +52,25 @@ type Controller struct {
 func (c *Controller) HandleIndex() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+
```
sethvargo, line 55: @jeremyfaller can you take a look at this? There was a subtle bug that would result in a deadlock if the same cacher was given to the public key cache and the jwk builder. I think I fixed it, but would appreciate a second look.

sethvargo (commented):
/hold to test other signers

pkg/controller/jwks/jwks.go
jeremyfaller: Not sure I completely see the deadlock, would be great to have a test to prove it's fixed. In any case:

AFA the JWK stuff, this separation looks like it makes sense.

jeremyfaller (approved):

mikehelmick (approved):

PR #616, opened 2020-09-21
Terraform smoke test: fetching DB instance name from terraform state

chaodaiG:
Fetching database instance name from terraform output isn't always reliable, to achieve mroe reliable terraform smoke test, deleting all database instances prefix with en-verification during smoke test

### DIFF
--- a/scripts/terraform.sh
+++ b/scripts/terraform.sh
@@ -63,14 +63,15 @@ function deploy() {
   # google_app_engine_application.app is global, cannot be deleted once created,
   # if this project already has it created then terraform apply will fail,
   # importing it can solve this problem
-  terraform import module.en.google_app_engine_application.app ${PROJECT_ID} || true
-  terraform import module.en.google_firebase_project.default ${PROJECT_ID} || true
+  terraform import module.en.google_app_engine_application.app ${PROJECT_ID} || best_effort
+  terraform import module.en.google_firebase_project.default ${PROJECT_ID} || best_effort
   for networkEndpointGroups in adminapi apiserver server; do
     terraform import module.en.google_compute_region_network_endpoint_group.${networkEndpointGroups} \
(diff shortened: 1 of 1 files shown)


scripts/terraform.sh
```diff
@@ -92,25 +93,37 @@ function destroy() {
...
-  gcloud sql instances delete ${db_inst_name} -q --project=${PROJECT_ID} || true
+  local db_inst_names
+  # Fetching databases from previous terraform deployment output is not always reliable,
+  # especially when previous terraform deployment failed.
+  # If a user invoke `destroy` function here, it should be safe to delete all databases
+  # prefix with `en-verification`in this project.
+  db_inst_names="$(gcloud sql instances list --project=${PROJECT_ID} | grep -Eo 'en-verification-[a-zA-Z0-9]+' || true)"
+  if [[ -n "${db_inst_names}" ]]; then
+    for db_inst_name in ${db_inst_names}; do
+      gcloud sql instances delete ${db_inst_name} -q --project=${PROJECT_ID}
```
sethvargo, line 106: I'm... worried about this. If we (or someone else) accidentally ran this in prod, it'd be disastrous.

chaodaiG: The alternative approach is to grab the db instance name from terraform state, how does that sound to you?

sethvargo: Safer 😄

chaodaiG: Updated to fetch from terraform state, PTAL

chaodaiG: @sethvargo ^^

chaodaiG:
@sethvargo PTAL

sethvargo (approved):
/lgtm

chaodaiG:
May I also get an /approve please? @sethvargo

chaodaiG:
My bad
/unhold

PR #617, opened 2020-09-21
Tabbed navigation

whaught:
## Proposed Changes

* Switch to tabbed navigation
    * Gives user feedback on their current page
* This gives the UI a lighter feel, let me know what you think. It's somewhat subjective.

### Old
oldnav

### New
newnav

**Release Note**

```release-note
Switch navigation to light tabs
```


cmd/server/assets/header.html
```diff
@@ -187,23 +191,23 @@
...
       {{template "navtoggle" .}}
 
-      <div class="collapse navbar-collapse" id="navbar">
+      <div class="collapse navbar-collapse mt-2" id="navbar">
         {{if .currentRealm}}
-        <ul class="navbar-nav mr-auto">
+        <ul class="nav mr-auto">
           <li class="nav-item">
-            <a class="nav-link" href="/home">Issue code</a>
+            <a class="nav-link {{if .homeTab}}active{{end}}" href="/home">Issue code</a>
```
sethvargo, line 207: This feels subtle and could easily break if someone doesn't set the right template vars. I'd rather do this in js or using the `currentpage` thing I created earlier.

whaught: I made _a_ generic currentpage thing... but where can I find the one created earlier - I don't want to duplicate work.

I don't like using javascript because it tends to load after the page has been rendered. Plus it seems unnecessary

pkg/controller/middleware/currentpage.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: I think this already exists in the templatevars middleware?

sethvargo: @whaught https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/controller/middleware/template.go#L44

whaught: I was going to replace that. This gives you the option to get the path, page, or full url. Slightly nicer than the stringContains?

pkg/controller/middleware/currentpath.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/gorilla/mux"
+)
+
+type Path struct {
+	uri *url.URL
+}
+
+func CurrentPath() mux.MiddlewareFunc {
```
sethvargo, line 31: This should probably be `InjectCurrentPath` or similar

pkg/controller/middleware/currentpath.go
```diff
@@ -0,0 +1,59 @@
...
+
+			ctx = controller.WithTemplateMap(ctx, m)
+			*r = *r.WithContext(ctx)
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func (p *Path) Path() string {
```
sethvargo, line 47: I still think there's going to be cases where we need a strings.Contains on these. What do you think about dangling functions off path like `IsChild(s string)` so we can call those in templates?

whaught: Well the idea is we have the full path:
/foo/bar/baz -> foo/bar/baz
Dir()
/foo/bar/baz -> /foo/bar
and File()
/foo/bar/baz -> baz

which I think pretty well covers what you'd want to match with eq. You can still do stringContains on any of em

sethvargo: Right, I'm more talking about the DevX in the template. Instead of `stringsContains`, we could write these functions to be able to be used like:

```liquid
{{ .currentpath.IsDir("/foo") }}
```

versus

```liquid
{{ stringsContains .currentpath.Dir "/foo" }}
```

whaught: oh sure we can do that

sethvargo (approved):
/lgtm

@whaught can you remove the other injection in the template middleware in another PR?

PR #621, opened 2020-09-21
Require region codes are globally unique

sethvargo:
This is required for the ENX redirector, but we've also always made the
assumption that these need to be globally unique.

Note that this requires changing the column to citext (for
case-insensitive matching) and to NULL out the default of '', since ''
is a duplicate of ''.

It also handles any existing duplicates by appending -N to them before
applying the database-level unique constraint.

**Release Note**

```release-note
**Potentially breaking** Require region codes be globally unique, add database constraint for realm name uniqueness
```

/cc @icco @mikehelmick

### DIFF
--- a/pkg/controller/realmadmin/settings.go
+++ b/pkg/controller/realmadmin/settings.go
@@ -46,29 +46,33 @@ func init() {
 
 func (c *Controller) HandleSettings() http.Handler {
 	type FormData struct {
+		General        bool   `form:"general"`
 		Name           string `form:"name"`
 		RegionCode     string `form:"region_code"`
 		WelcomeMessage string `form:"welcome_message"`
 
+		Codes                 bool              `form:"codes"`
 		AllowedTestTypes      database.TestType `form:"allowed_test_types"`
-		RequireDate           *bool             `form:"require_date"`
+		RequireDate           bool              `form:"require_date"`
 		CodeLength            uint              `form:"code_length"`
 		CodeDurationMinutes   int64             `form:"code_duration"`
 		LongCodeLength        uint              `form:"long_code_length"`
 		LongCodeDurationHours int64             `form:"long_code_duration"`
 		SMSTextTemplate       string            `form:"sms_text_template"`
 
-		TwilioSMSConfig  bool   `form:"twilio_sms_config"`
+		SMS              bool   `form:"sms"`
 		TwilioAccountSid string `form:"twilio_account_sid"`
 		TwilioAuthToken  string `form:"twilio_auth_token"`
 		TwilioFromNumber string `form:"twilio_from_number"`
 
-		MFAMode                     *int16 `form:"mfa_mode"`
-		EmailVerifiedMode           *int16 `form:"email_verified_mode"`
-		PasswordRotationPeriodDays  uint   `form:"password_rotation_period_days"`
-		PasswordRotationWarningDays uint   `form:"password_rotation_warning_days"`
+		Security                    bool  `form:"security"`
+		MFAMode                     int16 `form:"mfa_mode"`
+		EmailVerifiedMode           int16 `form:"email_verified_mode"`
+		PasswordRotationPeriodDays  uint  `form:"password_rotation_period_days"`
+		PasswordRotationWarningDays uint  `form:"password_rotation_warning_days"`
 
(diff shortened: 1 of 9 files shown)


pkg/database/migrations.go
```diff
@@ -1131,6 +1131,84 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+					return err
+				}
+
+				// Make all region codes uppercase.
+				if err := tx.Exec("UPDATE realms SET region_code = UPPER(region_code) WHERE region_code IS NOT NULL").Error; err != nil {
+					return err
+				}
+
+				// Find any existing duplicate region codes - this could be combined
+				// into a much large SQL statement with the next thing, but I'm
```
jeremyfaller, line 1149: s/large/larger/

pkg/controller/realmadmin/settings.go
```diff
@@ -109,13 +110,13 @@ func (c *Controller) HandleSettings() http.Handler {
 		}
 
 		// General
-		if v := form.Name; v != "" {
+		if v := strings.TrimSpace(form.Name); v != "" {
```
mikehelmick, line 113: beforeSave already does this 

https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/database/realm.go#L197

pkg/database/migrations.go
```diff
@@ -1131,6 +1131,84 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+				}
+
+				// Update any duplicate regions to not be duplicate anymore.
+				for _, dupRegionCode := range dupRegionCodes {
+					logger.Warn("de-duplicating region code %q", dupRegionCode)
+
+					// I call this the "Microsoft method". For each duplicate realm,
+					// append -N, starting with 1. If there are 3 realms with the region
+					// code "PA", their new values will be "PA", "PA-1", and "PA-2"
+					// respectively.
```
mikehelmick, line 1170: I don't think there is a 'good' way do this this automatically.

I also don't have another suggestion.

pkg/database/migrations.go
```diff
@@ -1131,6 +1131,84 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+					Where("region_code IS NOT NULL").
+					Group("region_code").
+					Having("COUNT(*) > 1").
+					Pluck("region_code", &dupRegionCodes).
+					Error; err != nil {
+					return err
+				}
+
+				// Update any duplicate regions to not be duplicate anymore.
+				for _, dupRegionCode := range dupRegionCodes {
```
jeremyfaller, line 1164: I hate to be "that guy" but do we want to write a test on this?

jeremyfaller (approved):

pkg/database/migrations.go
sethvargo: We don't have a way to test migrations right now besides "manually".

pkg/database/realm.go
```diff
@@ -80,20 +80,26 @@ type Realm struct {
 	// Name is the name of the realm.
 	Name string `gorm:"type:varchar(200);unique_index"`
 
+	// RegionCode is both a display attribute and required field for ENX. To
+	// handle NULL and uniqueness, the field is converted from it's ptr type to a
+	// concrete type in callbacks. Do not modify RegionCodePtr directly.
+	RegionCode    string  `gorm:"-"`
+	RegionCodePtr *string `gorm:"column:region_code; type:varchar(10);"`
```
mikehelmick, line 87: does gorm work w/ sql.NullString?

sethvargo: Yea, but our code doesn't (really). Changing the type of `RegionCode` requires every caller to update their type. So all cases of:

```go
realm.RegionCode = "foo"
printf(realm.RegionCode)
```

become

```go
if v != "" {
  realm.RegionCode = sql.NullString{Value:v, Valid: true}
} else {
  ream.RegionCode = sql.NullString{}
}

if realm.RegionCode.Valid() {
  printf(realm.RegionCode.Value)
} else {
  printf("")
}
```

and I don't want to put that burden on callers tbh.

sethvargo:
@mikehelmick PTAnotherL

mikehelmick (approved):

PR #623, opened 2020-09-21
Hide ENX settings once enabled behind advanced panel

sethvargo:
**Release Note**

```release-note
Make it more difficult to accidentally disable ENX
```

/cc @mikehelmick 
/assign @whaught


sethvargo:
screenshot-20200921-183115@2x

sethvargo:
screenshot-20200921-183118@2x

whaught (approved):
/lgtm

whaught:
I did find in my browser that if I got to #something and then click in the <nav> dropdown setting, my browser set it to #general but the page didn't actually go there. Weird stuff.

sethvargo:
Yea, I noticed that too and I think it's actually a bug in bootstrap 😦. I couldn't find a good solution.

PR #624, opened 2020-09-21
Suppress user-not-found error

whaught:
## Proposed Changes

* Respond the same for users who's emails are not found
* This won't fool anyone savvy - we need to be either:
1) sending these from our own SMTP server
2) find a firebase toggle that disables this response
2) run all firebase from node.js in a function as a proxy, filter server side, keep the firebase api key secret

### DIFF
--- a/cmd/server/assets/login/reset-password.html
+++ b/cmd/server/assets/login/reset-password.html
@@ -70,8 +70,12 @@
           flash.alert('Password reset email sent.');
         }).catch(function(error) {
           flash.clear();
-          flash.error(error.message);
-          $submit.prop('disabled', false);
+          if (error.code = "auth/user-not-found") {
+            flash.alert('Password reset email sent.');
+          } else {
+            flash.error(error.message);
+            $submit.prop('disabled', false);
+          }
         });
       });
     });


cmd/server/assets/login/reset-password.html
```diff
@@ -70,8 +70,12 @@
           flash.alert('Password reset email sent.');
         }).catch(function(error) {
           flash.clear();
-          flash.error(error.message);
-          $submit.prop('disabled', false);
+          if (error.code = "auth/user-not-found") {
```
sethvargo, line 73: So... this fixes the issue, but it doesn't actually stop anyone with a Web Inspector. You can still see the response sent.

sethvargo (approved):
/lgtm
/approve

cmd/server/assets/login/reset-password.html
mikehelmick: we should always say something like - "check your email for your reset password link" and leave it at that.

sethvargo: I mean the response from calling the firebase API - until this is all server side, someone can just look at what firebase is returning. Or they could just talk to firebase directly with our API key.

whaught: This is what I was talking about in the description. We need an SMTP server to send mail if we want to do it ourselves or we could wrap it all in a node.js cloud function (gross).

sethvargo: I mean. The other option is to bundle the node.js runtime into our container and exec...

Issue #608, opened 2020-09-21
Manage MFA factors


whaught:
The user should be able to add more than one auth-factor.

1. Select a factor by display name at login
2. Show enrolled factors in My Settings page
3. Allow un-enrollment of a factor

### Notes
The user should not be able to remove their MFA without first adding a new one or re-confirming the existing one. Or having very-recently logged-in. Firebase enforces this by requiring recent login and may give errors for older sessions.
Nice-to-have: We may want to catch the error and redirect to login, but then have login return to the setting instead of /home.

PR #628, opened 2020-09-22
Allow deletion of factors

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/608

## Proposed Changes

* If there are enrolled factors, show them in a list
* Delete link for removing an enrolled factor
* Minor fix to make pin-confirm div hide on success

oncemore

**Release Note**

```release-note
Allow users to delete enrolled factor for MFA
```

### DIFF
--- a/cmd/server/assets/login/register-phone.html
+++ b/cmd/server/assets/login/register-phone.html
@@ -19,6 +19,13 @@
     <div class="d-flex vh-100">
       <div class="d-flex w-100 justify-content-center">
         <div class="col-sm-6">
+          <div class="card shadow-sm mb-3" id="registered-div" style="display:none;">
+            <div class="card-header">Registered factors</div>
+            <div id="factors" class="card-body">
+              <ul class="list-group list-group-flush">
+              </ul>
+            </div>
+          </div>
 
           <div class="card shadow-sm" id="register-div">
             <div class="card-header">Multi-factor authentication</div>
@@ -34,7 +41,7 @@
                 <strong>{{$currentRealm.Name}}</strong>
                 {{if eq .currentRealm.MFAMode.String "required"}}requires{{else}}recommends{{end}}
                 enhanced security via SMS-based 2-factor authentication. Please
-                provide your information below.
+                provide your information below to enroll.
               </p>
 
               <form id="register-form" class="floating-form" action="/" method="POST">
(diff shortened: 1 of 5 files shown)


cmd/server/assets/login/register-phone.html
```diff
@@ -19,6 +19,13 @@
     <div class="d-flex vh-100">
       <div class="d-flex w-100 justify-content-center">
         <div class="col-sm-6">
+          <div class="card shadow-sm mb-3" id="registered-div" style="display:none;">
```
sethvargo, line 22:
```suggestion
          <div class="card shadow-sm mb-3 d-none" id="registered-div">
```

cmd/server/assets/login/register-phone.html
```diff
@@ -100,8 +107,62 @@
...
+            row.append(unenroll)
+
+            let time = $('<small class="row text-muted ml-1"></small>').text('Enrolled at: ' + factor.enrollmentTime);
+            row.append(time);
+
+            unenroll.on('click', function(event) {
+              unenrollFactor(factor);
+            });
+
+            $factors.append(li);
```
sethvargo, line 144: This is going to append the `<li>` to the parent `<div>` with an id of #factors. Did you mean to put the ID on the `<ul>`?

cmd/server/assets/login/register-phone.html
```diff
@@ -100,8 +107,62 @@
...
+          return
+        }
+
+        if (user.multiFactor.enrolledFactors.length > 0) {
+          $skip.text("Account settings");
+          $skip.attr("href","/account");
+
+          for (i = 0; i < user.multiFactor.enrolledFactors.length; i++) {
+            let factor = user.multiFactor.enrolledFactors[i];
+            let li = $('<li class="list-group-item"></li>');
```
sethvargo, line 126: It's usually convention to name jQuery variables prefixed with a `$`, so this would be `$li`.

sethvargo: It's also _much_ faster to build a blank DOM node and then call `addClass`. As written, jquery has to parse the entire string and build the class list.

cmd/server/assets/login/register-phone.html
```diff
@@ -100,8 +107,62 @@
...
+            li.attr("id", "factor"+factor.uid)
+
+            let row = $('<div></div>').text(factor.displayName);
+            li.append(row)
+
+            let unenroll = $('<a href="#" class="text-danger float-right" title="Delete this factor"></a>');
```
sethvargo, line 132: Same note about building classes, attrs, and titles later, and `$unenroll`.

cmd/server/assets/login/register-phone.html
```diff
@@ -100,8 +107,62 @@
...
+        if (user.multiFactor.enrolledFactors.length > 0) {
+          $skip.text("Account settings");
+          $skip.attr("href","/account");
+
+          for (i = 0; i < user.multiFactor.enrolledFactors.length; i++) {
+            let factor = user.multiFactor.enrolledFactors[i];
+            let li = $('<li class="list-group-item"></li>');
+            li.attr("id", "factor"+factor.uid)
+
+            let row = $('<div></div>').text(factor.displayName);
```
sethvargo, line 129:
```suggestion
            let row = $('<div />').text(factor.displayName);
```

sethvargo: Does `.text()` return the object? I think you may need to do:

```javascript
let $row = $('<div/>');
$row.text(factor.displayName);
```

cmd/server/assets/login/register-phone.html
```diff
@@ -145,7 +206,9 @@
         user.multiFactor.enroll(multiFactorAssertion, $displayName.val()).then(function() {
           flash.clear();
           flash.alert('SMS authentication enrolled successfully.');
-          $skip.text("Continue")
+          $skip.text("Continue");
```
sethvargo, line 209:
```suggestion
          $skip.text('Continue');
```

cmd/server/assets/login/register-phone.html
```diff
@@ -100,8 +107,62 @@
...
+          $registeredDiv.show();
+        }
+      });
+
+      function unenrollFactor(factor) {
+        firebase.auth().currentUser.multiFactor.unenroll(factor)
+          .then(function() {
+            $('#factor'+factor.uid).remove();
+            flash.clear();
+            flash.alert(factor.displayName + ' unenrolled successfully.');
```
sethvargo, line 155:
```suggestion
            flash.alert(`Successfully unenrolled ${factor.displayName}`);
```

sethvargo (changes requested):
Should there be any kind of check/warning if any of the user's realms require MFA and they are about to remove their last MFA devic?

cmd/server/assets/login/register-phone.html
```diff
@@ -145,7 +216,9 @@
         user.multiFactor.enroll(multiFactorAssertion, $displayName.val()).then(function() {
           flash.clear();
           flash.alert('SMS authentication enrolled successfully.');
-          $skip.text("Continue")
+          $skip.text('Continue');
+          $registerDiv.show();
```
sethvargo, line 220: I don't think this will work with d-none, but maybe jquery is smarter than me. I think you need to call `.removeClass`

sethvargo (approved):
/lgtm

PR #630, opened 2020-09-22
Minor fixes to MFA page

whaught:
## Proposed Changes

* Clear the text from the register fields after success (so you can register again)
* Redirect to /signout if auth becomes invalid
* Show/hide factors as they are added


mikehelmick (commented):
/lgtm
/approve

PR #634, opened 2020-09-22
Add redirect support for iOS and Android.

### LINKED ISSUE #580: redirect service support associated domain
The redirect service needs to support hosting the Apple associated domains data.
https://developer.apple.com/documentation/safariservices/supporting_associated_domains

And Android App Links ownership
https://developer.android.com/training/app-links/verify-site-associations

/kind enhancement

jeremyfaller:
Reviewers:

Please take a close look at caching and the terraform config work here. For caching, I went with purely the DB cache -- not going any deeper. If we need more, please LMK.

Fixes #580

## Proposed Changes

* Creates a pair of endpoints for iOS and Android to get app redirect information
* Adds the a few new fields to `authorized_apps` to hold data needed for the redirection service.

**Release Note**

```release-note
Redirection service now publishes the iOS and Android redirection metadata.
APIKeys now includes support for specifying the needed iOS and Android metadata for redirection.
```

### DIFF
--- a/cmd/enx-redirect/main.go
+++ b/cmd/enx-redirect/main.go
@@ -16,14 +16,17 @@ package main
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"net/http"
 	"os"
 	"strconv"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/buildinfo"
+	"github.com/google/exposure-notifications-verification-server/pkg/cache"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/associated"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/middleware"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/redirect"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
@@ -58,14 +61,14 @@ func main() {
 func realMain(ctx context.Context) error {
 	logger := logging.FromContext(ctx)
 
-	config, err := config.NewRedirectConfig(ctx)
+	cfg, err := config.NewRedirectConfig(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to process config: %w", err)
 	}
 
 	// Setup monitoring
 	logger.Info("configuring observability exporter")
-	oeConfig := config.ObservabilityExporterConfig()
+	oeConfig := cfg.ObservabilityExporterConfig()
 	oe, err := observability.NewFromEnv(ctx, oeConfig)
 	if err != nil {
 		return fmt.Errorf("unable to create ObservabilityExporter provider: %w", err)
@@ -76,25 +79,69 @@ func realMain(ctx context.Context) error {
 	defer oe.Close()
 	logger.Infow("observability exporter", "config", oeConfig)
 
+	// Setup cacher
+	cacher, err := cache.CacherFor(ctx, &cfg.Cache, cache.MultiKeyFunc(
+		cache.HMACKeyFunc(sha1.New, cfg.Cache.HMACKey),
+		cache.PrefixKeyFunc("cache:"),
+	))
+	if err != nil {
+		return fmt.Errorf("failed to create cacher: %w", err)
+	}
+	defer cacher.Close()
+
+	// Setup database
+	db, err := cfg.Database.Load(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to load database config: %w", err)
+	}
+	if err := db.OpenWithCacher(ctx, cacher); err != nil {
+		return fmt.Errorf("failed to connect to database: %w", err)
+	}
+	defer db.Close()
+
 	// Create the router
 	r := mux.NewRouter()
 
 	// Create the renderer
-	h, err := render.New(ctx, config.AssetsPath, config.DevMode)
+	h, err := render.New(ctx, cfg.AssetsPath, cfg.DevMode)
 	if err != nil {
 		return fmt.Errorf("failed to create renderer: %w", err)
 	}
 
 	// Install common security headers
-	r.Use(middleware.SecureHeaders(ctx, config.DevMode, "html"))
+	r.Use(middleware.SecureHeaders(ctx, cfg.DevMode, "html"))
 
 	// Enable debug headers
 	processDebug := middleware.ProcessDebug(ctx)
 	r.Use(processDebug)
 
+	// iOS and Android include functionality to associate data between web-apps
+	// and device apps. Things like handoff between websites and apps, or
+	// shared credentials are the common usecases. The redirect server
+	// publishes the metadata needed to share data between the two domains to
+	// offer a more seemless experience between the website and apps. iOS and
+	// Android publish specs as to what this format looks like, and both live
+	// under the /.well-known directory on the server.
+	//
+	//   Android Specs:
+	//     https://developer.android.com/training/app-links/verify-site-associations
+	//   iOS Specs:
+	//     https://developer.apple.com/documentation/safariservices/supporting_associated_domains
+	{ // .well-known directory
+		wk := r.PathPrefix("/.well-known").Subrouter()
+
+		// Enable the iOS and Android redirect handler.
+		assocHandler, err := associated.New(ctx, db, cacher, h)
+		if err != nil {
+			return fmt.Errorf("failed to create associated links handler %w", err)
+		}
+		wk.PathPrefix("/apple-app-site-association").Handler(assocHandler.HandleIos()).Methods("GET")
+		wk.PathPrefix("/assetlinks.json").Handler(assocHandler.HandleAndroid()).Methods("GET")
+	}
+
 	r.Handle("/health", controller.HandleHealthz(ctx, nil, h)).Methods("GET")
 
-	redirectController, err := redirect.New(ctx, config, h)
+	redirectController, err := redirect.New(ctx, cfg, h)
 	if err != nil {
 		return err
 	}
@@ -103,10 +150,10 @@ func realMain(ctx context.Context) error {
 	mux := http.NewServeMux()
 	mux.Handle("/", r)
 
-	srv, err := server.New(config.Port)
+	srv, err := server.New(cfg.Port)
 	if err != nil {
 		return fmt.Errorf("failed to create server: %w", err)
 	}
-	logger.Infow("server listening", "port", config.Port)
+	logger.Infow("server listening", "port", cfg.Port)
 	return srv.ServeHTTPHandler(ctx, mux)
 }
--- a/cmd/server/assets/mobileapp/edit.html
+++ b/cmd/server/assets/mobileapp/edit.html
@@ -1,0 +1,57 @@
+{{define "mobileapp/edit"}}
+
+{{$app := .app}}
+
+<!doctype html>
+<html lang="en">
+<head>
+  {{template "head" .}}
+  {{template "floatingform" .}}
+</head>
+
+<body class="tab-content" onload="osSelect()">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Edit Mobile App</h1>
+    <p>
+      Use the form below to edit the app.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <form method="POST" action="/mobileapp/{{$app.ID}}">
+          <input type="hidden" name="_method" value="PATCH">
+          {{ .csrfField }}
+
+          <div class="form-label-group">
+            <input type="text" id="name" name="name" class="form-control{{if $app.ErrorsFor "name"}} is-invalid{{end}}" value="{{$app.Name}}" placeholder="Application name" autofocus>
+            <label for="name">Application name</label>
+            {{if $app.ErrorsFor "name"}}
+            <div class="invalid-feedback">
+              {{joinStrings ($app.ErrorsFor "name") ", "}}
+            </div>
+            {{end}}
+          </div>
+
+		  {{template "app" .}}
+
+          <button type="submit" class="btn btn-primary btn-block">Update Mobile App</button>
+        </form>
+      </div>
+    </div>
+
+    <div>
+      <p>
+        <a href="/mobileapp">&larr; Back to all mobile apps</a>
+      </p>
+    </div>
+  </main>
+
+  {{template "scripts" .}}
+</body>
+</html>
+{{end}}
--- a/cmd/server/assets/mobileapp/index.html
+++ b/cmd/server/assets/mobileapp/index.html
@@ -1,0 +1,84 @@
+{{define "mobileapp/index"}}
+
+{{$apps := .apps}}
+
+<!doctype html>
+<html lang="en">
+<head>
+  {{template "head" .}}
+</head>
+
+<body class="tab-content">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Mobile Apps</h1>
+    <p>
+      These are the mobile apps for this realm. You
+      can also <a href="/mobileapp/new">create a new mobile app</a>.
+    </p>
+
+    {{if .apps}}
+    <div class="table-responsive">
+      <table class="table table-bordered table-striped bg-white">
+        <thead>
+          <tr>
+            <th scope="col">App</th>
+            <th scope="col" width="95">Enabled</th>
+            <th scope="col" width="40"></th>
+          </tr>
+        </thead>
+        <tbody>
+        {{range .apps}}
+          <tr>
+
+            <td>
+              <a href="/mobileapp/{{.ID}}" class="text-truncate">{{.Name}}</a>
+            </td>
+
+            <td>
+              {{if .DeletedAt}}
+                <span class="badge badge-pill badge-danger">Deleted</span>
+              {{else}}
+                <span class="badge badge-pill badge-success">Enabled</span>
+              {{end}}
+            </td>
+
+            <td class="text-center">
+              {{if .DeletedAt}}
+              <a href="/mobileapp/{{.ID}}/enable" class="d-block text-danger"
+                data-method="patch"
+                data-confirm="Are you sure you want to restore '{{.Name}}'?"
+                data-toggle="tooltip"
+                title="Restore this mobile app">
+                <span class="oi oi-loop-circular" aria-hidden="true"></span>
+              </a>
+              {{else}}
+              <a href="/mobileapp/{{.ID}}/disable" class="d-block text-danger"
+                data-method="patch"
+                data-confirm="Are you sure you want to disable '{{.Name}}'?"
+                data-toggle="tooltip"
+                title="Disable this mobile app">
+                <span class="oi oi-trash" aria-hidden="true"></span>
+              </a>
+              {{end}}
+            </td>
+
+          </tr>
+        {{end}}
+        </tbody>
+      </table>
+    </div>
+    {{else}}
+    <p class="text-center">
+      <em>There are no mobile apps.</em>
+    </p>
+    {{end}}
+  </main>
+
+  {{template "scripts" .}}
+</body>
+</html>
+{{end}}
--- a/cmd/server/assets/mobileapp/new.html
+++ b/cmd/server/assets/mobileapp/new.html
@@ -1,0 +1,50 @@
+{{define "mobileapp/new"}}
+
+{{$app := .app}}
+
+<!doctype html>
+<html lang="en">
+<head>
+  {{template "head" .}}
+  {{template "floatingform" .}}
+</head>
+
+<body class="tab-content" onload="osSelect()">>
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>New Mobile App</h1>
+    <p>
+      Use the form below to create a new mobile app.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <form method="POST" action="/mobileapp">
+          {{ .csrfField }}
+
+          <div class="form-label-group">
+            <input type="text" id="name" name="name" class="form-control{{if $app.ErrorsFor "name"}} is-invalid{{end}}" value="{{$app.Name}}" placeholder="Application name" autofocus>
+            <label for="name">Application name</label>
+            {{if $app.ErrorsFor "name"}}
+            <div class="invalid-feedback">
+              {{joinStrings ($app.ErrorsFor "name") ", "}}
+            </div>
+            {{end}}
+          </div>
+
+		  {{template "app" .}}
+
+          <button type="submit" class="btn btn-primary btn-block">Create mobile app</button>
+        </form>
+      </div>
+    </div>
+  </main>
+
+  {{template "scripts" .}}
+</body>
+</html>
+{{end}}
--- a/cmd/server/assets/mobileapp/show.html
+++ b/cmd/server/assets/mobileapp/show.html
@@ -1,0 +1,76 @@
+
+{{define "mobileapp/show"}}
+
+{{$app := .app}}
+
+<!doctype html>
+<html lang="en">
+
+<head>
+  {{template "head" .}}
+</head>
+
+<body class="tab-content">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>{{$app.Name}}</h1>
+    <p class="float-right">
+      <a href="/mobileapp/{{$app.ID}}/edit">Edit</a>
+    </p>
+    <p>
+      Here is information about the Mobile App.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <strong>App name</strong>
+        <div class="mb-3">
+          {{$app.Name}}
+        </div>
+        <strong>OS</strong>
+        <div>
+          {{if (eq $app.OS .iOS)}}
+          iOS
+          {{else if (eq $app.OS .android)}}
+          Android
+          {{end}}
+        </div>
+
+		{{if (eq $app.OS .iOS)}}
+          <strong>iOS App ID:</strong>
+          <div>
+          {{$app.AppID}}
+          </div>
+        {{end}}
+
+		{{if (eq $app.OS .android)}}
+          <strong>Android Bundle Identifier:</strong>
+          <div>
+          {{$app.AppID}}
+          </div>
+
+          <strong>SHA:</strong>
+          <div>
+          {{$app.SHA}}
+          </div>
+        {{end}}
+      </div>
+	</div>
+
+    <div>
+      <p>
+        <a href="/mobileapp">&larr; Back to all mobile apps</a>
+      </p>
+    </div>
+  </main>
+
+  {{template "scripts" .}}
+
+</body>
+
+</html>
+{{end}}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -36,6 +36,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/jwks"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/login"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/middleware"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/mobileapp"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/realmadmin"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/realmkeys"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/user"
@@ -299,6 +300,29 @@ func realMain(ctx context.Context) error {
 		sub.Handle("/status", codeStatusController.HandleIndex()).Methods("GET")
 		sub.Handle("/show", codeStatusController.HandleShow()).Methods("POST")
 		sub.Handle("/{uuid}/expire", codeStatusController.HandleExpirePage()).Methods("PATCH")
+	}
+
+	// mobileapp
+	{
+		sub := r.PathPrefix("/mobileapp").Subrouter()
+		sub.Use(requireAuth)
+		sub.Use(loadCurrentRealm)
+		sub.Use(requireRealm)
+		sub.Use(processFirewall)
+		sub.Use(requireAdmin)
+		sub.Use(requireVerified)
+		sub.Use(requireMFA)
+		sub.Use(rateLimit)
+
+		mobileController := mobileapp.New(ctx, config, cacher, db, h)
+		sub.Handle("", mobileController.HandleIndex()).Methods("GET")
+		sub.Handle("", mobileController.HandleCreate()).Methods("POST")
+		sub.Handle("/new", mobileController.HandleCreate()).Methods("GET")
+		sub.Handle("/{id}/edit", mobileController.HandleUpdate()).Methods("GET")
+		sub.Handle("/{id}", mobileController.HandleShow()).Methods("GET")
+		sub.Handle("/{id}", mobileController.HandleUpdate()).Methods("PATCH")
+		sub.Handle("/{id}/disable", mobileController.HandleDisable()).Methods("PATCH")
+		sub.Handle("/{id}/enable", mobileController.HandleEnable()).Methods("PATCH")
 	}
 
 	// apikeys
--- a/pkg/controller/associated/android.go
+++ b/pkg/controller/associated/android.go
@@ -1,0 +1,54 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package associated
+
+import (
+	"strings"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+)
+
+type AndroidData struct {
+	Relation []string `json:"relation,omitempty"`
+	Target   Target   `json:"target,omitempty"`
+}
+
+type Target struct {
+	Namespace    string   `json:"namespace,omitempty"`
+	PackageName  string   `json:"package_name,omitempty"`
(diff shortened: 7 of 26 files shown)


sethvargo (changes requested):
I'm concerned about tying the API key creation to a particular app/os architecture. This might be better as its own table. We should also be imploring more aggressive caching of these JSON responses, since every device will request them automatically like a favicon.

cmd/server/assets/apikeys/new.html
```diff
@@ -49,13 +49,67 @@ <h1>New API key</h1>
             {{end}}
           </div>
 
+          <div class="form-group">
```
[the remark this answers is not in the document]
jeremyfaller: As written, yes. I can update to handle both.

pkg/database/migrations.go
```diff
@@ -32,6 +32,22 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 	logger := logging.FromContext(ctx)
 	options := gormigrate.DefaultOptions
 
+	// Helper function to safely add columns to the database.
+	addColumn := func(tx *gorm.DB, table, name, colType, defaultVal string) error {
```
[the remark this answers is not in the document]
jeremyfaller: I put it there because it seems a common error-prone motif. I'll move it.

pkg/controller/associated/ios.go
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+package associated
+
+// The iOS format is specified by:
+//   https://developer.apple.com/documentation/safariservices/supporting_associated_domains
+
+import "github.com/google/exposure-notifications-verification-server/pkg/database"
+
+type IosData struct {
```
icco, line 22: nit: I would expect this to be IOSData given how you're capitalizing it in other places.

pkg/controller/apikey/create.go
```diff
@@ -16,19 +16,62 @@ package apikey
...
+	ErrBadSHALength     = errors.New("sha wasn't 95 characters")
+	ErrBadSHAFormat     = errors.New("sha must be of the form AA:...:BB")
+)
+
+type FormData struct {
+	Name  string               `form:"name"`
+	Type  database.APIUserType `form:"type"`
+	OS    database.OsType      `form:"os"`
+	AppID string               `form:"appId"`
+	Sha   string               `form:"sha"`
```
icco, line 39: nit: I'd expect this to be SHA

terraform/service_redirect.tf
```diff
@@ -31,6 +31,19 @@ resource "google_service_account_iam_member" "cloudbuild-deploy-enx-redirect" {
...
+  ])
+
+  secret_id = google_secret_manager_secret.db-secret[each.key].id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "serviceAccount:${google_service_account.server.email}"
+}
+
 resource "google_project_iam_member" "enx-redirect-observability" {
   for_each = toset([
     "roles/cloudtrace.agent",
```
icco, line 49: Github isn't letting me leave a comment, but in the `dynamic "env" {` block you need to add the configs for the cache and database.

cmd/server/assets/apikeys/show.html
```diff
@@ -57,6 +57,32 @@ <h1>{{$authApp.Name}} API key</h1>
           Unknown
           {{end}}
         </div>
+
+        <strong>OS</strong>
```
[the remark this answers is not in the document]
jeremyfaller: done

jeremyfaller:
Seth:

I think I addressed your comments here by improving the editing of the keys (to allow them to move between OS's etc.).

I still need to improve caching (of the JSON responses), so this isn't quite complete. But I will either address that in a followon, or this CL can go. Up to you.

jeremyfaller:
Caching is in now as well.

Unless you'd like to see something different from the design, @sethvargo, this is ready for your review.

jeremyfaller:
unintentionally closed.

cmd/server/assets/apikeys/new.html
```diff
@@ -49,6 +49,8 @@ <h1>New API key</h1>
             {{end}}
           </div>
 
+		  {{template "os.html" .}}
```
mikehelmick, line 52: I'm not sure this belongs in the API keys page.

Seems like it should be a separate realm level setting

pkg/database/authorized_app.go
```diff
@@ -61,8 +70,22 @@ type AuthorizedApp struct {
 	// APIKey is the HMACed API key.
 	APIKey string `gorm:"type:varchar(512);unique_index"`
 
-	// APIKeyType s the API key type.
+	// APIKeyType is the API key type.
 	APIKeyType APIUserType `gorm:"default:0"`
+
+	// OS is the type of the application we're using (eg, iOS, Android).
+	OS OsType `gorm:"os,type:int"`
```
mikehelmick, line 77: Having these on API keys isn't going to work - because API keys will be rotated eventually.
the current positioning would actually let a realm have multiple apps per-platform claimed, and I don't think we want that either.

This information should only be set once on the realm.

mikehelmick (changes requested):

cmd/server/assets/apikeys/new.html
jeremyfaller: Ok, so it looks like the original proposal I wrote in the issue (to put this in Health Authority) won't work. I'll move to a separate table.

/hold

mikehelmick: My fault - I didn't get to the proposal on the issue before the PR was sent

cmd/server/assets/apikeys/os.html
```diff
@@ -0,0 +1,76 @@
...
+    {{if $authApp.ErrorsFor "iosAppID"}}
+    <div class="invalid-feedback">
+      {{joinStrings ($authApp.ErrorsFor "iosAppID") ", "}}
+    </div>
+    {{end}}
+  </div>
+
+  <div class="form-label-group" id="androidDIV">
+    <input type="text" id="androidAppID" name="androidAppID" class="form-control{{if $authApp.ErrorsFor "androidAppID"}} is-invalid{{end}}" value="{{$authApp.AndroidAppID}}" placeholder="Application ID" autofocus>
+    <label for="androidAppID">Android Application ID</label>
```
peetsasaki, line 29: nit: By "Application ID" do we mean package name, e.g. "com.example.app"? If so, might call it "Package Name" in the UI since that's understood by Android devs and consistent with Google's docs re app links https://developer.android.com/training/app-links/verify-site-associations

cmd/server/assets/apikeys/os.html
```diff
@@ -0,0 +1,76 @@
...
+
+  <div class="form-label-group" id="shaDIV">
+    <input type="text" id="sha" name="sha" class="form-control{{if $authApp.ErrorsFor "sha"}} is-invalid{{end}}" value="{{$authApp.SHA}}" placeholder="SHA" autofocus>
+    <label for="sha">SHA</label>
+    {{if $authApp.ErrorsFor "sha"}}
+    <div class="invalid-feedback">
+      {{joinStrings ($authApp.ErrorsFor "sha") ", "}}
+    </div>
+    {{end}}
+  </div>
```
peetsasaki, line 45: I can see in other files that the underlying storage & code support multiple SHA hashes per app, which is great. I'm not familiar with this html templating engine, so I can't tell if the UI intends to support multiple hashes "top to bottom". If so great, but if not, it might be good to do so, to help smooth integration with the app, given developers will need to test with a debug-keyed build prior to publishing a release-keyed build.

cmd/server/assets/apikeys/os.html
```diff
@@ -0,0 +1,76 @@
...
+    {{if $authApp.ErrorsFor "os"}}
+    <div class="invalid-feedback">
+      {{joinStrings ($authApp.ErrorsFor "os") ", "}}
+    </div>
+    {{end}}
+  </div>
+
+  <div class="form-label-group" id="iosDIV">
+    <input type="text" id="iosAppID" name="iosAppID" class="form-control{{if $authApp.ErrorsFor "iosAppID"}} is-invalid{{end}}" value="{{$authApp.IOSAppID}}" placeholder="Application ID" autofocus>
+    <label for="iosAppID">iOS Application ID</label>
```
peetsasaki, line 19: nit: Similarly to below: if "Application ID" here means "bundle identifier", it might be good to call it that in UI.

cmd/server/assets/apikeys/os.html
jeremyfaller: Do we require multiple SHA hashes per app? Currently, I don't support this, but could.

cmd/server/assets/apikeys/os.html
jeremyfaller: I'll update the UI to move to bundle id for android.

cmd/server/assets/apikeys/os.html
jeremyfaller: I've moved to BundleID, LMK if something else is preferrable.

pkg/database/authorized_app.go
jeremyfaller: Moved to a non-per-realm basis.

cmd/server/assets/apikeys/new.html
jeremyfaller: Moved separately.

jeremyfaller:
There is one outstanding issue before this is ready for review -- I need clarification on whether or not an android app could have multiple bundle ids.

@peetsasaki please comment

jeremyfaller:
More research leads me to believe android does need multiple SHAs per app. Updating the CL. continue to hold.

cmd/server/assets/apikeys/os.html
jeremyfaller: Moved to supporting multiple SHAs for an android app.

jeremyfaller:
/unhold

This is ready for review.

cmd/server/assets/mobileapp/edit.html
```diff
@@ -0,0 +1,57 @@
...
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <form method="POST" action="/mobileapp/{{$app.ID}}">
+          <input type="hidden" name="_method" value="PATCH">
+          {{ .csrfField }}
+
+          <div class="form-label-group">
+            <input type="text" id="name" name="name" class="form-control{{if $app.ErrorsFor "name"}} is-invalid{{end}}" value="{{$app.Name}}" placeholder="Application name" autofocus>
+            <label for="name">Application name</label>
+            {{if $app.ErrorsFor "name"}}
```
sethvargo, line 33: errorable

cmd/server/assets/mobileapp/app.html
```diff
@@ -0,0 +1,52 @@
+{{$app := .app}}
```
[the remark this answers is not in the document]
sethvargo: And it should probably be `_form.html`  for consistency since this appears to be a partial.

cmd/server/assets/mobileapp/index.html
```diff
@@ -0,0 +1,84 @@
...
+  {{template "head" .}}
+</head>
+
+<body class="tab-content">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Mobile Apps</h1>
```
sethvargo, line 17:
```suggestion
    <h1>Mobile apps</h1>
```

cmd/server/assets/mobileapp/edit.html
```diff
@@ -0,0 +1,57 @@
...
+  {{template "floatingform" .}}
+</head>
+
+<body class="tab-content" onload="osSelect()">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Edit Mobile App</h1>
```
sethvargo, line 18:
```suggestion
    <h1>Edit mobile app</h1>
```

cmd/server/assets/mobileapp/app.html
```diff
@@ -0,0 +1,52 @@
...
+    <div class="invalid-feedback">
+      {{joinStrings ($app.ErrorsFor "sha") ", "}}
+    </div>
+    {{end}}
+  <div class="form-label-group" id="shaDIV">
+    <textarea type="text" id="sha" name="sha" class="form-control{{if $app.ErrorsFor "sha"}} is-invalid{{end}}" rows=5 autofocus>{{$app.SHA}}</textarea>
+  </div>
+
+  <script>
+    function osSelect() {
```
sethvargo, line 36: We should bind this to the select box via jquery or javascript instead of relying on the HTML author to add `osSelect` as an onChange handler.

cmd/server/assets/mobileapp/edit.html
```diff
@@ -0,0 +1,57 @@
...
+            {{if $app.ErrorsFor "name"}}
+            <div class="invalid-feedback">
+              {{joinStrings ($app.ErrorsFor "name") ", "}}
+            </div>
+            {{end}}
+          </div>
+
+		  {{template "app.html" .}}
+
+          <button type="submit" class="btn btn-primary btn-block">Update Mobile App</button>
```
sethvargo, line 42:
```suggestion
          <button type="submit" class="btn btn-primary btn-block">Update mobile app</button>
```

cmd/server/assets/mobileapp/new.html
```diff
@@ -0,0 +1,50 @@
...
+  {{template "floatingform" .}}
+</head>
+
+<body class="tab-content" onload="osSelect()">>
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>New Mobile App</h1>
```
sethvargo, line 18:
```suggestion
    <h1>New mobile app</h1>
```

cmd/server/assets/mobileapp/app.html
sethvargo, line 1: These assets should be under mobileapp**s** `mobileapps`

sethvargo: And the templates should be name as-such.

cmd/server/assets/mobileapp/new.html
```diff
@@ -0,0 +1,50 @@
...
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <form method="POST" action="/mobileapp">
+          {{ .csrfField }}
+
+          <div class="form-label-group">
+            <input type="text" id="name" name="name" class="form-control{{if $app.ErrorsFor "name"}} is-invalid{{end}}" value="{{$app.Name}}" placeholder="Application name" autofocus>
+            <label for="name">Application name</label>
+            {{if $app.ErrorsFor "name"}}
```
sethvargo, line 32: errorable

cmd/server/assets/mobileapp/new.html
```diff
@@ -0,0 +1,50 @@
...
+
+    <h1>New Mobile App</h1>
+    <p>
+      Use the form below to create a new mobile app.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <form method="POST" action="/mobileapp">
```
sethvargo, line 26: You need to add `class="floating-form"` to get the good HTML floating forms

cmd/server/assets/mobileapp/show.html
```diff
@@ -0,0 +1,76 @@
...
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>{{$app.Name}}</h1>
+    <p class="float-right">
+      <a href="/mobileapp/{{$app.ID}}/edit">Edit</a>
+    </p>
+    <p>
+      Here is information about the Mobile App.
```
sethvargo, line 24:
```suggestion
      Here is information about the mobile app.
```

cmd/server/assets/mobileapp/show.html
```diff
@@ -0,0 +1,76 @@
...
+        <strong>OS</strong>
+        <div>
+          {{if (eq $app.OS .iOS)}}
+          iOS
+          {{else if (eq $app.OS .android)}}
+          Android
+          {{end}}
+        </div>
+
+		{{if (eq $app.OS .iOS)}}
```
sethvargo, line 43: TvsS

cmd/server/assets/mobileapp/show.html
```diff
@@ -0,0 +1,76 @@
...
+          <strong>iOS App ID:</strong>
+          <div>
+          {{$app.AppID}}
+          </div>
+        {{end}}
+
+		{{if (eq $app.OS .android)}}
```
sethvargo, line 50: TvsS

cmd/server/main.go
```diff
@@ -301,6 +302,29 @@ func realMain(ctx context.Context) error {
 		sub.Handle("/{uuid}/expire", codeStatusController.HandleExpirePage()).Methods("PATCH")
 	}
 
+	// mobileapp
+	{
+		sub := r.PathPrefix("/mobileapp").Subrouter()
```
sethvargo, line 307: This should be `/mobileapps` or, better `/mobile-apps`

cmd/server/main.go
```diff
@@ -301,6 +302,29 @@ func realMain(ctx context.Context) error {
...
+		sub.Use(requireAdmin)
+		sub.Use(requireVerified)
+		sub.Use(requireMFA)
+		sub.Use(rateLimit)
+
+		mobileController := mobileapp.New(ctx, config, cacher, db, h)
+		sub.Handle("", mobileController.HandleIndex()).Methods("GET")
+		sub.Handle("", mobileController.HandleCreate()).Methods("POST")
+		sub.Handle("/new", mobileController.HandleCreate()).Methods("GET")
+		sub.Handle("/{id}/edit", mobileController.HandleUpdate()).Methods("GET")
```
sethvargo, line 321: Add constraints on these IDs (we haven't backported to the other things yet), but they should be:

```
{id:[0-9]+}
```

pkg/controller/associated/android.go
```diff
@@ -0,0 +1,54 @@
...
+	Namespace    string   `json:"namespace,omitempty"`
+	PackageName  string   `json:"package_name,omitempty"`
+	Fingerprints []string `json:"sha256_cert_fingerprints,omitempty"`
+}
+
+// getAndroidData finds all the android data apps.
+func (c *Controller) getAndroidData() ([]AndroidData, error) {
+	apps, err := c.db.FindActiveAppsByOS(database.OSTypeAndroid)
+	if err != nil {
+		return nil, err
```
sethvargo, line 38:
```suggestion
		return nil, fmt.Errorf("failed to get android data: %w", err)
```

pkg/controller/associated/android.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	ret := make([]AndroidData, 0, len(apps))
+	for i := range apps {
+		ret = append(ret, AndroidData{
+			Relation: []string{"delegate_permission/common.handle_all_urls"},
+			Target: Target{
+				Namespace:    "android_app",
+				PackageName:  apps[i].AppID,
+				Fingerprints: strings.Split(apps[i].SHA, "\n"),
```
sethvargo, line 48: Yea, I think we need some more intelligence and sanitization here. We should be a helper and be tolerant to blank fingerprints, CSV instead of NLSV, etc.

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type Controller struct {
+	ctxt   context.Context
+	db     *database.Database
+	cacher cache.Cacher
+	h      *render.Renderer
+}
+
+var ttl = 30 * time.Minute
```
sethvargo, line 43: const?

sethvargo: and doc

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (c *Controller) HandleIos() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := r.URL.String()
+
+		// Check the cache.
```
sethvargo, line 49: You should be able to use `c.cacher.Fetch` here instead of a manual read, create, write. 

```go
var data IOSData
if err := c.cacher.Fetch(ctx, key, &data, ttl, func() (interface{}, error){
  return c.getIosData()
}); err != nil {
  // handle error
}
```

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/google/exposure-notifications-verification-server/pkg/api"
+	"github.com/google/exposure-notifications-verification-server/pkg/cache"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/render"
+)
+
+type Controller struct {
+	ctxt   context.Context
```
sethvargo, line 37: Don't store contexts on structs - it's a [Go no-no](https://golang.org/pkg/context/):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+var ttl = 30 * time.Minute
+
+func (c *Controller) HandleIos() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := r.URL.String()
+
+		// Check the cache.
+		{
+			var data IOSData
+			if err := c.cacher.Read(c.ctxt, key, &data); err == nil {
```
sethvargo, line 52: This context should be pulled from the request (`r.Context()`) or passed in to `HandleIos`, depending on who we want to control cancelation.

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+
+		// Not in cache, go get it.
+		data, err := c.getIosData()
+		if err != nil {
+			c.h.RenderJSON(w, http.StatusOK, api.Error(err))
+			return
+		}
+
+		// Cache the data.
+		if err := c.cacher.Fetch(c.ctxt, key, &data, ttl, func() (interface{}, error) {
```
sethvargo, line 69: Yea, I think you can simplify this greatly but just calling getIosData() in here an removing everything above.

pkg/controller/associated/index.go
```diff
@@ -0,0 +1,116 @@
...
+
+		c.h.RenderJSON(w, http.StatusOK, data)
+	})
+}
+
+func (c *Controller) HandleAndroid() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := r.URL.String()
+
+		// Check the cache
```
sethvargo, line 84: Same notes about cache and context.

pkg/controller/mobileapp/create.go
```diff
@@ -0,0 +1,171 @@
...
+type FormData struct {
+	Name  string          `form:"name"`
+	OS    database.OSType `form:"os"`
+	AppID string          `form:"appID"`
+	SHA   string          `form:"sha"`
+}
+
+// cleanSHA cleans an input string.
+func cleanSHA(input string) string {
+	input = strings.ToUpper(input)
```
sethvargo, line 43: This is pretty expensive... it probably doesn't matter for this use case, but ReplaceAll is an O(n) op and we're calling it 5x

sethvargo: Also, it feels like `clean` and `verify` should be the same function.

pkg/controller/mobileapp/create.go
```diff
@@ -0,0 +1,171 @@
...
+		// Build the authorized app struct
+		app := &database.MobileApp{
+			Name:  form.Name,
+			OS:    form.OS,
+			AppID: form.AppID,
+			SHA:   form.SHA,
+		}
+
+		// Verify the form.
+		if err := form.verify(); err != nil {
```
sethvargo, line 145: Hmm - what does this do?

pkg/controller/mobileapp/create.go
```diff
@@ -0,0 +1,171 @@
...
+		err := realm.CreateMobileApp(c.db, app)
+		if err != nil {
+			flash.Error("Failed to create mobile app: %v", err)
+			c.renderNew(ctx, w, app)
+			return
+		}
+
+		// Store the app name in the session so we can properly display it on
+		// the next page.
+		session.Values["appName"] = form.Name
```
sethvargo, line 159: Hmm - why do we need this on the next page? It's generally frowned upon to store things in the session that aren't long-term persisted. We use flash for short-term things.

pkg/controller/mobileapp/create.go
```diff
@@ -0,0 +1,171 @@
...
+
+		flash.Alert("Successfully created Mobile App %v", form.Name)
+		http.Redirect(w, r, fmt.Sprintf("/mobileapp/%d", app.ID), http.StatusSeeOther)
+	})
+}
+
+// renderNew renders the new page.
+func (c *Controller) renderNew(ctxt context.Context, w http.ResponseWriter, app *database.MobileApp) {
```
sethvargo, line 167: context should be named `ctx` throughout (that's a Go-ism)

pkg/controller/mobileapp/disable.go
```diff
@@ -0,0 +1,62 @@
...
+				return
+			}
+
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		if err := app.Disable(c.db); err != nil {
+			flash.Error("Failed to disable mobile app: %v", err)
+			http.Redirect(w, r, "/mobileapp", http.StatusSeeOther)
```
sethvargo, line 56: Need a `return` after the redirect.

pkg/controller/mobileapp/show.go
```diff
@@ -0,0 +1,78 @@
...
+			m := templateMap(ctxt)
+			m["appName"] = appName
+			delete(session.Values, "appName")
+		}
+
+		// Pull the authorized app from the id.
+		app, err := realm.FindMobileApp(c.db, vars["id"])
+		if err != nil {
+			if database.IsNotFound(err) {
+				logger.Debugw("mobile app does not exist", "id", vars["id"])
```
sethvargo, line 60: Drop this log message

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+	Errorable
+
+	// Name is the name of the app.
+	Name string `gorm:"name,varchar(512);unique_index:realm_app_name"`
+
+	// RealmID is the id of the mobile app.
+	RealmID uint `gorm:"unique_index:realm_app_name"`
+
+	// OS is the type of the application we're using (eg, iOS, Android).
+	OS OSType `gorm:"os,type:int"`
```
sethvargo, line 39:
```suggestion
	OS OSType `gorm:"column:os; type:int"`
```

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+	OSTypeIOS OSType = iota
+	OSTypeAndroid
+)
+
+type MobileApp struct {
+	gorm.Model
+	Errorable
+
+	// Name is the name of the app.
+	Name string `gorm:"name,varchar(512);unique_index:realm_app_name"`
```
sethvargo, line 33:
```suggestion
	Name string `gorm:"column:name; type:citext; unique_index:realm_app_name;"`
```

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// RealmID is the id of the mobile app.
+	RealmID uint `gorm:"unique_index:realm_app_name"`
+
+	// OS is the type of the application we're using (eg, iOS, Android).
+	OS OSType `gorm:"os,type:int"`
+
+	// IOSAppID is a unique string representing the app.
+	AppID string `gorm:"app_id,type:varchar(512)"`
```
sethvargo, line 42:
```suggestion
	AppID string `gorm:"column:app_id; type:varchar(512);"`
```

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// SHA is a unique hash of the app.
+	// It is only present for Android devices, and should be of the form:
+	//   AA:BB:CC:DD...
+	SHA string `gorm:"sha,type:text"`
```
sethvargo, line 47:
```suggestion
	SHA string `gorm:"column:sha; type:varchar(TODO);"`
```

sethvargo: We know the max length of the SHA, right?

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// FindActiveAppsByOS finds all authorized by their OS.
+func (db *Database) FindActiveAppsByOS(os OSType) ([]*MobileApp, error) {
+	// Find the API key that matches the constraints.
+	var apps []*MobileApp
+	if err := db.db.
+		Model(&MobileApp{}).
+		Where("os = ?", os).
+		Where("deleted_at IS NULL").
```
sethvargo, line 57: This is part of the default query - you can drop it.

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+		Find(&apps).
+		Error; err != nil {
+		return nil, err
+	}
+	return apps, nil
+}
+
+// CreateMobileApp adds a mobile app to the DB.
+func (r *Realm) CreateMobileApp(db *Database, app *MobileApp) error {
```
sethvargo, line 66: I'm not sure we want mutating functions to live off other database models. This should probably live off of `Database`.

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+	app.RealmID = r.ID
+	return db.db.Save(&app).Error
+}
+
+// SaveMobileApp saves the authorized app.
+func (db *Database) SaveMobileApp(r *MobileApp) error {
+	if r.Model.ID == 0 {
```
sethvargo, line 73: Don't need this - just call `db.db.Save(r).Error`

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+		return db.db.Create(r).Error
+	}
+	return db.db.Save(r).Error
+}
+
+// Disable disables the mobile app.
+func (a *MobileApp) Disable(db *Database) error {
+	if err := db.db.Delete(a).Error; err != nil {
```
sethvargo, line 81: Just return `db.db.Delete(a).Error`

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	return nil
+}
+
+// Enable enables the mobile app.
+func (a *MobileApp) Enable(db *Database) error {
+	if err := db.db.Unscoped().Model(a).Update("deleted_at", nil).Error; err != nil {
```
sethvargo, line 89: Just return

pkg/database/mobile_app.go
```diff
@@ -0,0 +1,93 @@
...
+	AppID string `gorm:"app_id,type:varchar(512)"`
+
+	// SHA is a unique hash of the app.
+	// It is only present for Android devices, and should be of the form:
+	//   AA:BB:CC:DD...
+	SHA string `gorm:"sha,type:text"`
+}
+
+// FindActiveAppsByOS finds all authorized by their OS.
+func (db *Database) FindActiveAppsByOS(os OSType) ([]*MobileApp, error) {
```
sethvargo, line 51: This should be `List`, not `Find`

pkg/database/realm.go
```diff
@@ -571,6 +571,38 @@ func (r *Realm) FindAuthorizedApp(db *Database, id interface{}) (*AuthorizedApp,
 	return &app, nil
 }
 
+// ListMobileApps gets all the mobile apps for the realm.
+func (r *Realm) ListMobileApps(db *Database) ([]*MobileApp, error) {
+	var apps []*MobileApp
+	if err := db.db.
+		Unscoped().
```
sethvargo, line 578: Why unscoped? that would include deleted.

pkg/database/realm.go
```diff
@@ -571,6 +571,38 @@ func (r *Realm) FindAuthorizedApp(db *Database, id interface{}) (*AuthorizedApp,
...
+		Model(r).
+		Order("mobile_apps.deleted_at DESC, LOWER(mobile_apps.name)").
+		Related(&apps).
+		Error; err != nil {
+		if IsNotFound(err) {
+			return nil, nil
```
sethvargo, line 584:
```suggestion
			return apps, nil
```

We want to give people the empty list instead of nil.

pkg/database/realm.go
```diff
@@ -571,6 +571,38 @@ func (r *Realm) FindAuthorizedApp(db *Database, id interface{}) (*AuthorizedApp,
...
+	return apps, nil
+}
+
+// FindMobileApp finds the mobile app by the given id associated with the realm.
+func (r *Realm) FindMobileApp(db *Database, id interface{}) (*MobileApp, error) {
+	var app MobileApp
+	if err := db.db.
+		Unscoped().
+		Model(MobileApp{}).
+		Order("LOWER(name) ASC").
```
sethvargo, line 597: You're only expecting one record - no need to order

sethvargo: And there's a unique index

terraform/service_redirect.tf
```diff
@@ -69,6 +137,8 @@ resource "google_cloud_run_service" "enx-redirect" {
           for_each = merge(
             local.gcp_config,
             local.enx_redirect_config,
+            local.cache_config,
+            local.database_config,
```
sethvargo, line 141: I think you nee csrf and some other things too?

sethvargo (changes requested):

cmd/enx-redirect/main.go
```diff
@@ -58,14 +61,14 @@ func main() {
 func realMain(ctx context.Context) error {
 	logger := logging.FromContext(ctx)
 
-	config, err := config.NewRedirectConfig(ctx)
```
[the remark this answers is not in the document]
jeremyfaller: Sure, I'll followup with all the other servers. I think I can cobble a `gofmt` script to do it.

cmd/server/assets/mobileapp/app.html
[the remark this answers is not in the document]
jeremyfaller: The `{{template "app.html" .}}` worked as it was, but NBD your way, and more consistent!

pkg/database/realm.go
jeremyfaller: I went with the same approach of being able to resurrect apps (like you can resurrect keys). So, this is A-OK (the UI supports it, etc.)

image

terraform/service_redirect.tf
jeremyfaller: @icco Can you comment here? I don't want to mess this up, like I did jwks, and I am **NOT** the person you want to leave as a babysitter on this one. :)

pkg/database/realm.go
jeremyfaller: hehe. they're _mostly_ the same, but you're right -- not exactly.

cmd/enx-redirect/main.go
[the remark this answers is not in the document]
jeremyfaller: https://github.com/google/exposure-notifications-verification-server/issues/674

I'll file similar for ENS.

terraform/service_redirect.tf
icco: Looking at cmd/enx-redirect/main.go I don't think you should need any other configs.

Issue #629, opened 2020-09-22
Flash not showing on signout


whaught:
### TL;DR
We used to wait 3 seconds to redirect from the /signout page. This is enough time to read error flashes - now the flash is displayed and we immediately redirect.

We could consider a way to hang on to the session flash and carry-forward in middleware or bake signout into login to eliminate the redirect

sethvargo:
If we can move all the auth and signout to server-side, this problem goes away too 😄

PR #636, opened 2020-09-22
Wait for signout redirect if there is an error

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/629

## Proposed Changes

* For now, keep the 3 second wait on signout redirect but only if there are children in the flash container
* Minor formatting


sethvargo (approved):
/lgtm

Issue #432, opened 2020-09-01
Add system level twilio config that a system admin can 'share' with a realm


mikehelmick:
/kind enhancement

jeremyfaller:
Alright, getting started on this, and need some clarification: How are we looking to share this file? Is it the defaults loaded into a new realm, can a running realm be updated to the settings?

Can you walk me through a little of what you have in mind, and I'll fill in the more details?

sethvargo:
Right now there's an "SMSConfigs" table that has a 1-1 mapping with a realm. The easiest thing to do would be to create a "system" realm where a sysadmin can upload some twilio creds using the existing forms, then provide a way to attach that SMSConfig to another realm (changing the relationship from 1-1 to 1-n)

jeremyfaller:
And we should still allow a realm to its own config, correct? (So, a realm would have the system level config unless it overrides it, correct?)

sethvargo:
Yea, I think the precedence is realm-specific, then system-specific (if shared with that realm, else empty (no sms)

mikehelmick:
I'm encouraging realm operators to bring their own twilio credentials, but it's possible a system operator would want to just provide that.

PR #641, opened 2020-09-22
Add forms for system SMS configs and sharing logic

sethvargo:
Fixes GH-432

**Release Note**

```release-note
Add forms for system SMS configs and sharing logic
```

### DIFF
--- a/pkg/database/sms_config.go
+++ b/pkg/database/sms_config.go
@@ -16,6 +16,7 @@ package database
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/sms"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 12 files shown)


pkg/database/sms_config.go
```diff
@@ -71,6 +85,20 @@ func (db *Database) SaveSMSConfig(s *SMSConfig) error {
 			return nil
 		}
 
+		if s.IsSystem {
+			// We're about to delete the system SMS config, revoke everyone's
+			// permissions to use it. You would think there'd be a way to do this with
+			// gorm natively. You'd even find an example in the documentation that led
+			// you to believe that gorm does this.
+			//
+			// Narrator: gorm does not do this.
```
whaught, line 94: This surprises me

whaught (commented):
/lgtm

pkg/database/sms_config.go
sethvargo: Yea, if you actually print out the SQL for the example in gorm's docs, it's `SELECT * FROM realms WHERE id = '' UPDATE ...` which isn't even valid SQL

whaught: ORMs are probably expecting row-wise object updates rather than a blind update of a whole column

PR #643, opened 2020-09-23
Handle re-auth redirect

whaught:
## Proposed Changes

* Phone enrollment can fail if the session is old. Firebase needs re-authentication for sessions that aren't "recent" (which is vaguely unspecified)

**Release Note**

```release-note
Redirect to login if re-authorization is required
```


sethvargo (approved):
/lgtm

PR #644, opened 2020-09-23
Add application-level per-realm firewall configuration

sethvargo:
This adds per-realm firewall settings (by allowed CIDR blocks) for the adminapi, apiserver, and main server. 

**Release Note**

```release-note
Add per-realm firewall security settings
```

/assign @mikehelmick

### DIFF
--- a/cmd/server/assets/realmadmin/_form_security.html
+++ b/cmd/server/assets/realmadmin/_form_security.html
@@ -79,7 +79,8 @@
       all traffic is allowed from all IPs. These should be of <a
       href="https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing"
       target="_BLANK">CIDR notation</a>
-      of the format (e.g. <code>192.1.2.0/24</code>).
+      of the format (e.g. <code>192.1.2.0/24</code>). This is only enforced
+      post-authentication.
     </small>
   </div>
 
@@ -96,7 +97,8 @@
       from all IPs. These should be of <a
       href="https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing"
       target="_BLANK">CIDR notation</a>
-      of the format (e.g. <code>192.1.2.0/24</code>).
+      of the format (e.g. <code>192.1.2.0/24</code>). This is only enforced
+      post-authentication.
     </small>
   </div>
 
@@ -111,7 +113,8 @@
       from all IPs. These should be of <a
       href="https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing"
       target="_BLANK">CIDR notation</a>
-      of the format (e.g. <code>192.1.2.0/24</code>).
+      of the format (e.g. <code>192.1.2.0/24</code>). This is only enforced
+      post-authentication.
     </small>
   </div>
 
--- a/cmd/server/assets/realmadmin/_form_security.html
+++ b/cmd/server/assets/realmadmin/_form_security.html
@@ -17,6 +17,7 @@
       <option value="0" {{if eq $realm.EmailVerifiedMode.String "prompt"}}selected{{end}}>Prompt after login</option>
       <option value="2" {{if eq $realm.EmailVerifiedMode.String "optional"}}selected{{end}}>Optional</option>
     </select>
+    {{template "errorable" $realm.ErrorsFor "emailVerifiedMode"}}
     <small class="form-text text-muted">
       Email verification requires users to verify their email address before
       using the system.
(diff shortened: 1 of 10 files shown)


mikehelmick:
/hold

I think this is dangerous for api server - that’s intended to be accessed by mobile devices

sethvargo:
I wrote the help text to strongly discourage people from enabling the firewall against the apiserver, but I do think there's a legit use case where someone running their own server might want to do so. For example, they might be running a staging server and want to restrict devices on a particular subnet from uploading to it. One could argue they should add that protection at another layer, but I do see the use case.

sherifkozman:
@sethvargo i believe it had to be on a different layer as well , where logging will be normally available in case the wrong CIDR block is added there will be no way to trace or debug.

cmd/server/assets/realmadmin/_form_security.html
```diff
@@ -58,12 +61,60 @@
...
+      a certificate. It is highly recommended that you <strong>leave this
+      service publicly accessible</strong>. If blank, all traffic is allowed
+      from all IPs. These should be of <a
+      href="https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing"
+      target="_BLANK">CIDR notation</a>
+      of the format (e.g. <code>192.1.2.0/24</code>).
+    </small>
+  </div>
+
+  <div class="form-label-group">
```
mikehelmick, line 103: add a note that this is enforced POST login only?

pkg/controller/middleware/firewall.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+
+			var allowedCIDRs []string
+			switch typ {
+			case "adminapi":
+				allowedCIDRs = realm.AllowedCIDRsAdminAPI
+			case "apiserver":
+				allowedCIDRs = realm.AllowedCIDRsAPIServer
+			case "server":
+				allowedCIDRs = realm.AllowedCIDRsServer
```
mikehelmick, line 55: case default:
   log - you did a bad?

pkg/controller/middleware/firewall.go
```diff
@@ -0,0 +1,94 @@
...
+			ipStr := r.RemoteAddr
+
+			// Check if x-forwarded-for exists, the load balancer sets this, and the
+			// first entry is the real client IP.
+			xff := r.Header.Get("x-forwarded-for")
+			if xff != "" {
+				ipStr = strings.Split(xff, ",")[0]
+			}
+
+			ip := net.ParseIP(ipStr)
```
mikehelmick, line 76: do we need to nil check ip?

pkg/database/migrations.go
```diff
@@ -1275,6 +1275,27 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 					}
 				}
 
+				return nil
+			},
+		},
+		{
+			ID: "00052-CreateRealmAllowedCIDRs",
+			Migrate: func(tx *gorm.DB) error {
+				return tx.AutoMigrate(&Realm{}).Error
```
mikehelmick, line 1284: u did an auto migrate?!?!?

pkg/database/realm.go
```diff
@@ -816,3 +823,34 @@ func (r *Realm) RenderWelcomeMessage() string {
...
+			// rudimentary.
+			if !strings.Contains(v, "/") {
+				if strings.Contains(v, ":") {
+					v = fmt.Sprintf("%s/128", v)
+				} else {
+					v = fmt.Sprintf("%s/32", v)
+				}
+			}
+
+			cidrs = append(cidrs, v)
```
mikehelmick, line 850: should we ParseCIDR and check for errors here?

cmd/server/assets/realmadmin/_form_security.html
sethvargo: Done

pkg/controller/middleware/firewall.go
sethvargo: Done

pkg/controller/middleware/firewall.go
sethvargo: Done

pkg/database/migrations.go
sethvargo: Don't tell

pkg/database/realm.go
sethvargo: Done

sethvargo:
@sherifkozman the logs will include a message if a request is blocked due to the IP. It won't tell you _which_ IP, but you can correlate it to timestamps.

This can't be higher in the stack because it's a realm-level configuration, and users can be a member of multiple realms. For example, Jacky might be able to access realm1 from anywhere, but can only access realm2 from a corporate IP subnet.

sethvargo:
@mikehelmick PTAnotherL

mikehelmick:
> @mikehelmick PTAnotherL

👁️ looking

mikehelmick (approved):

PR #645, opened 2020-09-23
Add monitoring and alerting for realm capacity

### LINKED ISSUE #572: Set up monitoring and alerting when realms are at certain capacity levels
Set up monitoring and alerting when realms are at certain capacity levels (e.g. 90%). These will have to go to the server operators since we don't have a way to send email to realm admins.

There are already stats - we just need the monitoring and alerts.

/cc @icco @yegle - do one of you have capacity to work on this?

femnad:
Fixes #572

## Proposed Changes

* Record metric for realm capacity on issuing a verification token
* Add dashboard and alerting for monitoring realm capacity

**Release Note**

```release-note
Record and monitor realm verification token capacity
```

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -203,7 +203,7 @@ func (c *Controller) HandleIssue() http.Handler {
 			return
 		}
 		key := fmt.Sprintf("realm:quota:%s", dig)
-		limit, _, reset, ok, err := c.limiter.Take(ctx, key)
+		limit, remaining, reset, ok, err := c.limiter.Take(ctx, key)
 		if err != nil {
 			logger.Errorw("failed to take from limiter", "error", err)
 			stats.Record(ctx, c.metrics.QuotaErrors.M(1))
@@ -255,13 +255,7 @@ func (c *Controller) HandleIssue() http.Handler {
 			return
 		}
 
-		err = c.recordCapacity(ctx, realm)
-		if err != nil {
-			logger.Errorw("failed to record realm capacity", "error", err)
-			stats.Record(ctx, c.metrics.CodeIssueErrors.M(1))
-			c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to update realm capacity"))
-			return
-		}
+		c.recordCapacity(ctx, realm, remaining)
 
 		if request.Phone != "" && smsProvider != nil {
 			message := realm.BuildSMSText(code, longCode, c.config.GetENXRedirectDomain())
@@ -306,25 +300,13 @@ func (c *Controller) getAuthorizationFromContext(r *http.Request) (*database.Aut
 	return authorizedApp, user, nil
 }
 
-func (c *Controller) recordCapacity(ctx context.Context, realm *database.Realm) error {
+func (c *Controller) recordCapacity(ctx context.Context, realm *database.Realm, remaining uint64) {
 	if !realm.AbusePreventionEnabled {
-		return nil
+		return
 	}
 
-	dayStart := time.Now().UTC().Truncate(24 * time.Hour)
-	dayEnd := dayStart.Add(1)
-
-	realmStats, err := realm.Stats(c.db, dayStart, dayEnd)
-	if err != nil {
-		return err
-	}
-	if len(realmStats) != 1 {
-		return errors.New("found more than one row of daily stats")
-	}
-
-	dailyStats := realmStats[0]
-	capacity := 1 - (float64(dailyStats.CodesIssued) / float64(realm.AbusePreventionEffectiveLimit()))
+	limit := realm.AbusePreventionEffectiveLimit()
+	issued := uint64(limit) - remaining
+	capacity := float64(issued) / float64(limit)
 	stats.Record(ctx, c.metrics.RealmCapacity.M(capacity))
-
-	return nil
 }
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -304,9 +304,12 @@ func (c *Controller) recordCapacity(ctx context.Context, realm *database.Realm, 
 	if !realm.AbusePreventionEnabled {
 		return
 	}
+	stats.Record(ctx, c.metrics.RealmTokenRemaining.M(int64(remaining)))
 
 	limit := realm.AbusePreventionEffectiveLimit()
 	issued := uint64(limit) - remaining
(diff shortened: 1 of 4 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -254,6 +255,14 @@ func (c *Controller) HandleIssue() http.Handler {
 			return
 		}
 
+		err = c.recordCapacity(ctx, realm)
```
sethvargo, line 258: Use:

```go
if err := c.recordCapacity(ctx, realm); err != nil {
...
```

to avoid shadowing err

pkg/controller/issueapi/issue.go
```diff
@@ -254,6 +255,14 @@ func (c *Controller) HandleIssue() http.Handler {
...
+		if err != nil {
+			logger.Errorw("failed to record realm capacity", "error", err)
+			stats.Record(ctx, c.metrics.CodeIssueErrors.M(1))
```
sethvargo, line 261: Is this the right stat?

pkg/controller/issueapi/issue.go
```diff
@@ -254,6 +255,14 @@ func (c *Controller) HandleIssue() http.Handler {
...
+			c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to update realm capacity"))
```
sethvargo, line 262: Are you sure we want to fail if we don't update this stat?

pkg/controller/issueapi/issue.go
```diff
@@ -296,3 +305,26 @@ func (c *Controller) getAuthorizationFromContext(r *http.Request) (*database.Aut
...
+	realmStats, err := realm.Stats(c.db, dayStart, dayEnd)
+	if err != nil {
+		return err
+	}
+	if len(realmStats) != 1 {
+		return errors.New("found more than one row of daily stats")
+	}
+
+	dailyStats := realmStats[0]
+	capacity := 1 - (float64(dailyStats.CodesIssued) / float64(realm.AbusePreventionEffectiveLimit()))
```
sethvargo, line 326: I'm not sure I understand why this is 1 - %?

sethvargo (changes requested):
Hmm I was more thinking we'd generate a metric/alert when a realm reached their capacity (or when the capacity was less than some small number like "10". The result from `Take` should include all the info you need without making the extra db calls.

pkg/controller/issueapi/issue.go
femnad: It felt wrong to alert above a capacity of 90% but maybe the wording on the alert should be "capacity utilization" so changed to that.

pkg/controller/issueapi/issue.go
femnad: In my mind associated that error with the whole process, but now it's not fatal.

pkg/controller/issueapi/issue.go
femnad: Capacity recording does not return an error anymore.

femnad:
> Hmm I was more thinking we'd generate a metric/alert when a realm reached their capacity (or when the capacity was less than some small number like "10". The result from `Take` should include all the info you need without making the extra db calls.

Seeing the return of `Take` now I see that would have been more straightforward. Change to comparing the remaining to 10 (constant) and have a two-state alert based on that?

pkg/controller/issueapi/issue.go
```diff
@@ -254,6 +255,8 @@ func (c *Controller) HandleIssue() http.Handler {
 			return
 		}
 
+		c.recordCapacity(ctx, realm, remaining)
```
sethvargo, line 258: Yea, so we know their total configured limit and the number remaining. I'm gonna defer to @icco on whether we want to alert on a fixed val (e.g. 10) or a ratio here (e.g. < 10% quota remaining).

sethvargo (commented):
/assign @icco

pkg/controller/issueapi/issue.go
icco: We wanna alert on percentage, but the metrics should be the actual numbers. So record remaining and total here, and in the alert calculate the percentage

femnad: Can we have alerting based on multiple metrics? From what I can tell Stackdriver only seemed to consider a single metric, am I wrong?

icco (commented):
.

pkg/controller/issueapi/issue.go
icco: Uhm, my understanding was that we can do math now with Query Notation, but I have not tried

femnad: Yeah, I can do something like the following to calculate memory utilization in metrics explorer via MQL:

```
{ fetch 'compute.googleapis.com/instance/memory/balloon/ram_used'
; fetch 'compute.googleapis.com/instance/memory/balloon/ram_size' }
| join
| div
```
however, couldn't make something similar work via Terraform alert policy filter, it doesn't seem possible to select multiple metrics by `OR`ring the metric type.

I'll dig further, but in the meantime, added metrics for remaining and issued tokens, which is closer to what we want here.

pkg/controller/issueapi/metrics.go
```diff
@@ -117,25 +119,49 @@ func registerMetrics() (*Metrics, error) {
 		return nil, fmt.Errorf("stat view registration failure: %w", err)
 	}
 
-	mRealmCapacity := stats.Float64(MetricPrefix+"/realm_capacity", "Available capacity for issuing verification codes", stats.UnitDimensionless)
+	mRealmTokenRemaining := stats.Int64(MetricPrefix+"/realm_token_remaining", "Remaining number of verification codes", stats.UnitDimensionless)
 	if err := view.Register(&view.View{
-		Name:        MetricPrefix + "/realm_capacity_latest",
-		Description: "Latest realm capacity",
+		Name:        MetricPrefix + "/realm_token_remaining_latest",
```
sethvargo, line 124: Should these be `_count` instead of `_latest`?

femnad: I thought `_count` was for cases where aggregation is `Count`. But do you mean keep the suffix as `_count` regardless of aggregation type?

sethvargo: nvm I think you're right

sethvargo:
/lgtm
/hold

I'd like @icco to give the final approval

femnad:
> New changes are detected. LGTM label has been removed.

Had to rebase on main to fix a conflict.

icco (approved):
/lgtm

PR #647, opened 2020-09-23
Don't log a 500 when it's a user error

sethvargo:
**Release Note**

```release-note
Don't log a 500 when it's a user error
```


whaught (commented):
/lgtm

mikehelmick (approved):

PR #648, opened 2020-09-23
Add hard max and min to modeler

sethvargo:
**Release Note**

```release-note
Add upper max and lower min to modeler
```

### DIFF
--- a/pkg/controller/modeler/modeler_test.go
+++ b/pkg/controller/modeler/modeler_test.go
@@ -219,4 +219,43 @@ func TestRebuildModel(t *testing.T) {
 			t.Errorf("expected %v to be %v", got, want)
 		}
 	}
+
+	// Ensure we hit the ceiling
+	{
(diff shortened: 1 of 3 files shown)


pkg/controller/modeler/modeler_test.go
```diff
@@ -219,4 +219,43 @@ func TestRebuildModel(t *testing.T) {
...
+
+	// Ensure we hit the ceiling
+	{
+		if err := db.RawDB().Exec("TRUNCATE realm_stats").Error; err != nil {
+			t.Fatal(err)
+		}
+
+		line := make([]uint, 24)
+		for i := range line {
+			line[i] = uint(10000 * i)
```
jeremyfaller, line 231: some function on hardMax? Otherwise you'll have to change this test when you change the constant.

mikehelmick (approved):

PR #651, opened 2020-09-23
Add mini firebase auth SDK

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,7 @@ require (
 	go.uber.org/zap v1.16.0
 	golang.org/x/time v0.0.0-20200630173020-3af7569d3a1e
 	golang.org/x/tools v0.0.0-20200908191908-acefd226e2cc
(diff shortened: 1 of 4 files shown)


pkg/internal/firebase/firebase.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2020 Google LLC
```
mikehelmick, line 1: let's put these in just

/internal

I don't want anyone depending on our repo to provide this.

whaught (approved):
/lgtm

mikehelmick (changes requested):

pkg/internal/firebase/firebase.go
sethvargo: They are 😄

mikehelmick (approved):

Issue #609, opened 2020-09-21
Password change without reset email


whaught:
### TL;DR
Currently password-reset email is the only way to change your password. Logged-in users should be able to change it without re-verifying email.

**Proposal**
Let an authenticated user change their password by confirming the old one and creating a new one.

sethvargo:
If they have MFA enabled, we should also re-require it before changing the password.

PR #652, opened 2020-09-23
Allow authed user to change password

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/609

## Proposed Changes

* Allow an authenticated user to change their password
    * No longer need to send a reset password email
    * Handles recent-auth redirect

TODO: Switch this to Seth's server-side client library

**Release Note**

```release-note
Allow users to change their password
```


sethvargo (approved):
/lgtm

PR #653, opened 2020-09-23
Make certificate swapping easier for enx-redirect

icco:
## Proposed Changes

* If you are using enx-redirect, and you modify the map variable, this will recreate the cert, which will create unknown amounts of downtime. So instead, you need to manually create a new cert with the new domains, append it to the FE, and then wait for it to become healthy, and then change the FE again by removing the old cert.
* This changes the FE to use two certs, so that the "main" cert is just `www.` and then the region domains are on a separate cert, this makes the above process easier. 
* Also split the two LBs into separate files.

**Release Note**

```release-note
Changes the certificates on the redirect server.
```


sethvargo (approved):
/lgtm

PR #655, opened 2020-09-23
Use the internal Firebase client to send password reset

whaught:
## Proposed Changes

* Uses Seths internal FB client to send the password reset on user create
* Eliminates a fail-in-the-middle scenario

### DIFF
--- a/pkg/controller/user/importbatch.go
+++ b/pkg/controller/user/importbatch.go
@@ -67,6 +67,10 @@ func (c *Controller) HandleImportBatch() http.Handler {
 				continue
 			} else if created {
 				newUsers = append(newUsers, &batchUser)
+				if err := c.fbInternal.SendPasswordResetEmail(ctx, user.Email); err != nil {
+					batchErr = multierror.Append(batchErr, err)
+					continue
+				}
 			}
 
 			if err := c.db.SaveUser(user); err != nil {
--- a/pkg/controller/user/update.go
+++ b/pkg/controller/user/update.go
@@ -94,13 +94,19 @@ func (c *Controller) HandleUpdate() http.Handler {
 
 		if err := c.db.SaveUser(user); err != nil {
 			flash.Error("Failed to update user: %v", err)
-			c.renderNew(ctx, w, user)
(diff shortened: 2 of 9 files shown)


pkg/controller/user/importbatch.go
```diff
@@ -67,6 +67,10 @@ func (c *Controller) HandleImportBatch() http.Handler {
 				continue
 			} else if created {
 				newUsers = append(newUsers, &batchUser)
+				if err := c.fbInternal.SendPasswordResetEmail(ctx, user.Email); err != nil {
```
sethvargo, line 70: It'd be good to know what quotas, if any, exist for this

mikehelmick (approved):

pkg/controller/user/update.go
```diff
@@ -103,6 +103,12 @@ func (c *Controller) HandleUpdate() http.Handler {
 	})
 }
 
+func (c *Controller) renderUpdate(ctx context.Context, w http.ResponseWriter, user *database.User) {
+	m := controller.TemplateMapFromContext(ctx)
+	m["user"] = user
+	c.h.RenderHTML(w, "users/new", m)
```
sethvargo, line 109: It's a little weird that `renderUpdate` is named "new"

pkg/controller/user/user.go
```diff
@@ -31,24 +32,33 @@ import (
...
+	db         *database.Database
+	h          *render.Renderer
+	logger     *zap.SugaredLogger
 }
 
 // New creates a new controller for managing users.
-func New(ctx context.Context, client *auth.Client, cacher cache.Cacher, config *config.ServerConfig, db *database.Database, h *render.Renderer) *Controller {
+func New(
+	ctx context.Context,
+	fbInternal *iFB.Client,
```
sethvargo, line 47: So I actually have this dream of moving everything to `internal/firebase`, but that's for another day 😄

sethvargo (approved):
/lgtm

pkg/controller/user/update.go
whaught: new and update share the same page...

PR #658, opened 2020-09-23
Server side password selection

whaught:
## Proposed Changes

* Do new-password-selection server-side for password reset email
* Minor code reorganization for less duplication


cmd/server/assets/login/select-password.html
```diff
@@ -20,14 +20,15 @@
...
+                <input type="hidden" id="code" name="code" />
                 <div class="form-label-group">
                   <input type="email" id="email" name="email" class="form-control" placeholder="Email address" required
                     autofocus />
                   <label for="email">Email address</label>
                 </div>
 
                 <div class="form-label-group mb-2">
-                  <input type="password" id="password" class="form-control" placeholder="Password"
+                  <input type="password" id="password" name="password" class="form-control" placeholder="Password"
```
sethvargo, line 31: Nit: order should be `type`, `name`, `id`, `class`...

pkg/controller/context.go
```diff
@@ -31,6 +32,7 @@ const (
 	contextKeySession       = contextKey("session")
 	contextKeyTemplate      = contextKey("template")
 	contextKeyUser          = contextKey("user")
+	contextKeyFBUser        = contextKey("fbUser")
```
sethvargo, line 35: I'd prefer if we spelled out firebase, especially given fb could mean facebook...

sethvargo (commented):
Looking good - let me know when you want a full review

cmd/server/assets/login/_loginscripts.html
```diff
@@ -82,6 +82,41 @@
...
 {{end}}
 
+{{define "login/pwd-validate"}}
+{{if .requirements.HasRequirements}}
+<p class="card-text ml-4">
+  <small class="form-text text-muted">
+    <span class="row">Password should be:</span>
+    {{if gt .requirements.Length 0}}
+    <span class="row ml-1" id="length-req">
+      <span id="icon"></span>At least {{.requirements.Length}} characters long
```
sethvargo, line 92: These should have `aria-hidden="true"` for screenreaders

internal/firebase/error.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	InvalidOOBCode   = "INVALID_OOB_CODE"
+	CredentialTooOld = "CREDENTIAL_TOO_OLD_LOGIN_AGAIN"
+	TokenExpired     = "TOKEN_EXPIRED"
+	InvalidToken     = "INVALID_ID_TOKEN"
+)
+
+// ErrorDetails is the structure firebase gives back.
+type ErrorDetails struct {
```
sethvargo, line 26: This is pretty similar to the other PR (and I have the same feedback about making this an error and just returning a typed error(

pkg/controller/login/select_password.go
```diff
@@ -51,18 +59,72 @@ func (c *Controller) HandleSubmitNewPassword() http.Handler {
 			return
 		}
 
+		m := controller.TemplateMapFromContext(ctx)
+		f := flash.New(nil)
+
+		if err := c.validateComplexity(form.Password); err != nil {
+			logger.Errorw("validateComplexity failed", "error", err)
```
sethvargo, line 66: Anything that's error or errorw will eventually show up in an alert. We should only be using error when it's a _server_ error (i.e. something we should be paged for).

Since we're returning this error to the user in the flash, it seems like we shouldn't log it at all.

pkg/controller/login/select_password.go
```diff
@@ -51,18 +59,72 @@ func (c *Controller) HandleSubmitNewPassword() http.Handler {
...
+			f.Error("Select password failed. %v", err)
+			c.renderShowSelectPassword(ctx, w)
+			return
+		}
+
+		if _, err := c.firebaseInternal.VerifyPasswordResetCode(ctx, form.Code, form.Password); err != nil {
+			logger.Errorw("VerifyPasswordResetCode failed", "error", err)
```
sethvargo, line 73: Same note on logging; probably want to put this in a flash

sethvargo (changes requested):

pkg/controller/login/select_password.go
```diff
@@ -26,43 +29,98 @@ import (
...
-		m["requirements"] = &c.config.PasswordRequirements
-		c.h.RenderHTML(w, "login/select-password", m)
+		c.renderShowSelectPassword(ctx, w, nil)
 	})
 }
 
+func (c *Controller) renderShowSelectPassword(ctx context.Context, w http.ResponseWriter, f *flash.Flash) {
+	m := controller.TemplateMapFromContext(ctx)
+	m["flash"] = f
+	m["firebase"] = c.config.Firebase
```
sethvargo, line 39: Why do we need firebase on the html template now?

sethvargo (commented):
looking good - let me know when you want a full l-g-t-m

whaught:
/hold

still testing that oob code verification works on the server

whaught:
/hold cancel

Should all be working

cmd/server/assets/_flash.html
```diff
@@ -1,5 +1,6 @@
 {{define "flash"}}
 <div id="alerts-container" style="position:absolute; top:0; right:1rem; z-index:2000;">
+  {{- if .flash}}
```
sethvargo, line 3: Why do we have to guard this?

internal/firebase/error.go
```diff
@@ -36,6 +39,13 @@ func (err *ErrorDetails) Error() string {
 	return err.Err
 }
 
+func (err *ErrorDetails) Is(target error) bool {
```
sethvargo, line 42: I don't think you should need this? It also doesn't guarantee that the error is an ErrorDetails because it could be wrapped.

internal/firebase/verify_password_reset_code.go
```diff
@@ -34,38 +34,56 @@ type verifyPasswordResetCodeRequest struct {
...
+			d := m["error"]
+			return "", &d
 		}
-		return fmt.Errorf("failure %d: %s", status, string(b))
+
+		return "", fmt.Errorf("failure %d: %s", status, string(b))
+	}
+
+	var m map[string]string
+	if err := json.Unmarshal(b, &m); err == nil {
```
sethvargo, line 84: Do we need to handle the err case here?

pkg/controller/login/select_password.go
```diff
@@ -16,53 +16,140 @@
...
 	type FormData struct {
-		Email string `form:"email"`
+		Password string `form:"password"`
+		Email    string `form:"email"`
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
+		session := controller.SessionFromContext(ctx)
+		f := flash.New(session.Values)
```
sethvargo, line 79: There's a controller helper for this. I think it's FlashFromSession. Also we should name this `flash` to be consistent.

pkg/controller/login/select_password.go
```diff
@@ -16,53 +16,140 @@
...
+
+		code := r.FormValue("oobCode")
 
 		var form FormData
 		if err := controller.BindForm(w, r, &form); err != nil {
-			logger.Errorw("failed to bind form", "error", err)
-			controller.InternalError(w, r, c.h, err)
+			f.Error("Select password failed. %v", err)
```
sethvargo, line 85:
```suggestion
			f.Error("Select password failed: %v", err)
```

pkg/controller/login/select_password.go
```diff
@@ -16,53 +16,140 @@
...
+			c.renderShowSelectPassword(ctx, w, "", code, f)
+			return
+		}
+
+		if err := c.validateComplexity(form.Password); err != nil {
+			f.Error("Select password failed. %v", err)
```
sethvargo, line 91:
```suggestion
			f.Error("Select password failed: %v", err)
```

internal/firebase/error.go
whaught: I have the static ones up top, but they don't match unless it's a complete match and I also specify the http code for them. I was hoping to match only on this error enum firebase gives back

internal/firebase/verify_password_reset_code.go
whaught: I think we still want this to be understood as success if we fail to parse the body

PR #660, opened 2020-09-23
Give gorm the column stats so previous auto-migrates don't fail

sethvargo:
**Release Note**

```release-note
NONE
```


icco (approved):
/lgtm

PR #661, opened 2020-09-23
Make sure password change time is something

whaught:
## Proposed Changes

* Password change will be the account creation time if the new column is unset
* "unknown" string for the UI just in case, but we don't expect to see that

**Release Note**

```release-note
Password creation time field falls back to user creation time
```


sethvargo (approved):
/lgtm

PR #666, opened 2020-09-24
Ensure provider type is always set

sethvargo:
**Release Note**

```release-note
Ensure SMS provider type is set on system configs
```


whaught (approved):
/lgtm

PR #668, opened 2020-09-24
Server-side password reset

whaught:
## Proposed Changes

* Moves the password-reset logic to the server-side
    * Allows us to better hide the not-found response if the user email isn't in our system

**Release Note**

```release-note
Move password reset call to the server
```

### DIFF
--- a/internal/firebase/error.go
+++ b/internal/firebase/error.go
@@ -17,33 +17,25 @@ package firebase
 import "errors"
 
 var (
-	EmailNotFound    = &ErrorDetails{Err: "EMAIL_NOT_FOUND"}
-	InvalidOOBCode   = &ErrorDetails{Err: "INVALID_OOB_CODE"}
-	CredentialTooOld = &ErrorDetails{Err: "CREDENTIAL_TOO_OLD_LOGIN_AGAIN"}
-	TokenExpired     = &ErrorDetails{Err: "TOKEN_EXPIRED"}
-	InvalidToken     = &ErrorDetails{Err: "INVALID_ID_TOKEN"}
+	EmailNotFound    = errors.New("EMAIL_NOT_FOUND")
+	InvalidOOBCode   = errors.New("INVALID_OOB_CODE")
+	CredentialTooOld = errors.New("CREDENTIAL_TOO_OLD_LOGIN_AGAIN")
+	TokenExpired     = errors.New("TOKEN_EXPIRED")
+	InvalidToken     = errors.New("INVALID_ID_TOKEN")
 )
 
 // ErrorDetails is the structure firebase gives back.
 type ErrorDetails struct {
 	ErrorCode int    `json:"code"`
 	Err       string `json:"message"`
-	Message   string
 }
 
-func (err *ErrorDetails) Error() string {
-	return err.Message
-}
-
-func (err *ErrorDetails) Is(target error) bool {
-	if t, ok := target.(*ErrorDetails); ok {
-		return err.Err == t.Err
-	}
-	return err.Message == target.Error()
+func (err ErrorDetails) Error() string {
+	return err.Err
 }
 
 // ShouldReauthenticate returns true for errors that require a refreshed auth token.
-func (err *ErrorDetails) ShouldReauthenticate() bool {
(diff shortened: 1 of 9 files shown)


pkg/controller/login/reset_password.go
```diff
@@ -16,17 +16,60 @@
...
+			return
+		}
+
+		// There's no session yet, so make a one-time flash.
 		m := controller.TemplateMapFromContext(ctx)
-		m["firebase"] = c.config.Firebase
-		c.h.RenderHTML(w, "login/reset-password", m)
+		f := flash.New(nil)
+
+		if details, err := c.firebaseInternal.SendPasswordResetEmail(ctx, form.Email); err != nil {
```
sethvargo, line 60: See my comment above, but I think this API should be:

```go
if err := c.firebaseInternal.SendPasswordResetEmail(ctx, form.Email); err != nil {
  if errors.Is(err, firebase.EmailNotFound) {
    // special case
  }
}
```

pkg/controller/login/reset_password.go
```diff
@@ -16,17 +16,60 @@
...
+			// Treat not-found like success so we don't leak details.
+			if details.Error != firebase.EmailNotFound {
+				logger.Errorw("SendPasswordResetEmail failed", "error", err)
+				f.Error("reset password failed. %v", err)
+				c.renderResetPassword(ctx, w)
+				return
+			}
```
sethvargo, line 67: Need to handle the case where the error is not `EmailNotFound`

sethvargo (changes requested):

pkg/controller/login/reset_password.go
whaught: We do!

whaught: Or rather we purposely skip over it and flash "Password reset email sent." so that they all look the same

internal/firebase/error.go
```diff
@@ -0,0 +1,42 @@
...
+	InvalidToken     = errors.New("INVALID_ID_TOKEN")
+)
+
+// ErrorDetails is the structure firebase gives back.
+type ErrorDetails struct {
+	ErrorCode int    `json:"code"`
+	Err       string `json:"message"`
+}
+
+func (err ErrorDetails) Error() string {
```
sethvargo, line 33: Nit (can be followup): document this as implementing the error interface and add a compile time check:

```golang
var _ error = (nil)(*ErrorDetails)
```

pkg/controller/login/reset_password.go
```diff
@@ -16,17 +16,61 @@
...
+		if err := controller.BindForm(w, r, &form); err != nil {
+			f.Error("Password failed. %v", err)
+			c.renderResetPassword(ctx, w)
+			return
+		}
+
+		if err := c.firebaseInternal.SendPasswordResetEmail(ctx, form.Email); err != nil {
+			// Treat not-found like success so we don't leak details.
+			if !errors.Is(err, firebase.EmailNotFound) {
+				logger.Errorw("SendPasswordResetEmail failed", "error", err)
```
sethvargo, line 64: I still think we want to drop this log

internal/firebase/error.go
```diff
@@ -0,0 +1,42 @@
...
+	ErrInvalidToken     = errors.New("INVALID_ID_TOKEN")
+)
+
+// ErrorDetails is the structure firebase gives back.
+type ErrorDetails struct {
+	ErrorCode int    `json:"code"`
+	Err       string `json:"message"`
+}
+
+func (err ErrorDetails) Error() string {
```
sethvargo, line 33: This should be on the pointer `*ErrorDetails`

internal/firebase/error.go
```diff
@@ -0,0 +1,42 @@
...
+	return err.Err
+}
+
+// ShouldReauthenticate returns true for errors that require a refreshed auth token.
+func (err ErrorDetails) ShouldReauthenticate() bool {
```
sethvargo, line 38: Pointer receiver

whaught:
/unhold

tested that it works

sethvargo (approved):
/lgtm

PR #672, opened 2020-09-25
User search by email

whaught:
## Proposed Changes

* Allow for filtering the Users page by email address

### DIFF
--- a/cmd/server/assets/users/index.html
+++ b/cmd/server/assets/users/index.html
@@ -4,6 +4,7 @@
 {{$currentUser := .currentUser}}
 {{$currentRealm := .currentRealm}}
 {{$users := .users}}
+{{$s := .search}}
 
 <!doctype html>
 <html lang="en">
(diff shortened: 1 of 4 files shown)


cmd/server/assets/users/index.html
```diff
@@ -84,17 +86,31 @@ <h1>Users</h1>
           {{$pages := .pages}}
           <ul class="pagination">
             <li class="page-item {{if eq $pages.Previous -1}}disabled{{end}}"><a class="page-link"
-                href="?offset={{$pages.Previous}}">Previous</a>
+                href="?{{if $s}}search={{$s}}&{{end}}offset={{$pages.Previous}}">Previous</a>
```
icco, line 89: I'd argue there's no reason for this if statement. Throughout this file you check if $s exists, but given string's default empty state is empty string, there is no reason for these if statements.

sethvargo: Alternatively, you can build this once at the top where you define `$s` as `$searchQuery` or something.

cmd/server/main.go
```diff
@@ -339,7 +339,8 @@ func realMain(ctx context.Context) error {
 
 		userController := user.New(ctx, firebaseInternal, auth, cacher, config, db, h)
 		userSub.Handle("", userController.HandleIndex()).Methods("GET")
-		userSub.Handle("", userController.HandleIndex()).Queries("offset", "{[0-9]*?}").Methods("GET")
+		userSub.Handle("", userController.HandleIndex()).
+			Queries("offset", "{[0-9]*?}").Queries("email", "").Methods("GET")
```
sethvargo, line 343: The `?` here is redundant since `*` is "0 or more"

pkg/database/realm.go
```diff
@@ -585,15 +585,20 @@ func (r *Realm) CountUsers(db *Database) (int, error) {
...
-func (r *Realm) ListUsers(db *Database, offset, limit int) ([]*User, error) {
+func (r *Realm) ListUsers(db *Database, offset, limit int, emailPrefix string) ([]*User, error) {
 	if limit > MaxPageSize {
 		limit = MaxPageSize
 	}
 
+	realmDB := db.db.Model(r)
+
+	if emailPrefix != "" {
+		realmDB = realmDB.Where("email like ?", fmt.Sprintf("%s%%", emailPrefix))
```
sethvargo, line 596: Put the `%` in the query

sethvargo, line 596:
```suggestion
		realmDB = realmDB.Where("email like %?%", fmt.Sprintf("%s", emailPrefix))
```

I also think we want to allow searching at both ends.

cmd/server/assets/users/index.html
whaught: I like @sethvargo 's idea to remove the annoying redundancy.
This is to avoid `users/search=&offset=25` situations with an empty, useless query param

pkg/database/realm.go
whaught: We have to do it on the sprintf side, GORM puts quotes around ? and screws it up

whaught:
/unhold

should work

mikehelmick (approved):

PR #673, opened 2020-09-25
Better organize the project variables in the alerting TF module.

yegle:
The end goal is to have 3 variables
- monitoring-host-project: preferrably a separate project to host alerts
and dashboards.
- verification-server-project: Optional if empty
- key-server-project: Optional if empty

If the agency running the service only care about one of
verification/key server, they can choose to leave out the one they don't
care about.


icco (commented):
Please update the readme with these changes.

PR #679, opened 2020-09-25
Direct users back to password reset email if code invalid/expired

whaught:
## Proposed Changes

* This is based on some user feedback / confusing over expired invite links. I'm hoping this makes it clear what to do to make a login.
* Disable fields and link to password reset if there is an invalid code
    * We could just redirect there and skip over this with a flash message, but I find that jarring
* Wrap flash with if in case it's not set. That can be the case pre-login where there is no session/middleware setting flash to empty

### DIFF
--- a/cmd/server/assets/_flash.html
+++ b/cmd/server/assets/_flash.html
@@ -1,5 +1,6 @@
 {{define "flash"}}
 <div id="alerts-container" style="position:absolute; top:0; right:1rem; z-index:2000;">
+  {{- if .flash}}
   {{range $msg := .flash.Errors}}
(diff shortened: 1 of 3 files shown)


mikehelmick (approved):

cmd/server/assets/_flash.html
```diff
@@ -1,5 +1,6 @@
 {{define "flash"}}
 <div id="alerts-container" style="position:absolute; top:0; right:1rem; z-index:2000;">
+  {{- if .flash}}
```
sethvargo, line 3: I'm gonna regret teaching you about dashes, aren't I? 😄

whaught: mystery whitespace is bs

PR #682, opened 2020-09-25
Redirect firebase emails on 'mode' query param

whaught:
## Proposed Changes

* Redirect to verify email / select password based on the firebase mode

I like this less, but firebase doesn't allow for custom URL per email template. I think it's pretty geared for their UI component.

Note: We'll switch the template to manage-account and remove the old routing after deployment


sethvargo (approved):
/lgtm

PR #685, opened 2020-09-25
Recreate firebase user on password reset

whaught:
## Proposed Changes

* When resetting password, ensure that if we have a db.User that we also have a firebase user
* Move password-reset to server-side for Users/ page

**Release Note**

```release-note
Ensure that password-reset flows also check that the firebase user auth is created if it's missing and we have an entry for the user.
```


pkg/controller/login/reset_password.go
```diff
@@ -55,6 +55,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			return
 		}
 
+		// Ensure that if we have a user, they have auth
+		if user, err := c.db.FindUserByEmail(form.Email); err == nil {
+			user.CreateFirebaseUser(ctx, c.client)
```
mikehelmick, line 60: can you add a metric for this?

realistically this should never happen (but we're seeing it)

whaught: Metric added. I haven't used OpenCensus before - do these need to be declared or anything?

cmd/server/main.go
```diff
@@ -201,6 +201,12 @@ func realMain(ctx context.Context) error {
 	processFirewall := middleware.ProcessFirewall(ctx, h, "server")
 	rateLimit := httplimiter.Handle
 
+	// Shared metrics
+	userMetrics, err := user.RegisterMetrics()
```
mikehelmick, line 205: the pattern we've been using is to register these in the controller

https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/controller/issueapi/issueapi.go#L48

I don't think there's a reason to pipe them back out through here.

I think having a shared metrics register w/ a mutex to only initialize once and keep it out of main would be better

whaught: idk about mutex, but how about  putting it on context

pkg/controller/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+
+const contextKeyMetrics = contextKey("controllerMetricsKey")
+
+var MetricPrefix = observability.MetricRoot + "/server/shared"
+
+type Metrics struct {
+	FirebaseRecreates *stats.Int64Measure
+}
+
+func MetricsFromContext(ctx context.Context) (context.Context, *Metrics, error) {
```
mikehelmick, line 36: i don't love the dependency through context - it now implies an ordering between controller creation.

I still think there are other solutions that don't require that ordering / dependency

whaught: What do you mean by ordering? This is lazily initialized as a singleton on context, if the controller can't find it on context, it gets created. Controllers can be organized in any order (the first one will create)

mikehelmick: it requires that everyone who calls it shares a context though, otherwise it will be created twice.

that may be OK in opencensus - I'm not positive

sethvargo: I think OC actually requires you to use the ctx - all the data is stored on the ctx

whaught:
/close

I'm going to reconsider this. Not sure if we want to do this on the external login due to rate-limiting concerns

cmd/server/main.go
```diff
@@ -219,62 +219,64 @@ func realMain(ctx context.Context) error {
 		sub.Handle("/health", controller.HandleHealthz(ctx, &cfg.Database, h)).Methods("GET")
 	}
 
+	ctx, loginController, err := login.New(ctx, firebaseInternal, auth, cfg, db, h)
```
sethvargo, line 222: Why is this diff so large?

pkg/controller/metrics.go
sethvargo, line 36: I'm gonna defer to @yegle and @icco here on the best practice

whaught: I've removed all that and shoved it into the database per our conversation. I think sharing context makes sense but @mikehelmick prefers the tighter scoping in main.go

whaught:
/close

This has been replaced

PR #686, opened 2020-09-25
fix iso code validation

mikehelmick:
## Proposed Changes

* previous validation didn't all top level, 2 character, ISO regions.

**Release Note**

```release-note
Allow country level ISO regions for EN Express
```


sethvargo (approved):
/lgrm

PR #689, opened 2020-09-26
Use a dynamic block to simplify the code.

yegle:
Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```

### DIFF
--- a/terraform/verification-lb.tf
+++ b/terraform/verification-lb.tf
@@ -15,9 +15,9 @@
 locals {
   enable_lb = var.server-host != "" && var.apiserver-host != "" && var.adminapi-host != ""
   lb_mapping = {
-    "server": var.server-host,
-    "apiserver": var.apiserver-host,
-    "adminapi": var.adminapi-host,
+    "server" : var.server-host,
+    "apiserver" : var.apiserver-host,
+    "adminapi" : var.adminapi-host,
   }
 }
 
@@ -49,16 +49,16 @@ resource "google_compute_url_map" "urlmap-https" {
 
   dynamic "host_rule" {
     for_each = local.lb_mapping
-    content = {
-      hosts = [host_rule.value]
+    content {
+      hosts        = [host_rule.value]
       path_matcher = host_rule.key
     }
   }
   dynamic "path_matcher" {
     for_each = local.lb_mapping
-    content = {
-      name = path_matcher.key
-      default_service = google_compute_backend_service.get(path_matcher.key)[0].id
+    content {
+      name            = path_matcher.key
+      default_service = element(google_compute_backend_service, path_matcher.key)[0].id
     }
   }
 }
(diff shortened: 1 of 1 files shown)


terraform/verification-lb.tf
```diff
@@ -41,34 +46,20 @@ resource "google_compute_url_map" "urlmap-https" {
...
-  }
-
-  path_matcher {
-    name            = "adminapi"
-    default_service = google_compute_backend_service.adminapi[0].id
+  dynamic "path_matcher" {
+    for_each = local.lb_mapping
+    content = {
+      name = path_matcher.key
+      default_service = google_compute_backend_service.get(path_matcher.key)[0].id
```
yegle, line 61: @sethvargo This doesn't seem to work.

sethvargo: Is it an error or?

yegle: This is what I got:

```
Error: Missing attribute separator

  on ../terraform/verification-lb.tf line 61, in resource "google_compute_url_map" "urlmap-https":
  59:     content = {
  60:       name = path_matcher.key
  61:       default_service = google_compute_backend_service.get(path_matcher.key)[0].id

Expected a newline or comma to mark the beginning of the next attribute.
```

sethvargo: I don't think `.get` is a thing. What is `google_compute_backend_service`? You might want:

```
element(google_compute_backend_service, path_matcher.key).id
```

yegle: `google_compute_backend_service` is the resource type.

`element()` seems to be the right function to use, but it looks like I can't have dynamic reference to resources. Error I got with the latest commit:

```
Error: Invalid reference

  on ../terraform/verification-lb.tf line 61, in resource "google_compute_url_map" "urlmap-https":
  61:       default_service = element(google_compute_backend_service, path_matcher.key)[0].id

A reference to a resource type must be followed by at least one attribute
access, specifying the resource name.
```

sethvargo: I think you drop the `[0]`: 

```
element(google_compute_backend_service, path_matcher.key).id
```

yegle: That `[0]` is required because the `google_compute_backend_service` is conditionally created using `count = local.enable_lb ? 1 : 0`. 

Regardless, without `[0]` this is what I get:

```
Error: Invalid reference

  on ../terraform/verification-lb.tf line 61, in resource "google_compute_url_map" "urlmap-https":
  61:       default_service = element(google_compute_backend_service, path_matcher.key).id

A reference to a resource type must be followed by at least one attribute
access, specifying the resource name.```

sethvargo: I see. You are trying to get `google_compute_backend_service.THING` where `THING` is a variable. I don't think that's actually possible in Terraform.

yegle: :-(

Thanks for looking into this!

PR #690, opened 2020-09-28
Make latency alert a little more specific

icco:
## Proposed Changes

* Improve wording of alert text
* Change to 99%tile
* Limit to just Verification GCLB (ie not the redirect server)
* Exclude requests that GCLB ignored, ie didn't have a backend.

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

PR #694, opened 2020-09-28
Ensure password validity updated on retype

whaught:
## Proposed Changes

* Password validity should get re-checked on retype because we now validate that they match

**Release Note**

```release-note
Minor fix to password selection validation UI
```


mikehelmick (approved):

icco (approved):
/lgtm

PR #699, opened 2020-09-28
Make fixes to realm capacity alerts

icco:
## Proposed Changes

* For realm capacity alert, take max across all binaries per realm, instead of per realm and per binary.

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

Issue #697, opened 2020-09-28
Let system admins add themselves as a realm admin


mikehelmick:
### TL;DR

In order to facilitate support, allow a system admin to add them selves as a realm admin to an existing realm.

### Design

**Proposal**

This would be done from the current admin -> realms pages

/kind enhancement

PR #701, opened 2020-09-28
Make mobile app names unique by realm_id and platform

sethvargo:
Fixes GH-697

**Release Note**

```release-note
Make mobile app names unique by realm_id _and_ platform
```

### DIFF
--- a/pkg/database/mobile_app.go
+++ b/pkg/database/mobile_app.go
@@ -35,10 +35,10 @@ type MobileApp struct {
 	Errorable
 
 	// Name is the name of the app.
-	Name string `gorm:"column:name; type:citext; unique_index:realm_app_name;"`
+	Name string `gorm:"column:name; type:citext;"`
 
 	// RealmID is the id of the mobile app.
-	RealmID uint `gorm:"column:realm_id; unique_index:realm_app_name;"`
+	RealmID uint `gorm:"column:realm_id;"`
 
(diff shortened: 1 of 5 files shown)


pkg/database/mobile_app.go
```diff
@@ -72,6 +72,23 @@ func (a *MobileApp) BeforeSave(tx *gorm.DB) error {
 		a.AddError("os", "is invalid")
 	}
 
+	// Ensure name/os combo is unique - there's still a database constraint that
+	// enforces this in case there's a race.
+	var existing MobileApp
+	if err := tx.Unscoped().
+		Model(&MobileApp{}).
+		Where("name = ?", a.Name).
```
whaught, line 80: Nit:
`Where("name = ? AND realm_id = ? AND os = ?", a.name, a.RealmID, a.OS)`
might be more SQL-y

Is this redundant with the constraint?

sethvargo: It's more SQL-y, but less Gorm-y. /shrug

No - this doesn't strictly guarantee that no duplicates occur. I just asserts that no duplicates occur at the time of the check. The constraint is a stronger promise, but that error is much harder to catch and throw back at the user. Doing this allows us to give the average user a nice "this already exists" response while still keeping the consistency guarantee of the uniqueness index.

sethvargo:
@icco fixed the test failure

PR #704, opened 2020-09-28
remove attempts at country code validation and just require one is set

mikehelmick:
## Proposed Changes

* Go back to allowing any non-empty region code - the validation was breaking test environments.

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

Issue #698, opened 2020-09-28
Audit trail of user add / remove


mikehelmick:
Provide an audit trail of user account add/remove to a realm.

A realm admin should be able to when users were added to a realm.

The record of the user being in a realm can go away after 30d (cleanup period).

The cleanup jobs will need to be updated as well for this.

/kind enhancement

PR #705, opened 2020-09-29
Add auditing and system admin realm joining

sethvargo:
- Fixes GH-697
- Fixes GH-698

**Release Note**

```release-note
Add basic auditing and allow system admins to join realms for recovery purposes
```

### DIFF
--- a/cmd/server/assets/realmadmin/events.html
+++ b/cmd/server/assets/realmadmin/events.html
@@ -1,0 +1,64 @@
+{{define "realmadmin/events"}}
+
+{{$realm := .realm}}
+{{$events := .events}}
+
+<!doctype html>
+<html lang="en">
+<head>
+  {{template "head" .}}
+</head>
+
+<body class="tab-content">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Realm event log</h1>
+    <p>
+      The list below shows the past 30 days of events that have occurred on this
+      realm. Not all events are recorded for auditing to preserve privacy.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Events</div>
+      <div class="list-group list-group-flush">
+        {{range $event := $events}}
+          <div class="list-group-item flex-column align-items-start">
+            <div class="d-flex w-100 justify-content-between">
+              <h5 class="mb-1">{{$event.Action}}</h5>
+              <small
+                data-timestamp="{{$event.CreatedAt.Format "1/02/2006 3:04:05 PM UTC"}}"
+                data-toggle="tooltip" title="{{$event.CreatedAt.Format "2006-02-01 15:04 UTC"}}">
+                {{$event.CreatedAt.Format "2006-02-01 15:04"}}
+              </small>
+            </div>
+            <div>
+              <span class="text-primary text-nowrap">{{$event.ActorDisplay}}</span>
+
+              <span>{{$event.Action}}</span>
+
+              <span class="text-primary text-nowrap">{{$event.TargetDisplay}}</span>
+
+              {{if $event.Diff}}
+              <br>
+                <a href="#" data-toggle="collapse" data-target="#collapseDiff{{$event.ID}}"
+                  aria-expanded="true" aria-controls="collapseDiff{{$event.ID}}"
+                  class="small text-muted">
+                  Toggle diff
+                </a>
+                <pre id="collapseDiff{{$event.ID}}" class="collapse mt-3 mb-1"><code>{{$event.Diff}}</code></pre>
+              {{end}}
+            </div>
+          </div>
+        {{end}}
+      </div>
+    </div>
+  </main>
+
+  {{template "scripts" .}}
+
+</body>
+</html>
+{{end}}
--- a/cmd/server/assets/users/index.html
+++ b/cmd/server/assets/users/index.html
@@ -72,13 +72,6 @@
                   title="Remove this user">
                   <span class="oi oi-trash" aria-hidden="true"></span>
                 </a>
-                {{else if $currentUser.Admin}}
-                {{- /* system admins can remove themselves */ -}}
-                <a href="/users/{{.ID}}" class="d-block text-danger" data-method="DELETE"
-                  data-confirm="Are you sure you want to leave {{.Name}}?" data-toggle="tooltip"
-                  title="Leave this realm">
-                  <span class="oi oi-account-logout" aria-hidden="true"></span>
-                </a>
                 {{end}}
               </td>
             </tr>
--- a/pkg/controller/admin/realms.go
+++ b/pkg/controller/admin/realms.go
@@ -59,8 +59,8 @@ func (c *Controller) HandleRealmsCreate() http.Handler {
 		}
 		flash := controller.Flash(session)
 
-		user := controller.UserFromContext(ctx)
-		if user == nil {
+		currentUser := controller.UserFromContext(ctx)
+		if currentUser == nil {
 			controller.MissingUser(w, r, c.h)
 			return
 		}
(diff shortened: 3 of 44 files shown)


cmd/server/assets/users/index.html
```diff
@@ -72,13 +72,6 @@ <h1>Users</h1>
                   title="Remove this user">
                   <span class="oi oi-trash" aria-hidden="true"></span>
                 </a>
-                {{else if $currentUser.Admin}}
```
sethvargo, line 75: I removed this because it's available in the system admin console now, which feels like a better place.

pkg/controller/admin/realms.go
```diff
@@ -199,3 +201,130 @@ func (c *Controller) renderEditRealm(ctx context.Context, w http.ResponseWriter,
...
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		user.Realms = append(user.Realms, realm)
+		user.AdminRealms = append(user.AdminRealms, realm)
+
+		// Do the membership update and audit entry in a transaction because we need
+		// both to succeed to continue.
+		if err := c.db.RawDB().Transaction(func(tx *gorm.DB) error {
```
sethvargo, line 234: I'm really not a fan of this, but I can't seem to find a better way. I thought about using callbacks, but there's far too many fields to set on the auditable resource.

I'm was thinking it might be better to allow auditable resource to have a *AuditEntry field that can be nil. If it's populated, have the callback create the entry. I just... couldn't get it to work with gorm.

cmd/server/assets/realmadmin/events.html
```diff
@@ -0,0 +1,62 @@
...
+<body class="tab-content">
+  {{template "navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>Realm event log</h1>
+    <p>
+      The list below shows the past 30 days of events that have occurred on this
+      realm. Not all events are audited to preserve privacy.
```
whaught, line 21:
```suggestion
      realm. Not all events are recorded for auditing to preserve privacy.
```

cmd/server/assets/users/index.html
whaught: You mean the My Account page? I think its there too. Regardless I like removing it here.

pkg/controller/realmadmin/events.go
```diff
@@ -0,0 +1,50 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		realm := controller.RealmFromContext(ctx)
+		if realm == nil {
+			controller.MissingRealm(w, r, c.h)
+			return
+		}
+
+		events, err := realm.Audits(c.db)
```
sethvargo, line 35: Should we cache these? I'm hesitant given it's for forensics, but maybe a 30s cache would help?

pkg/database/realm.go
```diff
@@ -694,8 +730,207 @@ func (db *Database) GetRealms() ([]*Realm, error) {
...
+				audits = append(audits, audit)
+			}
+
+			if existing.PasswordRotationWarningDays != r.PasswordRotationWarningDays {
+				audit := BuildAuditEntry(actor, "updated password rotation warning", r, r.ID)
+				audit.Diff = uintDiff(existing.PasswordRotationWarningDays, r.PasswordRotationWarningDays)
+				audits = append(audits, audit)
+			}
+
+			// TODO(sethvargo): diff allowed CIDRs
```
sethvargo, line 862: Gonna do this later.

sethvargo:
@whaught @mikehelmick okay - this is ready for review now after a few iterations and virtual rubber-ducking.

High-level overview:

- There's a new interface in the database package, `Auditable`, that determines if a resource can be audited. These resources can act as the "actor" (entity that did the action) or "target" (entity upon which the action was taken).

- Audits are optionally tied to a realm, although some audits are not. There's currently no UI to display non-realm-specific audits, but they are captured and we can always add that later.

- Audits are automatically cleaned up after 30d (default).

- Audits intentionally don't use foreign keys so that they remain intact even if the resourcing record is deleted.

- Audited resources require an auditable actor when calling their `Save` function. There's a `database.System` user for tests/when the actor is the system (e.g. seeding, model calculation).

- Audited resources are currently manually diffed. We already have enough runtime reflection with gorm, and we want to have tight control over which fields are diffed and which order they are diffed. This means the `Save` functions are verbose, but I think it's worth the tradeoff.

- Audit entries cannot be deleted (unless you do it directly from the database).

- With this, system admins can now join and leave realms.

pkg/controller/login/select.go
```diff
@@ -43,8 +43,16 @@ func (c *Controller) HandleSelectRealm() http.Handler {
 			return
 		}
 
-		// If the user is only a member of one realm, set that and bypass selection.
-		if len(user.Realms) == 1 {
+		switch len(user.Realms) {
+		case 0:
```
mikehelmick, line 47: good idea!

pkg/controller/realmadmin/events.go
mikehelmick: I don't think it's necessary. it's locked to realm admins and there is a UI rate limit.

pkg/database/audit_entry.go
```diff
@@ -0,0 +1,77 @@
...
+
+// AuditEntry represents an event in the system. These records are purged after
+// a configurable number of days by the cleanup job. The AuditEntry specifically
+// does NOT make use of foreign keys or relationships to avoid breaking an audit
+// entry if the upstream data which was audited is removed or changed. These
+// records should be considered immutable.
+type AuditEntry struct {
+	Errorable
+
+	// ID is the entry's ID.
```
mikehelmick, line 29: any reason why not the gorm.Model embed?

pkg/database/audit_entry.go
```diff
@@ -0,0 +1,77 @@
...
+	Diff string `gorm:"column:diff; type:text;"`
+
+	// CreatedAt is when the entry was created.
+	CreatedAt time.Time
+}
+
+// PurgeAuditEntries will delete audit entries which were created longer than
+// maxAge ago.
+func (db *Database) PurgeAuditEntries(maxAge time.Duration) (int64, error) {
+	if maxAge > 0 {
```
mikehelmick, line 67: can we code a safety here, or in the config that prevents lowing this below a reasonable number, 14d, 30d?

pkg/database/migrations.go
```diff
@@ -1367,6 +1367,31 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+		{
+			ID: "00055-AddAuditEntries",
+			Migrate: func(tx *gorm.DB) error {
+				if err := tx.AutoMigrate(&AuditEntry{}).Error; err != nil {
+					return err
+				}
+
+				sqls := []string{
+					`CREATE INDEX IF NOT EXISTS idx_audit_entries_realm_id ON audit_entries (realm_id)`,
+					`CREATE INDEX IF NOT EXISTS idx_audit_entries_actor_id ON audit_entries (realm_id)`,
```
mikehelmick, line 1379: I think this is on the wrong column

mikehelmick (changes requested):

pkg/database/migrations.go
mikehelmick: this is the reason I marked requested chagnes.

pkg/database/audit_entry.go
```diff
@@ -0,0 +1,77 @@
...
+		maxAge = -1 * maxAge
+	}
+	createdBefore := time.Now().UTC().Add(maxAge)
+
+	result := db.db.
+		Unscoped().
+		Where("created_at < ?", createdBefore).
+		Delete(&AuditEntry{})
+	return result.RowsAffected, result.Error
```
icco, line 76: Could we add a metric, audit_rows_cleaned or something here?

pkg/database/auditable.go
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+// Auditable represents a resource that can be audited as an actor or actee.
+type Auditable interface {
```
icco, line 18: My knowledge of gorm is limited. Is there anyway to do a callback here so that we can increment a counter on every audit entry inserted? If there is, can we do it?

icco (commented):
I'd like two metrics, they can be saved for future work if you'd like.

pkg/database/auditable.go
sethvargo: We have access to the tx in the callback, so we can increment a SQL counter (that's how we do the in-app stats actually). 

We can create a global callback that has access to metrics and then basically guard it with a `if table == "audit_entries"`. What do you hope to glean from counting the number of audits?

pkg/database/audit_entry.go
sethvargo: `gorm.Model` changes the behavior and semantics for delete. If the resource has `DeletedAt`, it soft deletes. Audit entries are also immutable and `gorm.Model` includes an unnecessary `UpdatedAt` field. Given I expect audit_entries to be our largest table (even with purging), I wanted to keep the number of columns as small as possible.

pkg/database/audit_entry.go
sethvargo: Done - added a 7d min it in the config parsing.

pkg/database/migrations.go
sethvargo: Ah good catch. I also didn't add it for target_id. Fixed.

pkg/database/audit_entry.go
icco: Ack, ty.

pkg/database/auditable.go
icco: I would love a graph in our dashboards to see how many things are happening on a server from an admin point of view. Basically a stackdriver stat that says x number of actions lets us see the rate of actions happening on an instance. Good for debugging or tracking admin abuse.

PR #714, opened 2020-09-29
Add integration tests for verification server

p3y1c2n:
## Proposed Changes

* Add integration tests for verification server
    * Add an integration test config, some helper functions and test clients
    * Test valid token
    * Test expired token


pkg/config/integration_config.go
```diff
@@ -0,0 +1,134 @@
...
+		KeyManagerType: "FILESYSTEM",
+		FilesystemRoot: tmpdir,
+	}
+
+	kms, err := keys.KeyManagerFor(ctx, &keyConfig)
+	if err != nil {
+		tb.Fatal(err)
+	}
+	tb.Cleanup(func() {
+		if err := os.RemoveAll(tmpdir); err != nil {
```
chaodaiG, line 76: Shouldn't this be best effort and not failing test if it failed?

chaodaiG (commented):
Looks good to me, left a nit.

/cc @sethvargo @mikehelmick

pkg/config/integration_config.go
sethvargo: I think this matches what we do elsewhere.

pkg/config/integration_config.go
```diff
@@ -0,0 +1,134 @@
...
+			tb.Fatal(err)
+		}
+	})
+
+	parent := keys.TestSigningKey(tb, kms)
+	keyID, err := kms.(keys.SigningKeyManager).CreateKeyVersion(ctx, parent)
```
sethvargo, line 82: While unlikely, this will panic if you change the key manager and it doesn't implement that interface.

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+// NewTestSuite creates a Suite for integration tests.
+func NewTestSuite(tb testing.TB, ctx context.Context) *Suite {
+	tb.Helper()
+	cfg, db := config.NewIntegrationTestConfig(ctx, tb)
+
+	// Create or reuse the existing realm
+	realm, err := db.FindRealmByName(realmName)
+	if err != nil {
+		if !database.IsNotFound(err) {
+			tb.Errorf("error when finding the realm %q: %w", realmName, err)
```
sethvargo, line 76: This should be `Fatal` I think

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+		realm = database.NewRealmWithDefaults(realmName)
+		realm.RegionCode = realmRegionCode
+		if err := db.SaveRealm(realm); err != nil {
+			tb.Errorf("failed to create realm %+v: %w: %v", realm, err, realm.ErrorMessages())
```
sethvargo, line 81: This should be fatal

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		tb.Errorf("failed to create suffix string for API keys: %w", err)
+	}
+
+	adminKey, err := realm.CreateAuthorizedApp(db, &database.AuthorizedApp{
+		Name:       adminKeyName + suffix,
+		APIKeyType: database.APIUserTypeAdmin,
+	})
+	if err != nil {
+		tb.Errorf("error trying to create a new Admin API Key: %w", err)
```
sethvargo, line 96: I think all of these should be fatal. In general, use fatal when you don't want execution to continue.

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+	cacher, err := cache.CacherFor(ctx, &s.cfg.APISrvConfig.Cache, cache.MultiKeyFunc(
+		cache.HMACKeyFunc(sha1.New, s.cfg.APISrvConfig.Cache.HMACKey),
+		cache.PrefixKeyFunc("apiserver:cache:"),
+	))
+	if err != nil {
+		tb.Fatalf("failed to create cacher: %v", err)
+	}
+	tb.Cleanup(func() {
+		if err := cacher.Close(); err != nil {
+			tb.Errorf("got err when cleanup: %v", err)
```
sethvargo, line 156:
```suggestion
			tb.Fatalf("failed to close cacher: %v", err)
```

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	})
+
+	// Create LimitStore
+	limiterStore, err := ratelimit.RateLimiterFor(ctx, &s.cfg.AdminAPISrvConfig.RateLimit)
+	if err != nil {
+		tb.Errorf("failed to create the limit store %v", err)
```
sethvargo, line 163: fatal

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+	// Stop the server on cleanup
+	stopCtx, stop := context.WithCancel(ctx)
+	tb.Cleanup(stop)
+
+	go func() {
+		if err := srv.ServeHTTPHandler(stopCtx, handlers.CombinedLoggingHandler(os.Stdout, adminRouter)); err != nil {
+			tb.Error(err)
+		}
+	}()
+	return srv
```
sethvargo, line 203: It doesn't look like this server is ever gracefully stopped or terminated.

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func (s *Suite) newAPIServer(ctx context.Context, tb testing.TB) *server.Server {
+	// Create the renderer
+	h, err := render.New(ctx, "", s.cfg.APISrvConfig.DevMode)
+	if err != nil {
+		tb.Errorf("failed to create the renderer %v", err)
```
sethvargo, line 210: Same notes about fatal, throughout

pkg/integration/helpers.go
```diff
@@ -0,0 +1,343 @@
...
+	// Stop the server on cleanup
+	stopCtx, stop := context.WithCancel(ctx)
+	tb.Cleanup(stop)
+
+	go func() {
+		if err := srv.ServeHTTPHandler(stopCtx, handlers.CombinedLoggingHandler(os.Stdout, apiRouter)); err != nil {
+			tb.Error(err)
+		}
+	}()
+	return srv
```
sethvargo, line 284: This should be closed when done

pkg/integration/integration_test.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const (
+	oneDay         = 24 * time.Hour
+	intervalLength = 10 * time.Minute
+	maxInterval    = 144
+)
+
+func TestIntegration(t *testing.T) {
+	cases := []struct {
```
sethvargo, line 38: t.Parallel()

sethvargo (changes requested):

pkg/integration/helpers.go
p3y1c2n: I think the server is terminated when the provided context is cancelled. The corresponding `stop` function is registered to tb.Cleanup.

pkg/config/integration_config.go
p3y1c2n: Done. Check if the kms implements SigningKeyManager before proceed.

pkg/integration/helpers.go
p3y1c2n: Same here, the server should be closed by the cancelled context.

pkg/integration/helpers.go
p3y1c2n: Hey Seth, I just realized I can't resolve suggested edits myself. Could you please resolve this one and also approve the PR? Thanks!

chaodaiG:
@sethvargo , could you add a /approve on this PR please?

whaught (approved):
/approve

PR #715, opened 2020-09-30
Update session as MFA factors are removed

whaught:
## Proposed Changes

* Followup to https://github.com/google/exposure-notifications-verification-server/pull/695
* Also sets session as factors are removed. This will disallow the user to navigate on an MFA-required realm after their last factor is removed.


sethvargo (approved):
/lgtm

PR #721, opened 2020-09-30
Update APIKeyType to have a basis of 0

sethvargo:
Since Go's default value for ints is 0, there have been a number of edge
cases where we have to manually set or check the APIKeyType as -1 to
indicate a value of "unset".

This PR changes the APIKeyType to be an iota with a default value of 0.
That means existing database records are updated 0->1 and 1->2.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1393,6 +1393,30 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return tx.Exec(`DROP TABLE audit_entries`).Error
 			},
 		},
+		{
+			ID: "00056-AuthorzedAppsAPIKeyTypeBasis",
+			Migrate: func(tx *gorm.DB) error {
+				if err := tx.AutoMigrate(&AuditEntry{}).Error; err != nil {
+					return err
+				}
+
+				sqls := []string{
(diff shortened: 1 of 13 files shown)


pkg/database/migrations.go
```diff
@@ -1393,6 +1393,32 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+			ID: "00056-AuthorzedAppsAPIKeyTypeBasis",
+			Migrate: func(tx *gorm.DB) error {
+				if err := tx.AutoMigrate(&AuditEntry{}).Error; err != nil {
+					return err
+				}
+
+				sqls := []string{
+					`UPDATE authorized_apps ALTER COLUMN api_key_type DROP DEFAULT`,
+					`UPDATE authorized_apps ALTER COLUMN api_key_type SET NOT NULL`,
+					`UPDATE authorized_apps SET api_key_type = 2 WHERE api_key_type = 1`,
```
mikehelmick, line 1406: I'm opposed to changing the values of these live. I'm worry about outages or inconsistencies

sethvargo: Updated

mikehelmick (approved):

PR #722, opened 2020-09-30
Explicitly depend on database migrations and IAM during deployment

sethvargo:
**Release Note**

```release-note
Explicitly depend on database migrations and IAM during deployment
```

/assign @chaodaiG @icco

### DIFF
--- a/terraform/service_apiserver.tf
+++ b/terraform/service_apiserver.tf
@@ -149,8 +149,19 @@ resource "google_cloud_run_service" "apiserver" {
 
   depends_on = [
     google_project_service.services["run.googleapis.com"],
+
     google_secret_manager_secret_iam_member.apiserver-db,
+    google_kms_key_ring_iam_member.kms-signerverifier,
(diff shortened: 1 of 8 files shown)


terraform/service_apiserver.tf
```diff
@@ -149,8 +149,19 @@ resource "google_cloud_run_service" "apiserver" {
 
   depends_on = [
     google_project_service.services["run.googleapis.com"],
+
     google_secret_manager_secret_iam_member.apiserver-db,
+    google_kms_key_ring_iam_member.kms-signerverifier,
```
chaodaiG, line 154: Is it possible to define this collection somewhere and be referenced here?

sethvargo: That would certainly be nice, but no.

jeremyfaller: I am consistently disappointed in aspects of this tool. Either we're doing something wrong, or there's an opportunity here.

PR #724, opened 2020-09-30
Doc cleanup largely based on spelling/grammar check

greggles:
As I was reading through these I noticed some spelling issues at the end of the realm_guide. In training with some users, they mentioned it was confusing setting up 2fa due to the need to prefix the phone number with +1 and suggested adding a note about that to the docs. I also ran the 2 files through grammarly and accepted most of the suggestions. Some items felt like personal preference not worth changing.

## Proposed Changes

* Fix spelling issues in realm_guide
* Explain multiple 2fa phones are useful
* General small fixes

**Release Note**

NONE

### DIFF
--- a/docs/realm_guide.md
+++ b/docs/realm_guide.md
@@ -1,6 +1,6 @@
 # Realm admin guide
 
-This guide provides high level steps for realm administrators to follow.
+This guide provides high-level steps for realm administrators to follow.
 
 If you are not a realm administrator, you will not have access to these screens.
 
@@ -8,10 +8,10 @@ If you are not a realm administrator, you will not have access to these screens.
 
 ### Account protection
 
-We provide a base level of account protection measures that we urge you to share with your case workers that are issuing verification codes.
+We provide a base level of account protection measures that we urge you to share with your caseworkers that are issuing verification codes.
 
 * All user accounts must verify ownership of their email address before using the system.
-* Two factor authentication (2FA) is available, we strongly suggest you require your users to enroll in 2FA
+* Two-factor authentication (2FA) is available, we strongly suggest you require your users to enroll in 2FA
   using a mobile device under their sole control.
 * Users should not share logins to the verification system.
 * Users should only issue codes to people who have a verified COVID-19 diagnosis.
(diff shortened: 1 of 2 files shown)


docs/realm_guide.md
```diff
@@ -110,15 +109,15 @@ Click the "Create a new signing key version" button. This will _create_ but not
 
 ![api keys](images/admin/keys01.png "API key created")
 
-If successful, you will get a message indiciating the new key version that was created.
+If successful, you will get a message indicating the new key version that was created.
```
greggles, line 112: Here to the end of the file are some spelling fixes.

docs/user_guide.md
```diff
@@ -1,19 +1,19 @@
...
-Expect to receive an email for new account / password reset.
+Expect to receive an email for a new account / password reset.
 This will provide a link to set up your account password.
 
 ![new user](images/users/step02.png "Create a password")
 
 ### Second factor authentication
 
-On your next login, you will be given the option to enroll a second factor for authentication (SMS sent to your personal mobile phone). It is highly recommended to enroll the second factor.
+On your next login, you will be given the option to enroll a second factor for authentication (SMS sent to your mobile phone). It is highly recommended to enroll in a second factor. The phone number requires the international prefix (e.g. +1 for USA).
```
greggles, line 16: I'm not sure if it's appropriate to put in USA specific advice. The overall idea was suggested by a few admins who noted they were confused by this input.

greggles: On double-checking the form it looks like maybe the input widget was improved in a way that makes this change unneccessary. Happy to remove this if you agree.

whaught: Let's remove. We added the nicer widget just this week https://github.com/google/exposure-notifications-verification-server/pull/702 because this is a known point of confusion

whaught: We could also note that more-than-one second factor can be useful in case the user loses access to their primary device.

whaught (commented):
/ok-to-test

And welcome! Thanks for helping out.

icco:
/approved

I'll let @whaught put the lgtm on, looks great though, thank you!

greggles:
Cool, now addressed those suggested changes and udpated the summary. Happy to hear any more feedback if there is any.

sethvargo (approved):
/lgtm

PR #731, opened 2020-10-01
Always include build_id and build_tag in metrics

sethvargo:
Also refactor realm_id injection to be simpler, save a database lookup.

Refs https://github.com/google/exposure-notifications-server/issues/1023

**Release Note**

```release-note
Always include build_id and build_tag in metrics
```

### DIFF
--- a/pkg/controller/certapi/certificate.go
+++ b/pkg/controller/certapi/certificate.go
@@ -28,14 +28,13 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/jwthelper"
 	"github.com/google/exposure-notifications-verification-server/pkg/observability"
 	"go.opencensus.io/stats"
-	"go.opencensus.io/tag"
 
 	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1"
 )
 
 func (c *Controller) HandleCertificate() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+		ctx := observability.WithBuildInfo(r.Context())
 
 		authApp := controller.AuthorizedAppFromContext(ctx)
 		if authApp == nil {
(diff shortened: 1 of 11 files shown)


pkg/controller/certapi/certificate.go
```diff
@@ -45,17 +44,8 @@ func (c *Controller) HandleCertificate() http.Handler {
 			return
 		}
 
-		// This is a non terminal error, as we're only using the realm for stats.
-		realm, err := authApp.Realm(c.db)
```
sethvargo, line 49: We don't need to lookup the realm to get it's ID - it's already a field on the authapp, so dropped this db lookup.

pkg/controller/certapi/certificate.go
```diff
@@ -45,17 +44,8 @@ func (c *Controller) HandleCertificate() http.Handler {
...
-		if err != nil {
-			c.logger.Errorf("unable to load realm", "error", err)
-		} else {
-			ctx, err = tag.New(ctx,
```
sethvargo, line 53: I think this was actually a subtle bug that we just never hit... If `tag.New` has an error, I think the returned context is nil.

pkg/database/database.go
```diff
@@ -271,16 +270,46 @@ func callbackIncrementMetric(ctx context.Context, m *stats.Int64Measure, table s
...
 		if ok && field.Field.CanInterface() && field.Field.Interface() != nil {
-			realmID := field.Field.Interface()
-
-			var err error
-			ctx, err = tag.New(ctx, tag.Upsert(observability.RealmTagKey, fmt.Sprintf("%d", realmID)))
-			if err != nil {
-				_ = scope.Err(fmt.Errorf("failed to add realm tag to metrics: %w", err))
+			realmIDRaw := field.Field.Interface()
+
+			var realmID uint
```
icco, line 275: Could you add a comment why this all needs to happen?

pkg/database/database.go
```diff
@@ -271,16 +270,46 @@ func callbackIncrementMetric(ctx context.Context, m *stats.Int64Measure, table s
...
+			switch t := realmIDRaw.(type) {
+			case uint:
+				realmID = t
+			case uint8:
```
jeremyfaller, line 279: slightly more succinct is:

```
case uint8, uint16, uint32, ... :
  realmId = uint(t)
```

jeremyfaller (approved):
I love the new warnings we have on Errorw

pkg/database/database.go
sethvargo: @jeremyfaller that does not work - maybe you know a good Go compiler engineer who could fix it, but `t` is only the concrete type when given a single element in the case statement. Otherwise t is `interface{}`. I would love to do the more succinct version 😄

pkg/observability/observability.go
```diff
@@ -12,13 +15,46 @@ const (
...
-func init() {
-	var err error
-	RealmTagKey, err = tag.NewKey("realm")
+// CommonTagKeys returns the slice of common tag keys that should used in all
+// views.
+func CommonTagKeys() []tag.Key {
+	return []tag.Key{
+		BuildIDTagKey,
+		BuildTagTagKey,
+		RealmTagKey,
```
yegle, line 29: We should also include the revision name (from env var `K_REVISION`). A redeployment (e.g. changing an environment variable) would change the revision name.

sethvargo: Done in a later PR

PR #732, opened 2020-10-01
Emit buildinfo on health checks

sethvargo:
Part of https://github.com/google/exposure-notifications-server/issues/1023

**Release Note**

```release-note
Emit buildinfo on health checks
```

### DIFF
--- a/pkg/controller/healthz.go
+++ b/pkg/controller/healthz.go
@@ -7,7 +7,9 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/observability"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
+	"go.opencensus.io/stats"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 
@@ -16,15 +18,21 @@ import (
 
 var (
 	rl = rate.NewLimiter(rate.Every(time.Minute), 1)
+
+	mHealthzBuildInfo = stats.Int64("build_info", "Current build information", "1")
 )
 
-func HandleHealthz(hctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
-	logger := logging.FromContext(hctx).Named("healthz")
+func HandleHealthz(ctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
+	logger := logging.FromContext(ctx).Named("healthz")
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+		ctx := observability.WithBuildInfo(r.Context())
+
+		// Emit the buildinfo metric
+		stats.Record(ctx, mHealthzBuildInfo.M(1))
+
 		params := r.URL.Query()
-		if s := params.Get("service"); s == "database" {
+		if params.Get("service") == "database" {
 			if cfg == nil {
 				InternalError(w, r, h, fmt.Errorf("database not configured for health check"))
 				return


sethvargo:
@yegle @icco I think we need to also setup a view for this? I'm not sure the best way to initiate that besides a package-level `init` (which I prefer to ignore) since a controller is never "instantiated". 

We also probably need a way to add another tag to this metric which is the name of the service being deployed?

sethvargo:
TBH, I don't entirely know the motivation or uses of https://github.com/google/exposure-notifications-server/issues/1023, so I'm not sure this is the right solution.

yegle:
If I'm not mistaken, what @icco want is to add the build id information to the monitored labels (https://github.com/google/exposure-notifications-server/blob/b1b596c1434224b2bb9e582a3a463db75e9c79c9/pkg/observability/resource.go#L90), so all of our graphs can be group by the build id. Right now, if there's a change to our graph, we won't know if it was due to a new binary.

I would also suggest to add the revision id to the labels, since config changes (e.g. environment variable values) will not change build labels, but is something we may want to group by.

yegle:
Upon closer look, I think what @icco want is to have a way to know the # of running instances. I just checked https://cloud.google.com/monitoring/api/metrics_gcp#gcp-run and was surprised we don't have such a built-in metric already.

pkg/controller/healthz.go
```diff
@@ -16,15 +18,21 @@ import (
...
-	logger := logging.FromContext(hctx).Named("healthz")
+func HandleHealthz(ctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
+	logger := logging.FromContext(ctx).Named("healthz")
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+		ctx := observability.WithBuildInfo(r.Context())
+
+		// Emit the buildinfo metric
+		stats.Record(ctx, mHealthzBuildInfo.M(1))
```
yegle, line 32: We'll have to treat this metric as normal metric, create a view for it and register with `view.Register()`, even if we only record the metric once for the lifetime of the instance.

The godoc on this: https://godoc.org/go.opencensus.io/stats#example-Record

> // Without any views or exporters registered, this statement has no observable effects.

sethvargo:
>  so all of our graphs can be group by the build id.

I did that in https://github.com/google/exposure-notifications-verification-server/pull/731. 

> I would also suggest to add the revision id to the labels, since config changes (e.g. environment variable values) will not change build labels, but is something we may want to group by.

Ack - I can do that: https://github.com/google/exposure-notifications-verification-server/pull/736

@yegle do you think it makes sense to do it like this? Or can we pull the data we need from any of the existing metrics?

pkg/controller/healthz.go
```diff
@@ -16,15 +18,21 @@ import (
 
 var (
 	rl = rate.NewLimiter(rate.Every(time.Minute), 1)
+
+	mHealthzBuildInfo = stats.Int64("build_info", "Current build information", "1")
```
icco, line 22: observability.MetricRoot + "/build_info" for first arg, stats.UnitDimensionless for last arg.

pkg/controller/healthz.go
```diff
@@ -16,15 +18,21 @@ import (
...
 )
 
-func HandleHealthz(hctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
-	logger := logging.FromContext(hctx).Named("healthz")
+func HandleHealthz(ctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
+	logger := logging.FromContext(ctx).Named("healthz")
```
icco, line 26: Why not create the view here, this function is only called once right?

sethvargo: It's called multiple times in the e2e/performance things.

yegle:
Per some internal communication, the Cloud Monitoring built-in metric for Cloud Run instances will be released soon and we can just wait for that to happen.

If we really want to know the # of instances without the built-in metric, I think queries like this would be more useful:

```
fetch generic_task
| metric
    'custom.googleapis.com/opencensus/opencensus.io/http/server/request_count'
| align rate(1m)
| every 1m
| group_by [resource.task_id], [row_count: row_count()]
```

(querying the num of instances by counting the task_ids of actual served HTTP request).

PR #734, opened 2020-10-01
Show skip link properly with grace period

whaught:
## Proposed Changes

* Ensures that the 'skip' option is shown for MFARequired realms if the user is within the grace-period.

### DIFF
--- a/pkg/controller/login/register.go
+++ b/pkg/controller/login/register.go
@@ -31,7 +31,17 @@ func (c *Controller) HandleRegisterPhone() http.Handler {
 			return
 		}
 
+		user := controller.UserFromContext(ctx)
+		if user == nil {
+			controller.MissingUser(w, r, c.h)
+			return
+		}
+
+		realm := controller.RealmFromContext(ctx)
+
 		m := controller.TemplateMapFromContext(ctx)
+
+		m["mfaMode"] = realm.EffectiveMFAMode(user)
 
 		if controller.MFAPromptedFromSession(session) {
 			m["isPrompt"] = true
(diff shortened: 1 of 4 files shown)


pkg/database/realm.go
```diff
@@ -207,6 +207,19 @@ type Realm struct {
 	Tokens []*Token            `gorm:"PRELOAD:false; SAVE_ASSOCIATIONS:false; ASSOCIATION_AUTOUPDATE:false, ASSOCIATION_SAVE_REFERENCE:false"`
 }
 
+func (realm *Realm) EffectiveMFAMode(user *User) *AuthRequirement {
```
jeremyfaller, line 210: return by value, saves a bunch of this mucking about with temps.

jeremyfaller: Also saves the indirection in the caller.

whaught: lol idk how I got so far with these const -> var pointers without thinking "this is easier as a value". Thanks

pkg/controller/login/register.go
```diff
@@ -31,8 +31,18 @@ func (c *Controller) HandleRegisterPhone() http.Handler {
 			return
 		}
 
+		user := controller.UserFromContext(ctx)
```
sethvargo, line 34: Nit: `currentUser`

pkg/database/realm.go
```diff
@@ -207,6 +207,17 @@ type Realm struct {
 	Tokens []*Token            `gorm:"PRELOAD:false; SAVE_ASSOCIATIONS:false; ASSOCIATION_AUTOUPDATE:false, ASSOCIATION_SAVE_REFERENCE:false"`
 }
 
+func (r *Realm) EffectiveMFAMode(user *User) AuthRequirement {
```
sethvargo, line 210: Needs docs since the value changes depending on time.

sethvargo (changes requested):

sethvargo (approved):
/lgtm

PR #736, opened 2020-10-02
Include knative info in metrics if present

sethvargo:
**Release Note**

```release-note
Include knative info in metrics if present
```

### DIFF
--- a/pkg/observability/observability.go
+++ b/pkg/observability/observability.go
@@ -3,6 +3,7 @@ package observability
 
 import (
 	"context"
+	"os"
 	"strconv"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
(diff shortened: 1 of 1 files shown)


whaught (commented):
/lgtm

pkg/observability/observability.go
```diff
@@ -45,16 +55,31 @@ func WithRealmID(octx context.Context, realmID uint) context.Context {
...
+// to the observability context.
 func WithBuildInfo(octx context.Context) context.Context {
-	ctx, err := tag.New(octx,
-		tag.Upsert(BuildIDTagKey, buildinfo.BuildID),
-		tag.Upsert(BuildTagTagKey, buildinfo.BuildTag))
+	tags := make([]tag.Mutator, 0, 5)
+	tags = append(tags, tag.Upsert(BuildIDTagKey, buildinfo.BuildID))
+	tags = append(tags, tag.Upsert(BuildTagTagKey, buildinfo.BuildTag))
+
+	if knativeService != "" {
```
yegle, line 65: Either log it when it's empty, or add the tag even if the value is empty.

Per the Cloud Run contract, the env var should not be empty. When they are empty, we won't want to silently ignore the issue.

yegle (approved):

pkg/observability/observability.go
sethvargo: We want to support cases where this runs outside of Cloud Run though. We know other people are running this code on other platforms. We aren't specifically tied to GCP.

PR #739, opened 2020-10-02
Nit: switch user -> currentUser

whaught:
## Proposed Changes

* Nit: we're calling the user from context `currentUser`
* Let's do this all at once so I don't forget this in other PRs and copy/paste propagate it


sethvargo (approved):
/lgtm

Issue #735, opened 2020-10-02
Add page to see current realm quotas


sethvargo:
At present, the only way to see the current remaining quota for a realm for the day is to go to Redis. It would be nice to have a page where realm admins (or at least system admins) could see the currently-available quota.

PR #743, opened 2020-10-02
Attempt to fix realm quota alert

sethvargo:
- Display the currently configured limit and remaining tokens on the realm settings page.
- Only create the metrics for realms that have abuse prevention enabled.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/735

**Release Note**

```release-note
Display the currently configured limit and remaining tokens on the realm settings page
```

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -192,30 +192,34 @@ func (c *Controller) HandleIssue() http.Handler {
 			}
 		}
 
-		// If we got this far, we're about to issue a code.
-		dig, err := digest.HMACUint(realm.ID, c.config.GetRateLimitConfig().HMACKey)
-		if err != nil {
-			controller.InternalError(w, r, c.h, err)
-			return
-		}
-		key := fmt.Sprintf("realm:quota:%s", dig)
-		limit, remaining, reset, ok, err := c.limiter.Take(ctx, key)
-		if err != nil {
-			logger.Errorw("failed to take from limiter", "error", err)
-			stats.Record(ctx, c.metrics.QuotaErrors.M(1))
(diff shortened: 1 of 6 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -192,31 +192,35 @@ func (c *Controller) HandleIssue() http.Handler {
...
+		// to ensure this is permitted.
+		if realm.AbusePreventionEnabled {
+			dig, err := digest.HMACUint(realm.ID, c.config.GetRateLimitConfig().HMACKey)
+			if err != nil {
+				controller.InternalError(w, r, c.h, err)
 				return
 			}
+			key := fmt.Sprintf("realm:quota:%s", dig)
+			limit, remaining, reset, ok, err := c.limiter.Take(ctx, key)
+			c.recordCapacity(ctx, limit, remaining)
```
sethvargo, line 205: Moved this up - we were only recording capacity later, which I think would cause us to miss some data points.

pkg/controller/issueapi/issue.go
```diff
@@ -296,13 +298,9 @@ func (c *Controller) getAuthorizationFromContext(r *http.Request) (*database.Aut
 	return authorizedApp, currentUser, nil
 }
 
-func (c *Controller) recordCapacity(ctx context.Context, realm *database.Realm, remaining uint64) {
-	if !realm.AbusePreventionEnabled {
-		return
-	}
+func (c *Controller) recordCapacity(ctx context.Context, limit, remaining uint64) {
```
sethvargo, line 301: I changed this to rely on the value that comes from the limiter, not the value on the realm. The enforcement happens based on what's in the limiter, so we should alert and record off of that.

jeremyfaller (approved):

PR #745, opened 2020-10-03
Too many attempts message

whaught:
## Proposed Changes

* Show "Too many attempts" error message in flash
     * We currently eat firebase errors and show a generic one. This was done to hide certain responses (user not-found) that could leak information. However too-many-attempts/requests we should let the user know about so they don't keep trying.
* This issue was found by an early tester who was rapidly resetting password.

**Release Note**

```release-note
Show error message on too-many-attempts
```

### DIFF
--- a/cmd/server/assets/login/login.html
+++ b/cmd/server/assets/login/login.html
@@ -138,6 +138,10 @@
                 flash.clear();
                 flash.error('Unsupported 2nd factor authentication type.');
               }
+            } else if (error.code == 'auth/too-many-requests'){
+              flash.clear();
+              flash.error(err.message);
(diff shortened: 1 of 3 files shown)


cmd/server/assets/login/login.html
```diff
@@ -138,6 +138,10 @@
                 flash.clear();
                 flash.error('Unsupported 2nd factor authentication type.');
               }
+            } else if (error.code == 'auth/too-many-requests'){
```
sethvargo, line 141: Up on Line 114, I think we should also `console.error(error)`

sethvargo (approved):

PR #748, opened 2020-10-05
Minor tweaks

whaught:
## Proposed Changes

* Add a note about quota to import
* Log errors on login to the console

### DIFF
--- a/cmd/server/assets/login/login.html
+++ b/cmd/server/assets/login/login.html
@@ -112,7 +112,6 @@
           .then(function(userCredential) {
             loginSuccess();
           }).catch(function(error) {
-            console.error(error);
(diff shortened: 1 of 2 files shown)


cmd/server/assets/login/login.html
```diff
@@ -112,6 +112,7 @@
           .then(function(userCredential) {
             loginSuccess();
           }).catch(function(error) {
+            console.error(error);
```
mikehelmick, line 115: do we want to always log this to the console?

whaught: I can switch it to just unhandled to be more conservative. This was a suggestion from @sethvargo to make debugging easier

sethvargo (approved):
/lgtm

PR #751, opened 2020-10-06
[WIP] Custom email verify page

whaught:
## Proposed Changes

Now that we've switched to a custom password-verify page, the firebase one is gone. It's all-or-nothing with the custom email links, so we need a custom email-verify too.

* I'm trying to use an internal API but get 'missing_local_id' from the API response. It doesn't appear to like our credential on the API call from the server (even with an API key attached).
    * We may have to go back to JS on the client for this call. Note: with the oobCode included that may be okay, this API doesn't leak details.
    * I also may have missed something here on my request. It looks like account:update is highly overloaded for many purposes.

### DIFF
--- a/internal/firebase/verify_email_code.go
+++ b/internal/firebase/verify_email_code.go
@@ -24,52 +24,45 @@ import (
 	"net/http"
 )
 
-type verifyPasswordResetCodeRequest struct {
-	Code        string `json:"oobCode"`
-	NewPassword string `json:"newPassword,omitempty"`
+type verifyEmailRequest struct {
+	Code string `json:"oobCode"`
 }
 
-// VerifyPasswordResetCode is called with the one-time-code given from a reset email to the user.
-// It can be used to check that the code is valid without making changes to the user.
+// VerifyEmailCode sends a password reset email to the user. If the new
+// password is given, it applies the password reset change with the new password
+// using the code.
 //
 // See: https://firebase.google.com/docs/reference/rest/auth#section-send-password-reset-email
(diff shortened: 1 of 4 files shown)


whaught:
@sethvargo have you seen `missing_local_id` responses or is there anything patently off about my request to account:update?

internal/firebase/verify_email_code.go
```diff
@@ -0,0 +1,80 @@
...
+// See: https://firebase.google.com/docs/reference/rest/auth#section-send-password-reset-email
+func (c *Client) VerifyEmailCode(ctx context.Context, code, key string) error {
+	r := &verifyEmailRequest{Code: code}
+
+	var body bytes.Buffer
+	if err := json.NewEncoder(&body).Encode(r); err != nil {
+		return fmt.Errorf("failed to create json body: %w", err)
+	}
+
+	u := c.buildURL("/v1/accounts:update")
```
sethvargo, line 44: Is this correct? Should it be `/v1/accounts:resetPassword` per https://firebase.google.com/docs/reference/rest/auth#section-verify-password-reset-code ?

whaught: Unfortunately no: https://cloud.google.com/identity-platform/docs/use-rest-api#section-confirm-email-verification

sethvargo: `<rant>`I hate anchor links.`</rant>`

In that case, I don't see any obvious reason why this wouldn't work?

whaught:
We're just going to do this in javascript, as Firebase intended

/close

PR #756, opened 2020-10-06
Recreate firebase user, no metric

whaught:
## Proposed Changes

* When resetting password, ensure that if we have a db.User that we also have a firebase user
* Move password-reset to server-side for Users/ page

**Note: ** This does send the user back through the email verification flow (which is currently bundled with our password-reset invitation and therefore not seen by most users)

_This PR fixes the issue without adding a metric. That will be done as a separate PR. See https://github.com/google/exposure-notifications-verification-server/pull/685_

**Release Note**

```release-note
Ensure that password-reset flows also check that the firebase user auth is created if it's missing and we have an entry for the user.
```

### DIFF
--- a/pkg/controller/user/create.go
+++ b/pkg/controller/user/create.go
@@ -81,16 +81,9 @@ func (c *Controller) HandleCreate() http.Handler {
 		}
 
 		// Create firebase user first, if this fails we don't want a db.User entry
-		if created, err := user.CreateFirebaseUser(ctx, c.client); err != nil {
-			flash.Alert("Failed to create user: %v", err)
+		if _, err := c.createFirebaseUser(ctx, user); err != nil {
 			c.renderNew(ctx, w)
 			return
-		} else if created {
-			if err := c.firebaseInternal.SendPasswordResetEmail(ctx, user.Email); err != nil {
-				flash.Error("Failed sending new user invitation: %v", err)
-				c.renderNew(ctx, w)
-				return
-			}
 		}
 
 		// Build the user struct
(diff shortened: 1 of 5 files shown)


pkg/controller/user/reset_password.go
```diff
@@ -0,0 +1,58 @@
...
+	created, err := user.CreateFirebaseUser(ctx, c.client)
+	if err != nil {
+		flash.Alert("Failed to create user auth: %v", err)
+		return created, err
+	}
+
+	// Reset if we just created or a reset was asked for.
+	if created || reset {
+		if err := c.firebaseInternal.SendPasswordResetEmail(ctx, user.Email); err != nil {
+			flash.Error("Failed sending new user invitation: %v", err)
```
jeremyfaller, line 53: s/Failed sending/Could not send/

jeremyfaller (changes requested):

sethvargo (approved):
/lgtm

pkg/controller/user/create.go
```diff
@@ -81,16 +81,9 @@ func (c *Controller) HandleCreate() http.Handler {
 		}
 
 		// Create firebase user first, if this fails we don't want a db.User entry
-		if created, err := user.CreateFirebaseUser(ctx, c.client); err != nil {
-			flash.Alert("Failed to create user: %v", err)
+		if _, err := c.createFirebaseUser(ctx, user); err != nil {
 			c.renderNew(ctx, w)
```
sethvargo, line 85: I don't think we should swallow this error.

whaught: it's not swallowed. c.createFirebaseUser does the flash now (the flash is the same for the places that use the helper)

sethvargo: ah i see

PR #762, opened 2020-10-06
Don't fail terraform if no custom domain supplied

chaodaiG:


terraform/verification-lb.tf
```diff
@@ -181,5 +181,5 @@ resource "google_compute_managed_ssl_certificate" "default" {
 }
 
 output "lb_ip" {
-  value = google_compute_global_address.verification-server[0].address
+  value = local.enable_lb ? google_compute_global_address.verification-server[0].address : ""
```
sethvargo, line 184: Does this fail? I thought I tried and it just did not print the output.

chaodaiG: oh yeah it failed:
```
Error: Invalid index
  on ../terraform/verification-lb.tf line 184, in output "lb_ip":
 184:   value = google_compute_global_address.verification-server[0].address
    |----------------
    | google_compute_global_address.verification-server is empty tuple

```

https://prow.k8s.io/view/gcs/oss-prow/logs/ci-en-verification-server-terraform-smoke/1313600219765542912#1:build-log.txt%3A189

yegle: Use null to be explicit.

sethvargo: What if you make it `[google_compute_global_address.verification-server.*.address]`? I don't like conditionals in outputs

chaodaiG: trying it out

chaodaiG: Updated to use `*`, it works for e2e, but not sure whether it works in prod though

Issue #532, opened 2020-09-11
Create a new project for Cloud Monitoring workspace


yegle:
### TL;DR

Per the doc, the recommended way to set up multi-project monitoring is to have a host project for the monitoring workspace: https://cloud.google.com/monitoring/workspaces/create#multi-project-ws

We need a host project to host all the alerts and dashboards. Maybe create a project named `encv-monitoring`?

yegle:
We might not need this after all. We are currently fine with having dashboards/alerts in different projects.

PR #763, opened 2020-10-06
Move verification server related monitoring configs to a sub folder.

yegle:
Move verification server related configs (all configs right now) to a sub folder. We will be creating key server related configs in its own sub folder.

Part of #532


terraform/alerting/main.tf
```diff
@@ -0,0 +1,10 @@
+module "verification-server" {
```
icco, line 1: Can we import a module into another module?

yegle: Running `tf apply` on my test project seem to work as intended.

PR #765, opened 2020-10-07
Handle edge case when realm quota is first enabled

sethvargo:
If realm quota is enabled and then someone "takes" from the bucket before the modeler runs, then the realm gets the default quota (which is low). Worse, it gets the default TTL (60s), which also isn't correct.

More comments inline in the code

**Release Note**

```release-note
Handle edge case when realm quota is first enabled, enforce realm quota by default.
```

### DIFF
--- a/pkg/controller/modeler/modeler.go
+++ b/pkg/controller/modeler/modeler.go
@@ -35,6 +35,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	oneWeek = 7 * 24 * time.Hour
+)
+
 // Controller is a controller for the modeler service.
 type Controller struct {
 	config  *config.Modeler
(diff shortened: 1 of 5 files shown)


pkg/controller/modeler/modeler.go
```diff
@@ -211,7 +215,7 @@ func (c *Controller) rebuildModel(ctx context.Context, id uint64) error {
 		return fmt.Errorf("failed to digest realm id: %w", err)
 	}
 	key := fmt.Sprintf("realm:quota:%s", dig)
-	if err := c.limiter.Set(ctx, key, uint64(effective), 24*time.Hour); err != nil {
+	if err := c.limiter.Set(ctx, key, uint64(effective), oneWeek); err != nil {
```
whaught, line 218: We're changing this from one-day to one-week. We might want to note in the description. Should `ttl` in realmadmin/settings ref this constant?

sethvargo: Oops this shouldn't be in. This is remnant from my debugging... Removed.

whaught (approved):
/lgtm

PR #766, opened 2020-10-07
Adding support for App Links, fallback to Play Store.

flagxor:
Redirector will first redirect to itself,
on Android adding bounce=1 to ensure app link has a chance
to trigger.

iOS is also supported, but for now assuming it doesn't need
the extra bounce.

When an App Link is not configured for a platform,
continue to fall back to ens://

**Release Note**

```release-note
Apps that can support app links should now configure a store URL.
```

### DIFF
--- a/pkg/controller/redirect/appstore.go
+++ b/pkg/controller/redirect/appstore.go
@@ -1,0 +1,54 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package redirect
+
+import (
+	"fmt"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+)
+
+type AppStoreData struct {
+	AndroidURL string `json:"androidURL"`
+	IOSURL     string `json:"iosURL"`
+}
+
+// getAppStoreData finds data tied to app store listings.
+func (c *Controller) getAppStoreData(realmID uint) (*AppStoreData, error) {
+	// Pick first Android app (in the realm) for Play Store redirect.
+	androidURL := ""
+	androidApps, err := c.db.ListActiveAppsByOS(realmID, database.OSTypeAndroid)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Android Apps: %w", err)
+	}
+	if len(androidApps) > 0 {
+		androidURL = androidApps[0].URL
+	}
+
+	// Pick first iOS app (in the realm) for Store redirect.
+	iosURL := ""
+	iosApps, err := c.db.ListActiveAppsByOS(realmID, database.OSTypeIOS)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get iOS Apps: %w", err)
+	}
+	if len(iosApps) > 0 {
+		iosURL = iosApps[0].URL
+	}
+
+	return &AppStoreData{
+		AndroidURL: androidURL,
+		IOSURL:     iosURL,
+	}, nil
+}
--- a/pkg/controller/redirect/index.go
+++ b/pkg/controller/redirect/index.go
@@ -15,6 +15,7 @@
 package redirect
 
 import (
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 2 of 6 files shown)


pkg/controller/redirect/index.go
```diff
@@ -18,20 +18,62 @@ import (
...
 			host = host[0:i]
 		}
 
+		// Check for browser type.
+		onAndroid := strings.Contains(strings.ToLower(r.UserAgent()), "android")
+		onIOS := strings.Contains(strings.ToLower(r.UserAgent()), "iphone")
+
+		// Bounce to self if we haven't already (on Android only).
+		if onAndroid && r.FormValue("bounce") == "" {
+      q := r.URL.Query()
```
mikehelmick, line 44: indentation

pkg/controller/redirect/index.go
```diff
@@ -18,20 +18,62 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 )
 
+// cacheTTL is the amount of time for which to cache these values.
+const cacheTTL = 30 * time.Minute
```
mikehelmick, line 27: best to make this part of config w/ default of 30m so runtime configurable

pkg/controller/redirect/index.go
```diff
@@ -18,20 +18,62 @@ import (
...
 		if i := strings.Index(host, ":"); i > 0 {
 			host = host[0:i]
 		}
 
+		// Check for browser type.
+		onAndroid := strings.Contains(strings.ToLower(r.UserAgent()), "android")
+		onIOS := strings.Contains(strings.ToLower(r.UserAgent()), "iphone")
+
+		// Bounce to self if we haven't already (on Android only).
+		if onAndroid && r.FormValue("bounce") == "" {
```
mikehelmick, line 43: What's the reason for the "bounce" ?
I couldn't find documentation supporting this

sethvargo, line 43: Will this be a form value or a URL query param coming back?

pkg/controller/redirect/index.go
sethvargo, line 44: Nit: Run `go fmt`.

flagxor: done

flagxor: done

pkg/controller/redirect/index.go
flagxor: FormValue favors explicit forms in PUT/POST requests, but falls back to query params, which is all you got for GET.

pkg/controller/redirect/index.go
flagxor: done

pkg/controller/redirect/index.go
flagxor: Added a comment, but also seeking better documentation / kicking the tires on SMS apps that anecdotally need this: Samsung SMS & Facebook messenger.

sethvargo: Huh TIL. Thanks!

tools/seed/main.go
```diff
@@ -172,7 +172,7 @@ func realMain(ctx context.Context) error {
 		{
 			Name:    "Example iOS app",
 			RealmID: realm1.ID,
-			URL:     "http://apple.com",
+			URL:     "",
```
jeremyfaller, line 175: Should this still be here? I assume this was from development/debugging.

flagxor: Unsure, we'd typically want these blank as App links on iOS will initially be rare, whereas on Android all ENX apps will have them.
WDYT?

pkg/controller/redirect/appstore.go
```diff
@@ -0,0 +1,54 @@
...
+	IOSURL     string `json:"iosURL"`
+}
+
+// getAppStoreData finds data tied to app store listings.
+func (c *Controller) getAppStoreData() (*AppStoreData, error) {
+	// Pick first Android app for Play Store redirect.
+	androidURL := ""
+	androidApps, err := c.db.ListActiveAppsByOS(database.OSTypeAndroid)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get Android Apps: %w", err)
```
sethvargo, line 34:
```suggestion
		return nil, fmt.Errorf("failed to get Android Apps: %w", err)
```

pkg/controller/redirect/appstore.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	if len(androidApps) > 0 {
+		androidURL = androidApps[0].URL
+	}
+
+	// Pick first iOS app for Store redirect.
+	iosURL := ""
+	iosApps, err := c.db.ListActiveAppsByOS(database.OSTypeIOS)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get iOS Apps: %w", err)
```
sethvargo, line 44:
```suggestion
		return nil, fmt.Errorf("failed to get iOS Apps: %w", err)
```

sethvargo (commented):
These are why the linter is sad

pkg/controller/redirect/index.go
flagxor: Confirmed Facebook Messenger needs this.

pkg/controller/redirect/appstore.go
flagxor: done

pkg/controller/redirect/appstore.go
flagxor: done

pkg/controller/redirect/index_test.go
```diff
@@ -14,41 +14,77 @@
...
 		name   string
 		path   string
+		query  url.Values
 		region string
 		exp    string
 	}{
 		{
 			name:   "leading_slash_region",
 			path:   "v",
+			query:  url.Values{},
```
icco, line 35: In the future you could save yourself some typing by not including this line in your table since this is the default value.

icco (approved):
/lgtm

pkg/controller/redirect/index_test.go
flagxor: Ah, good point.
Done.

PR #767, opened 2020-10-07
Fix bad ens:// protocol (E_MISSING_ONE_SLASH)

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/redirect/index.go
+++ b/pkg/controller/redirect/index.go
@@ -17,6 +17,7 @@ package redirect
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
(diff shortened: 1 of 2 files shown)


mikehelmick (approved):

pkg/controller/redirect/index.go
```diff
@@ -34,8 +35,15 @@ func (c *Controller) HandleIndex() http.Handler {
 
 		for hostname, region := range c.hostnameToRegion {
 			if host == hostname {
-				sendTo := fmt.Sprintf("ens:/%s&r=%s", path, region)
```
flagxor, line 37: FWIW, this was "okay" in the sense that path has a leading slash.
But building the URI explicitly is better.

flagxor (approved):
lgtm

sethvargo (commented):
Updated and added some tests 

/unhold
/assign @mikehelmick

mikehelmick (approved):

flagxor:
Actually this introduced a regression (it eats the codes now).
Fixing as part of the url CL.

PR #772, opened 2020-10-07
Remove prevent_destroy

sethvargo:
Terraform already requires confirmation, and this makes it difficult to iterate.

/cc @chaodaiG for thoughts

**Release Note**

```release-note
NONE
```


chaodaiG:
Since I would not imagine anyone does `terraform destroy` on prod server without proper backing up, this change makes sense to me

/lgtm

Issue #773, opened 2020-10-07
Deduplicate metrics


yegle:
In many (all?) places, we are using two separate metrics to count failed/successful states. E.g. we have `code_issued_count` and `code_issue_error_count` [here](https://github.com/google/exposure-notifications-verification-server/blob/44a426a524442aedc164f3f2a7df809adb55c3cf/pkg/controller/issueapi/metrics.go#L60-L72).

Having two metrics makes it impossible to calculate an error ratio which is a very important metric to monitor. And we will need to add more metrics if we want to further break down what caused the error.

Instead we should have a single `code_issued_count` metric, with additional labels like `error_reason` (short text on why it failed) and `blame` (whether the failure is to blame the client or the server). This, together with the realm id information that we already have, allows us to setup flexible alerting rules:

- Overall `blame=server` error rate/ratio increase
- Per-realm-id `blame=client` error rate/ratio increase
- One particular `error_reason` rate/ratio increase

sethvargo:
@yegle I know you've done a lot of work on this - are there specific remaining items you'd like to finish before closing this out?

yegle:
Yes I want to use the new metrics in the alert policies. One of the blocking issue is Terraform lack support to define alert using MQL.

I'll do a review and close this, then track defining alerts using MQL in a separate issue.

PR #775, opened 2020-10-07
Add a /request metric with blame/result field.

yegle:
Part of https://github.com/google/exposure-notifications-verification-server/issues/773

The new metric will eventually replace all other metrics under /api/certificate.

### DIFF
--- a/pkg/controller/certapi/certificate.go
+++ b/pkg/controller/certapi/certificate.go
@@ -28,6 +28,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/jwthelper"
 	"github.com/google/exposure-notifications-verification-server/pkg/observability"
 	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
 
 	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1"
 )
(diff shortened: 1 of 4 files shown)


icco (approved):
/lgtm

pkg/controller/certapi/certificate.go
```diff
@@ -36,11 +37,18 @@ func (c *Controller) HandleCertificate() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := observability.WithBuildInfo(r.Context())
 
+		var blame = controller.BlameNone
+		var result = controller.APIResultOK()
+
+		defer stats.RecordWithTags(ctx, []tag.Mutator{blame, result}, mRequest.M(1))
```
yegle, line 43: @icco Do you know if the linter (complain about blame/result variables not used later in the code) is correct?

I'm a bit fuzzy here, would this defer work as intended (use the blame/result variable's value right before return), or will it capture their current value and thus the blame/result variables later in the file is ignored?

icco: I'm pretty sure it's the latter.

yegle:
Can I get another LGTM?

pkg/controller/certapi/certificate.go
yegle: Fixed by capturing a pointer in the closure. Linter no longer complain about unused variables.

icco (approved):
/lgtm

Issue #232, opened 2020-08-14
Show history of issued codes with status


whaught:
### TL;DR

Create a page where a user or realm admin can see a list of recently issued codes.

### Design

**Proposal**

- Create a List API for codes.
    - Users get the codes they've issued themselves
    - Realm Admins see all codes
    - Consider pagination and limits

- The original mock calls for showing the
    - test result
    - code
    - expiry
    - anonymized phone (***) ***-**31
        - Note: no. nothing to identify. just uuid.
    - Invalidate code button?

Note: one problem with this (and why we have uuid) is that not all systems will use phone# and may require the user to reach out manually. We don't have any other associated data on who the patient is which makes it tough for the user to know who the code is intended for and follow-up if unclaimed. The uuid allows this or alternative clients to keep such data locally.

**Alternatives considered**

I've created https://github.com/google/exposure-notifications-verification-server/pull/227 where a user can check a single-id at a time. This, however, requires the user to keep a list of their issued UUIDs which may be less useful (the API is possible more useful programmatically rather than as UX) but allows the user to track the private patient data locally and separately. The history table could replace this if its exhaustive.

whaught:
Note: We've since decided to hash the codes. That may make this less useful

mikehelmick:
Would it make sense to show the UUIDs instead?

It would for someone doing an external integration, because they could tie that back into the CRM

sethvargo:
If the use case is external integration, it seems like an API might be more appropriate?

mikehelmick:
both - in general, things we put in the UI should probably also have APIs

PR #780, opened 2020-10-08
Show code createdAt in local time

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/232

## Proposed Changes

* Server sends createdAt in UTC, now we interpret them on the client in javascript and replace them in local time
    * Should be a little user-friendlier

**Release Note**

```release-note
Show code created time in local timezone
```

### DIFF
--- a/cmd/server/assets/codestatus/index.html
+++ b/cmd/server/assets/codestatus/index.html
@@ -50,12 +50,12 @@
     <div class="card mb-3 shadow-sm">
       <div class="card-header">Recently issued codes</div>
       <div class="card-body">
-        <div class="list-group">
+        <div class="list-group" id="recent-codes">
           {{range $code := .recentCodes}}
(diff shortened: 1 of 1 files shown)


cmd/server/assets/codestatus/index.html
```diff
@@ -50,7 +50,7 @@ <h1>Verification code status</h1>
     <div class="card mb-3 shadow-sm">
       <div class="card-header">Recently issued codes</div>
       <div class="card-body">
-        <div class="list-group">
+        <div class="list-group" id="recent-codes">
```
sethvargo, line 53: @whaught there's actually a `data-datetime` attribute that will do this automagically and accounts for some edges (see header.html)

whaught: Done. Although I did kind of think the `PST` was a little reassuring

PR #782, opened 2020-10-08
use redirect hostname to lookup realm for correct mobile app

### LINKED ISSUE #779: enx-redreict
incorrect app ownership shown in .well-known/assetlinks.json

if none is set, something is still present.

/kind bug
/assign

mikehelmick:
Fixes #779

## Proposed Changes

* use hostname when deciding what mobile app config to serve on .well-known
* if no configuration, serve a 404
* make cache TTL configurable

**Release Note**

```release-note
BUGFIX: Ensure that .well-known is served for the correct region.
```

### DIFF
--- a/pkg/controller/associated/android.go
+++ b/pkg/controller/associated/android.go
@@ -33,8 +33,17 @@ type Target struct {
 }
 
 // getAndroidData finds all the android data apps.
-func (c *Controller) getAndroidData() ([]AndroidData, error) {
-	apps, err := c.db.ListActiveAppsByOS(database.OSTypeAndroid)
+func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
+	realm, err := c.db.FindRealmByRegion(region)
+	if err != nil {
+		return nil, fmt.Errorf("unable to lookup realm")
+	}
+	if realm == nil {
+		// no realm exists for this region
+		return nil, nil
+	}
+
+	apps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get android data: %w", err)
 	}
--- a/pkg/controller/associated/android.go
+++ b/pkg/controller/associated/android.go
@@ -38,10 +38,6 @@ func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
 	if err != nil {
 		return nil, fmt.Errorf("unable to lookup realm")
 	}
-	if realm == nil {
-		// no realm exists for this region
-		return nil, nil
-	}
 
 	apps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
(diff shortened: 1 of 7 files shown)


pkg/database/realm.go
```diff
@@ -716,6 +716,18 @@ func (db *Database) CreateRealm(name string) (*Realm, error) {
 	return realm, nil
 }
 
+func (db *Database) FindRealmByRegion(region string) (*Realm, error) {
```
sethvargo, line 719: I've moved us<sup>1</sup> away from database lookup functions returning nil because it tends to lead to NPEs. I'd like this to return an error instead, and then the controller should use `database.IsNotFound` to flow control.

<sup>1. </sup> The exception is when the function returns a list. In that case, the empty list (but NOT an uninitialized list) is returned.

pkg/database/mobile_app.go
```diff
@@ -136,12 +136,12 @@ func (a *MobileApp) BeforeSave(tx *gorm.DB) error {
 }
 
 // ListActiveAppsByOS finds all mobile apps by their OS.
-func (db *Database) ListActiveAppsByOS(os OSType) ([]*MobileApp, error) {
+func (db *Database) ListActiveAppsByOS(realmID uint, os OSType) ([]*MobileApp, error) {
```
sethvargo, line 139: Update comment

sethvargo (changes requested):

pkg/controller/associated/android.go
```diff
@@ -33,8 +33,13 @@ type Target struct {
 }
 
 // getAndroidData finds all the android data apps.
-func (c *Controller) getAndroidData() ([]AndroidData, error) {
-	apps, err := c.db.ListActiveAppsByOS(database.OSTypeAndroid)
+func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
+	realm, err := c.db.FindRealmByRegion(region)
+	if err != nil {
+		return nil, fmt.Errorf("unable to lookup realm")
```
sethvargo, line 39:
```suggestion
		return nil, fmt.Errorf("unable to lookup realm: %w", err)
```

pkg/controller/associated/ios.go
```diff
@@ -62,15 +66,21 @@ func (c *Controller) getAppIds() ([]string, error) {
 }
 
 // getIosData gets the iOS app data.
-func (c *Controller) getIosData() (*IOSData, error) {
-	var ids []string
-	var err error
+func (c *Controller) getIosData(region string) (*IOSData, error) {
+	realm, err := c.db.FindRealmByRegion(region)
+	if err != nil {
+		return nil, fmt.Errorf("unable to lookup realm")
```
sethvargo, line 72:
```suggestion
		return nil, fmt.Errorf("unable to lookup realm: %w", err)
```

sethvargo (approved):
Nits, but can be a follow up

PR #792, opened 2020-10-08
Remove unnecessary metric increment in validateToken()

yegle:
We already increse the counter outside of this function.

- https://github.com/google/exposure-notifications-verification-server/blob/f2e04d8eb1abff6da33c77c1952d5931d7591201/pkg/controller/certapi/certificate.go#L91-L98
- https://github.com/google/exposure-notifications-verification-server/blob/f2e04d8eb1abff6da33c77c1952d5931d7591201/pkg/controller/certapi/certificate.go#L173-L181


icco (approved):
/lgtm

PR #793, opened 2020-10-08
Remove old metrics in /api/certificate API

yegle:


yegle:
This PR is ready for review now.

icco (approved):
/lgtm
/hold

Should we wait till the other metrics have rolled out?

yegle:
If we always `tf apply` after a new binary is deployed, we should be good (new binary deployed -> new metric generated -> `tf apply` to update dashboard). If not (run `tf apply` against an old deployed binary, e.g. run it right now), we will be breaking the dashboard regardless if this PR is merged or not.

So I guess we don't need to wait?

yegle:
We've temporarily pin'ed the alerting module version in our private Terraform config repo. I'll remove the hold label now.

Issue #787, opened 2020-10-08
Allow realms to integrate with an SMTP server


whaught:
### TL;DR
Currently we use the default firebase email flow for password reset.
If we wish to use email for a 2nd factor, controler/eliminate rate-limiting concerns, and customize invitations, we need to be able to send our own emails.

### Design

**Proposal**

1. Use the golang SMTP package to send email from a custom template
2. Include a realm setting for integrating with an email server credential
3. Create a realm setting that includes a text template for inviting new users.
4. Replace our hacked SendVerificationEmail with the template and use the server SDK generate link (which has 10x quota).

**Alternatives considered**
Integrate directly with SendGrid APIs

**Additional information**
Once we have this capability we can consider using this for TOTP sent via email for 2nd factor auth along with Twilio to replace the SMS based auth.

PR #796, opened 2020-10-08
Allow custom email server for invitations

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/787

## Proposed Changes

* This PR allows for a server-level override to sending new user invitations via a custom SMTP server
* If no secret is present, it falls back to our firebase hack with the password-reset template

In the future we can use this to
* Allow per-realm overrides of the email server
* Allow per-realm templates for the invitation email itself

This gets around our server-side call to the firebase client API which may face IP rate limiting.
https://cloud.google.com/identity-platform/quotas#email_link_generation_limits_2
Note that generating the link has 10x quota and is intended to be called from the server.

We could someday send other kinds of emails (eg. TOTP for 2nd factor auth if we want to roll our own 2nd factor via this and Twilio), but that sort of thing is not currently planned.

**Release Note**

```release-note
Allow for invitations from a custom SMTP server
```

### DIFF
--- a/pkg/controller/user/reset_password.go
+++ b/pkg/controller/user/reset_password.go
@@ -49,7 +49,8 @@ func (c *Controller) ensureFirebaseUserExists(ctx context.Context, user *databas
 	}
 
 	if created {
-		if err := c.firebaseInternal.SendNewUserInvitation(ctx, user.Email); err != nil {
+		err := c.emailer.SendNewUserInvitation(ctx, user.Email)
+		if err != nil {
 			flash.Error("Could not send new user invitation: %v", err)
 			return true, err
 		}
--- a/pkg/email/config.go
+++ b/pkg/email/config.go
@@ -28,7 +28,7 @@ type ProviderType string
 const (
 	ProviderTypeNoop     ProviderType = "NOOP"
 	ProviderTypeFirebase ProviderType = "FIREBASE"
-	ProviderTypeSmtp     ProviderType = "SIMPLE_SMTP"
+	ProviderTypeSMTP     ProviderType = "SIMPLE_SMTP"
 )
 
 // Config represents the env var based configuration for email SMTP server connection.
@@ -37,8 +37,8 @@ type Config struct {
 
 	User     string `env:"EMAIL_USER" json:",omitempty"`
 	Password string `env:"EMAIL_PASSWORD" json:",omitempty"`
-	SmtpHost string `env:"EMAIL_SMTP_HOST" json:",omitempty"`
-	SmtpPort string `env:"EMAIL_SMTP_PORT" json:",omitempty"`
+	SMTPHost string `env:"EMAIL_SMTP_HOST" json:",omitempty"`
+	SMTPPort string `env:"EMAIL_SMTP_PORT" json:",omitempty"`
 
 	// Secrets is the secret configuration. This is used to resolve values that
 	// are actually pointers to secrets before returning them to the caller. The
@@ -56,8 +56,8 @@ func ProviderFor(ctx context.Context, c *Config, auth *auth.Client) (Provider, e
 	switch typ := c.ProviderType; typ {
 	case ProviderTypeFirebase:
 		return NewFirebase(ctx)
-	case ProviderTypeSmtp:
-		return NewSmtp(ctx, c.User, c.Password, c.SmtpHost, c.SmtpPort, auth)
+	case ProviderTypeSMTP:
+		return NewSMTP(ctx, c.User, c.Password, c.SMTPHost, c.SMTPPort, auth)
 	default:
 		return nil, fmt.Errorf("unknown email provider type: %v", typ)
 	}
--- a/pkg/email/config.go
+++ b/pkg/email/config.go
@@ -52,6 +52,10 @@ type Provider interface {
 	SendNewUserInvitation(ctx context.Context, email string) error
 }
 
+func (c *Config) HasSMTPCreds() bool {
+	return c.User != "" && c.Password != "" && c.SMTPHost != "" && c.SMTPPort != ""
+}
+
 func ProviderFor(ctx context.Context, c *Config, auth *auth.Client) (Provider, error) {
 	switch typ := c.ProviderType; typ {
 	case ProviderTypeFirebase:
--- a/pkg/email/config.go
+++ b/pkg/email/config.go
@@ -1,0 +1,64 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


pkg/controller/user/reset_password.go
```diff
@@ -49,7 +49,8 @@ func (c *Controller) ensureFirebaseUserExists(ctx context.Context, user *databas
 	}
 
 	if created {
-		if err := c.firebaseInternal.SendNewUserInvitation(ctx, user.Email); err != nil {
+		err := c.emailer.SendNewUserInvitation(ctx, user.Email)
+		if err != nil {
 			flash.Error("Could not send new user invitation: %v", err)
 			return true, err
```
jeremyfaller, line 55: dunno if we want to wrap errors. We're typically trying to.

pkg/email/smtp.go
```diff
@@ -0,0 +1,94 @@
...
+// SendNewUserInvitation sends a password reset email to the user.
+func (s *SmtpProvider) SendNewUserInvitation(ctx context.Context, toEmail string) error {
+	// Header
+	header := make(map[string]string)
+	header["From"] = s.User
+	header["To"] = toEmail
+	header["Subject"] = "COVID-19 Verification Server Invitation"
+
+	header["MIME-Version"] = "1.0"
+	header["Content-Type"] = fmt.Sprintf("%s; charset=\"utf-8\"", "text/html")
```
jeremyfaller, line 60: You can use the back-tick string for formats like this.

https://golang.org/ref/spec#String_literals

pkg/email/config.go
```diff
@@ -0,0 +1,64 @@
...
+	ProviderTypeSmtp     ProviderType = "SIMPLE_SMTP"
+)
+
+// Config represents the env var based configuration for email SMTP server connection.
+type Config struct {
+	ProviderType ProviderType
+
+	User     string `env:"EMAIL_USER" json:",omitempty"`
+	Password string `env:"EMAIL_PASSWORD" json:",omitempty"`
+	SmtpHost string `env:"EMAIL_SMTP_HOST" json:",omitempty"`
```
jeremyfaller, line 40: Dunno why these are separate, especially when you just combine them later....

pkg/email/smtp.go
```diff
@@ -0,0 +1,94 @@
...
+
+var _ Provider = (*SmtpProvider)(nil)
+
+// Smtp sends messages via an external SMTP server.
+type SmtpProvider struct {
+	FirebaseAuth *auth.Client
+
+	User     string
+	Password string
+	SmtpHost string
```
jeremyfaller, line 36: ditto

jeremyfaller (approved):
Nothing I've commented on here requires changing. Feel free to address in a followup if necessary.

jeremyfaller (approved):

pkg/email/config.go
```diff
@@ -0,0 +1,68 @@
...
+	ProviderTypeSMTP     ProviderType = "SIMPLE_SMTP"
+)
+
+// Config represents the env var based configuration for email SMTP server connection.
+type Config struct {
+	ProviderType ProviderType
+
+	User     string `env:"EMAIL_USER" json:",omitempty"`
+	Password string `env:"EMAIL_PASSWORD" json:",omitempty"`
+	SMTPHost string `env:"EMAIL_SMTP_HOST" json:",omitempty"`
```
sethvargo, line 40: It's probably worth noting that this _won't_ work for most modern Gmail accounts, and won't work for ~G Suite~ Google Workplace either.

pkg/email/config.go
```diff
@@ -0,0 +1,68 @@
...
+	// Secrets is the secret configuration. This is used to resolve values that
+	// are actually pointers to secrets before returning them to the caller. The
+	// table implementation is the source of truth for which values are secrets
+	// and which are plaintext.
+	Secrets secrets.Config
+}
+
+type Provider interface {
+	// SendNewUserInvitation sends an invite to join the server.
+	SendNewUserInvitation(ctx context.Context, email string) error
```
sethvargo, line 52: It would probably be better to abstract the email content from the email sending. I'd expect the email provider to only know how to `SendEmail`:

```go
type Provider interface {
  SendEmail(ctx context.Context, subject, to, cc, bcc string, body io.Reader)
}
```

pkg/email/smtp.go
```diff
@@ -0,0 +1,94 @@
...
+		SMTPHost:     host,
+		SMTPPort:     port,
+	}, nil
+}
+
+// SendNewUserInvitation sends a password reset email to the user.
+func (s *SMTPProvider) SendNewUserInvitation(ctx context.Context, toEmail string) error {
+	// Header
+	header := make(map[string]string)
+	header["From"] = s.User
```
sethvargo, line 55: We should try to make most of these constants. Both the map keys and many of their values are constant.

pkg/email/smtp.go
```diff
@@ -0,0 +1,94 @@
...
+	header["Content-Transfer-Encoding"] = "quoted-printable"
+
+	headerMessage := ""
+	for key, value := range header {
+		headerMessage += fmt.Sprintf("%s: %s\r\n", key, value)
+	}
+
+	inviteLink, err := s.FirebaseAuth.PasswordResetLink(ctx, toEmail)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
sethvargo: bump

pkg/email/smtp.go
```diff
@@ -0,0 +1,94 @@
...
+// SendNewUserInvitation sends a password reset email to the user.
+func (s *SMTPProvider) SendNewUserInvitation(ctx context.Context, toEmail string) error {
+	// Header
+	header := make(map[string]string)
+	header["From"] = s.User
+	header["To"] = toEmail
+	header["Subject"] = "COVID-19 Verification Server Invitation"
+
+	header["MIME-Version"] = "1.0"
+	header["Content-Type"] = `text/html; charset="utf-8"`
```
sethvargo, line 60: I'm not sure we want to send as text/html. The firebase ones are text/plain.

pkg/email/config.go
whaught: Wait... but it does. I'm running encvtest@gmail.com to test this

sethvargo: You don't have MFA enabled then? It won't work with advanced protection accounts either.

whaught: https://support.google.com/mail/answer/185833

With MFA you need to create an app password for this to use

sethvargo: > App Passwords aren’t recommended and are unnecessary in most cases.

> If you don’t have this option, it might be because:
>
> c. Your account is through work, school, or other organization.
> d. You turned on Advanced Protection.

c is very likely. d will probably become a default in the future.

whaught: https://github.com/google/exposure-notifications-verification-server/pull/797/commits/72cc99b5a91e929a91aa285b6b1b3aaeb4c338c7

Adding in some notes

sethvargo: Thank

PR #797, opened 2020-10-09
Generate invitation email from template

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/787

## Proposed Changes

* Email template for invites
* Default smtp port
* logs some errors
* noop email sender


pkg/email/config.go
```diff
@@ -56,12 +56,14 @@ func (c *Config) HasSMTPCreds() bool {
 	return c.User != "" && c.Password != "" && c.SMTPHost != "" && c.SMTPPort != ""
 }
 
-func ProviderFor(ctx context.Context, c *Config, auth *auth.Client) (Provider, error) {
+func ProviderFor(ctx context.Context, c *Config, assetsRoot string, auth *auth.Client) (Provider, error) {
 	switch typ := c.ProviderType; typ {
+	case ProviderTypeNoop:
+		return NewNoop()
```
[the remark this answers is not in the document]
whaught: It's currently not referenced anywhere

pkg/email/smtp.go
```diff
@@ -31,20 +31,28 @@ var _ Provider = (*SMTPProvider)(nil)
...
 	User     string
 	Password string
 	SMTPHost string
 	SMTPPort string
 }
 
 // NewSMTP creates a new Smtp email sender with the given auth.
-func NewSMTP(ctx context.Context, user, password, host, port string, auth *auth.Client) (Provider, error) {
+func NewSMTP(ctx context.Context, user, password, host, port, assetsRoot string, auth *auth.Client) (Provider, error) {
+	t, err := template.ParseFiles(assetsRoot + "/email/default_invitation.html")
```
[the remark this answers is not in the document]
whaught: at the moment the renderer is intended for html and takes a response writer, deals with error codes and headers. It's a good idea, but I think we'd want to write another email_renderer.go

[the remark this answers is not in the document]
whaught: Added the renderer

cmd/server/assets/email/default_invitation.html
```diff
@@ -0,0 +1,22 @@
+{{define "email/invite"}}
+{{- /* Message Header */ -}}
+Subject: COVID-19 Verification Server Invitation
+To: {{.ToEmail}}
+From: {{.FromEmail}}
+MIME-Version: 1.0
+Content-Type: text/html; charset="utf-8"
+hContent-Disposition: inline
```
sethvargo, line 8:
```suggestion
Content-Disposition: inline
```

cmd/server/assets/email/default_invitation.html
```diff
@@ -0,0 +1,22 @@
+{{define "email/invite"}}
+{{- /* Message Header */ -}}
+Subject: COVID-19 Verification Server Invitation
+To: {{.ToEmail}}
+From: {{.FromEmail}}
+MIME-Version: 1.0
+Content-Type: text/html; charset="utf-8"
```
sethvargo, line 7: If we set this as `text/plain`, you can remove the `<br>` and the email will "just work".

pkg/controller/user/importbatch.go
```diff
@@ -74,7 +75,8 @@ func (c *Controller) HandleImportBatch() http.Handler {
 			} else if created {
 				newUsers = append(newUsers, &batchUser)
 				if err := c.emailer.SendNewUserInvitation(ctx, user.Email); err != nil {
```
sethvargo, line 77: Previously, this was a non-blocking operation (calling the firebase API). However, now it's blocking. SMTP sending is happening inline with the request, and SMTP is blocking. We either need to so something primitive (like push this into a background go routine and hope it finishes before the instance spins down) or build a job processing queue.

pkg/email/config.go
```diff
@@ -20,25 +20,32 @@ import (
...
 	ProviderType ProviderType
 
-	User     string `env:"EMAIL_USER" json:",omitempty"`
-	Password string `env:"EMAIL_PASSWORD" json:",omitempty"`
-	SMTPHost string `env:"EMAIL_SMTP_HOST" json:",omitempty"`
-	SMTPPort string `env:"EMAIL_SMTP_PORT" json:",omitempty"`
+	User     string `env:"EMAIL_USER"`
+	Password string `env:"EMAIL_PASSWORD"`
+	SMTPHost string `env:"EMAIL_SMTP_HOST"`
+	SMTPPort string `env:"EMAIL_SMTP_PORT, default=587"`
```
sethvargo, line 48: It might be worth commenting (somewhere) that port 25 is blocked on GCP and most other places.

pkg/email/smtp.go
```diff
@@ -31,62 +30,60 @@ var _ Provider = (*SMTPProvider)(nil)
...
 		User:         user,
 		Password:     password,
 		SMTPHost:     host,
 		SMTPPort:     port,
-	}, nil
+	}
 }
 
 // SendNewUserInvitation sends a password reset email to the user.
 func (s *SMTPProvider) SendNewUserInvitation(ctx context.Context, toEmail string) error {
```
sethvargo, line 54: See my note in the other PR, but I think this interface is too coupled to the SMTP provider. SMTP should _only_ know how to send mail.

sethvargo (changes requested):

pkg/controller/user/importbatch.go
whaught: Done as a simple goroutine. I think the firebase operation is still a blocking call? These seem to be pretty quick

sethvargo: Firebase was a blocking HTTP call, but it queued the email. The HTTP response didn't depend on SMTP.

cmd/server/assets/email/_plainheader.html
```diff
@@ -0,0 +1,9 @@
+{{- define "email/plainheader" -}}
+Subject: COVID-19 Verification Server Invitation
```
sethvargo, line 2: Can be a follow up, but I think we want to say "Exposure Notifications", not "COVID"

sethvargo (approved):
/lgtm

PR #800, opened 2020-10-12
Changing some words

whaught:
## Proposed Changes

* This PR is intended for folks to look at the wording of our invitation. Looking for reviewer feedback.


cmd/server/assets/email/_plainheader.html
```diff
@@ -1,5 +1,5 @@
 {{- define "email/plainheader" -}}
-Subject: COVID-19 Verification Server Invitation
+Subject: Exposure Notification Verification Server Invite
```
sethvargo, line 2:
```suggestion
Subject: Your Exposure Notifications Verification Server Account
```

cmd/server/assets/email/default_invite.html
```diff
@@ -9,5 +9,5 @@
 {{.InviteLink}}
 
 Thanks,
-The encv team
```
sethvargo, line 12: I think we should omit the salutation entirely.

sethvargo (changes requested):

cmd/server/assets/email/default_invite.html
```diff
@@ -9,5 +9,5 @@
...
+The ENCV team
```
mikehelmick, line 12: this needs to be configurable - ENCV is the URL in the US.

Ideally this would be realm specific

cmd/server/assets/email/_plainheader.html
mikehelmick: ideally - something realm specific

cmd/server/assets/email/default_invite.html
whaught: The idea is to allow realms to override entirely. This is just the default

mikehelmick (approved):

PR #801, opened 2020-10-12
Fix realm token related metrics in /api/issue.

yegle:
Prior to this change, the # of used tokens are calculated by:

    issued := uint64(limit) - remaining

This calculation doesn't take burst tokens into consideration. Imaging
the limit is set to 10, and at a time someone added 1000 burst into the
bucket, and the actual used token is 10. We now have limit=10,
remaining=1000, and the calculated issued=-1000, which doesn't make
sense.

The used token should be a CUMULATIVE/view.Count metric, which is
recorded each time we successfully called `limiter.Take()`.


whaught (approved):
/lgtm

PR #802, opened 2020-10-12
Also modify how /api/issue/realm_token_latest is recorded.

yegle:
Since we record the used tokens in a separate metric, we can now keep
using /api/issue/realm_token_latest for the purpose of getting metrics
from the limiter.

We can now calculate the capacity of the limiter by:

    generic_task ::
        'custom.googleapis.com/opencensus/en-verification-server/api/issue/realm_token_latest'
    | {
      filter metric.state == "AVAILABLE";
      filter metric.state == "LIMIT"
    }
    | [metric.realm]
    | ratio
    | value if(val() > 1.0, 1.0, val())


icco:
/lgtm
/hold

This will break the alert won't it?

yegle:
> /lgtm
> /hold
> 
> This will break the alert won't it?

No, alert is not using this metric yet.

PR #803, opened 2020-10-12
Fixing a nil pointer in e2e

whaught:
## Proposed Changes

* I was paged for a nil pointer here.
    * Is there additional action we should take in this scenario?

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -155,6 +155,7 @@ func realMain(ctx context.Context) error {
 		app, err := db.FindAuthorizedAppByAPIKey(deviceKey)
 		if err != nil {
 			logger.Errorf("device API key cleanup failed: %w", err)
+			return
 		}
 		now := time.Now().UTC()
 		app.DeletedAt = &now


yegle:
The key clean up is best effort. We should be fine with keys left in db.

/lgtm

jeremyfaller (approved):

cmd/e2e-runner/main.go
```diff
@@ -155,6 +155,7 @@ func realMain(ctx context.Context) error {
 		app, err := db.FindAuthorizedAppByAPIKey(deviceKey)
 		if err != nil {
 			logger.Errorf("device API key cleanup failed: %w", err)
+			return
```
chaodaiG, line 158: Is this best effort? Should this be a panic instead?

whaught: Yes. Best-effort according to @yegle. I'm not sure if we want to panic - if this were a concern we could consider an async db cleanup later perhaps

yegle: I was wondering why this log looks like:

```device API key cleanup failed: %!w(*net.OpError=&{read tcp 0xc000371b90 0xc00040f2f0 0xc0006a3760})```

PR #814, opened 2020-10-13
Update/deduplicate metrics in the cleanup handler.

yegle:
The old metrics are not used anywhere, delete old ones as well.

Part of #772


mikehelmick (approved):

pkg/controller/cleanup/metrics.go
```diff
@@ -20,104 +20,40 @@ import (
...
+	mClaimRequests = stats.Int64(metricPrefix+"/claim_requests", "The number of cleanup claim requests.", stats.UnitDimensionless)
+)
 
-	// Audit entries
-	mPurgeAuditEntriesAttempts = stats.Int64(metricPrefix+"/purge_audit_entries_attempts", "The number of attempts to purge audit entries", stats.UnitDimensionless)
-	mPurgeAuditEntriesErrors   = stats.Int64(metricPrefix+"/purge_audit_entries_errors", "The number of attempts to purge audit entries", stats.UnitDimensionless)
-	mPurgeAuditEntriesPurged   = stats.Int64(metricPrefix+"/purge_audit_entries_purged", "The number of audit entries that were purged", stats.UnitDimensionless)
+var (
+	// itemTagKey indicating what type of items is cleaned up in this step.
+	// Potential values:
```
sethvargo, line 35: I'm slightly worried about this approach as it wouldn't catch typo until the metric hit prod (e.g. `VERIFICATON_CODE`). Is there a way we can validate the allowed list somehow?

pkg/controller/issueapi/issue.go
```diff
@@ -308,14 +308,13 @@ func (c *Controller) HandleIssue() http.Handler {
...
-				stats.RecordWithTags(ctx, []tag.Mutator{observability.APIResultError("NOT_OK")}, mSMSRequest.M(1))
+				stats.RecordWithTags(ctx, []tag.Mutator{observability.ResultError("NOT_OK")}, mSMSRequest.M(1))
 				c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to send sms"))
 				blame = observability.BlameServer
-				result = observability.APIResultError("FAILED_TO_SEND_SMS")
+				result = observability.ResultError("FAILED_TO_SEND_SMS")
 				return
 			}
-			stats.RecordWithTags(ctx, []tag.Mutator{observability.APIResultOK()}, mSMSRequest.M(1))
+			stats.RecordWithTags(ctx, []tag.Mutator{observability.ResultOK()}, mSMSRequest.M(1))
```
sethvargo, line 317: Is it intended that there's no "item" here?

pkg/controller/cleanup/metrics.go
yegle: We should start by alerting on the over all error ratio. From my day job experience, further breakdown to a particular error is usually not very useful.

We can set up a chart that breaks down to detailed errors, but that is used by the oncaller during alert troubleshooting.

FWIW the alert policy can also set up a "data point has been missing for x min" condition so we should be able to catch typos in the filter.

pkg/controller/issueapi/issue.go
yegle: Yeah this is a metric specifically for SMS requests.

PR #822, opened 2020-10-14
Remove quota related metrics.

yegle:
The metric is already recorded as part of /requests_count metric.

Part of #773


sethvargo (approved):
/lgtm

PR #825, opened 2020-10-14
Make alert names consistent

icco:
Part of #824

## Proposed Changes

* Makes all alert names space seperated.

**Release Note**

```release-note
Changes two alert names to have spaces in them.
```


mikehelmick (approved):

yegle (commented):
Hmm I was going to rename them so there's no space in the displayName :-(

PR #828, opened 2020-10-14
Add a per-realm-per-user chart on the stats page.

### LINKED ISSUE #799: user-issue statistics
Currently there are statistics per realm and statistics per user (if you navigate to the user)

Request:
Show codes issued by user - possibly over time, possibly per day.

/kind enhancement

jeremyfaller:
Fixes #799

NB: We display the user's name, **NOT** their ID. The bug was ill-specified. If this is wrogn, please comment.

## Proposed Changes

* Add a per-user-per-realm stats chart
image

**Release Note**

```release-note
Stats page now contains a per-user-per-realm chart.
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -62,6 +62,7 @@ func (t TestType) Display() string {
 
 var (
 	ErrNoSigningKeyManagement = errors.New("no signing key management")
+	ErrBadDateRange           = errors.New("bad date range")
 )
 
 const (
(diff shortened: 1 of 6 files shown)


pkg/database/realm.go
```diff
@@ -1159,3 +1160,39 @@ func ToCIDRList(s string) ([]string, error) {
...
+	}
+
+	var stats []*RealmUserStats
+	if err := db.db.
+		Model(&UserStats{}).
+		Select("users.id, users.name, codes_issued, date").
+		Where("realm_id = ?", r.ID).
+		Where("date >= ? AND date <= ?", start, stop).
+		Joins("INNNER JOIN users ON users.id = user_id").
+		Order("date ASC").
```
sethvargo, line 1188: Did you verify this actually works? I've had weird issues with Gorm where the order call has to come before select/where.

pkg/database/realm.go
```diff
@@ -1159,3 +1160,39 @@ func ToCIDRList(s string) ([]string, error) {
...
+		Scan(&stats).
+		Error; err != nil {
+		if IsNotFound(err) {
+			fmt.Println("NOT FOUND")
```
sethvargo, line 1192: Please remove

pkg/database/user_stats.go
```diff
@@ -30,3 +32,14 @@ type UserStats struct {
 func (UserStats) TableName() string {
 	return "user_stats"
 }
+
+// Save saves some UserStats to the database.
+// This function is provided for testing only.
+func (u *UserStats) Save(db *Database) error {
```
sethvargo, line 38: Is it too painful to call `db.SaveStats(stat)` - it's more in line with our current pattern to put the `SaveX` on the database instead of on the model.

jeremyfaller: No, it's not too painful, but that function doesn't exist. Changed this function over to that format.

pkg/database/realm.go
jeremyfaller: my mistake, done.

pkg/database/realm.go
jeremyfaller: Added to test case. PTAL

pkg/database/realm.go
```diff
@@ -1159,3 +1160,38 @@ func ToCIDRList(s string) ([]string, error) {
...
+// This is a structure joined from multiple tables in the DB.
+type RealmUserStats struct {
+	UserID      uint
+	Name        string
+	CodesIssued uint
+	Date        time.Time
+}
+
+// CodesPerUser returns a set of UserStats for a given date range.
+func (r *Realm) CodesPerUser(db *Database, start, stop time.Time) ([]*RealmUserStats, error) {
```
mikehelmick, line 1174: I think there's a subtle issue here w/ removing a user from a realm.

we likely need to add tracking to "soft-delete" a user from a realm until they roll off of the stats.

mikehelmick (approved):

Issue #810, opened 2020-10-13
Convert our alerts to MQL


yegle:
There are certain alerts that we want to have cannot be created from the UI. E.g. we want to have an alert on realm token capacity like this:

```
generic_task ::
  custom.googleapis.com/opencensus/en-verification-server/api/issue/realm_token_latest
| {
      filter metric.state == 'AVAILABLE';
      filter metric.state == 'LIMIT'
}
| ratio
| value if(val() > 1.0, 1.0, val())
| window 10m
| condition val() < 0.1
```

This depends on upstream bug https://github.com/hashicorp/terraform-provider-google/issues/7464

PR #831, opened 2020-10-14
Add RealmTokenRemainingCapacityLow alert.

yegle:
This alert is currently set up using a local-exec provisioner, due to
the lack of support to define alert policies using MQL in Terraform.

Part of #810

### DIFF
--- a/terraform/alerting/verification-server/alerts.tf
+++ b/terraform/alerting/verification-server/alerts.tf
@@ -1,0 +1,35 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


sethvargo (commented):
This is putting a pretty hard dependency on gcloud and the alpha components. Do you have an idea of when it'll be upstream?

yegle:
I reached out to the team internally and there's currently no timeline on when it will be implemented.

Depending on `gcloud` is less of an issue since you have to run `gcloud` to authenticate before running terraform. Depending on `gcloud alpha` however is indeed a bit of a concern.

sethvargo:
> Depending on gcloud is less of an issue since you have to run gcloud to authenticate before running terraform.

That's not entirely true - you can run Terraform by downloading a service account from the console and setting an envvar.

yegle:
> > Depending on gcloud is less of an issue since you have to run gcloud to authenticate before running terraform.
> 
> That's not entirely true - you can run Terraform by downloading a service account from the console and setting an envvar.

True...

Anyway, would it be possible to accept this as a workaround, until it's officially supported? I imaging most of our alerts will still be defined using Terraform supported features.

sethvargo:
I personally feel that the toil of taking on and maintaining this exceeds that which we get by waiting and just creating them manually for now, but I'm just one opinion. My main concern is that there's a lot of people depending on our Terraform configs, and many of them are not very familiar. This is going to dramatically complicate their installations if they use the alerting module.

yegle:
I see https://github.com/hashicorp/terraform-provider-google/pull/7568 was created yesterday and this would hopefully get landed soon.

I'll close this PR.

Issue #818, opened 2020-10-14
Cannot deploy e2e-runner


sethvargo:
The e2e-runner in prod fails to deploy with:

```text
"error trying to create a new Admin API Key: failed to save API key: pq: duplicate key value violates unique constraint "realm_apikey_name""
```

This is an intermittent failure, but blocks deployment.

/cc @chaodaiG

chaodaiG:
There are two DB columns that have this constraint: `RealmID ` and `Name`. Since `Name` is generated by [crypto/rand](https://github.com/google/exposure-notifications-verification-server/blob/c47a42cf4b63d530b9637e1991c616e98bcdbe15/cmd/e2e-runner/main.go#L120) it's highly unlikely innocent based on how often the problem appeared

Link: https://github.com/google/exposure-notifications-verification-server/blob/3b3e584551cd772198137a9f7168797ea7504bc6/pkg/database/authorized_app.go#L67

chaodaiG:
I can take a stab fixing this later today, if my hypothesis is correct

chaodaiG:
The hypothesis above was incorrect, the collision was still on `Name`. It's likely due to the fact that the random suffix was created with an int below 10000, and without cleanup the chance of collision increase over time

PR #832, opened 2020-10-14
e2e test runner random suffix 32 chars instead of picking an int under 10000

chaodaiG:
Fixes: https://github.com/google/exposure-notifications-verification-server/issues/818


chaodaiG:
Not quite sure whether this is the most proper fix. A better way in my opinion would be reusing authorized app between runs. Doing this means either expose an API deleting authorized app or directly deleting from DB

sethvargo (approved):
/lgtm

PR #833, opened 2020-10-15
Move email send out to controller

whaught:
## Proposed Changes

* Moves the composition of emails out of the email provider and out to the controller
    * Now emailer only knows how to send mail. This divorces it from what its sending and allows for better dependency patterns. This is a bit of a prerequisite to per-realm overrides as with SMS.


pkg/controller/user/importbatch.go
```diff
@@ -105,3 +106,30 @@ func (c *Controller) HandleImportBatch() http.Handler {
 		c.h.RenderJSON(w, http.StatusOK, response)
 	})
 }
+
+func (c *Controller) sendInvitation(ctx context.Context, toEmail string) error {
+	realm := controller.RealmFromContext(ctx)
+
+	// Fallback to Firebase
```
sethvargo, line 113: Nit: logically, it might make sense to put this at the end (and change the conditional to != nil, since it's the "fallback"

pkg/email/config.go
```diff
@@ -65,8 +59,10 @@ type Config struct {
 
 // Provider is an interface for email-sending mechanisms.
 type Provider interface {
-	// SendNewUserInvitation sends an invite to join the server.
-	SendNewUserInvitation(ctx context.Context, email string) error
+	// SendEmail sends an email with the given message.
+	SendEmail(ctx context.Context, toEmail string, message []byte) error
+
+	From() string
```
sethvargo, line 65: Doc

sethvargo (approved):
/lgtm

PR #834, opened 2020-10-15
Give system admins ability to clear caches

### LINKED ISSUE #821: Let admins clear caches via the UI
Redis is only accessible via a bastion host on the VPC, so clearing the cache is painful. While it should be unnecessary, let's add some buttons for system admins to clear caches.

sethvargo:
In some situations, a system admin may want to clear the caches, either
for development/testing or because a misconfiguration has made it into
the system. This introduces a way for system admins to perform cache
resets.

This required a bit of a refactor of the cacher, separating the
namespace property from the actual key value.

Fixes #821 

**Release Note**

```release-note
Give system admins the ability to clear (Redis) caches.
```

### DIFF
--- a/pkg/cache/in_memory.go
+++ b/pkg/cache/in_memory.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 )
@@ -73,15 +74,12 @@ func NewInMemory(i *InMemoryConfig) (Cacher, error) {
 // returns the value. If the value does not exist, it calls f and caches the
 // result of f in the cache for ttl. The ttl is calculated from the time the
 // value is inserted, not the time the function is called.
-func (c *inMemory) Fetch(_ context.Context, key string, out interface{}, ttl time.Duration, f FetchFunc) error {
+func (c *inMemory) Fetch(_ context.Context, k *Key, out interface{}, ttl time.Duration, f FetchFunc) error {
 	now := time.Now().UnixNano()
 
-	if c.keyFunc != nil {
-		var err error
-		key, err = c.keyFunc(key)
-		if err != nil {
-			return fmt.Errorf("failed to execute keyFunc: %w", err)
-		}
+	key, err := k.Compute(c.keyFunc)
+	if err != nil {
+		return fmt.Errorf("failed to compute key: %w", err)
 	}
 
 	// Try a read-only lock first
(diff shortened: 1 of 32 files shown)


pkg/cache/in_memory.go
```diff
@@ -199,26 +191,41 @@ func (c *inMemory) Read(_ context.Context, key string, out interface{}) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to compute key: %w", err)
 	}
 
 	delete(c.data, key)
 	return nil
 }
 
+// DeletePrefix removes all items that start with the given prefix.
+func (c *inMemory) DeletePrefix(_ context.Context, prefix string) error {
```
whaught, line 212: Why the unused context?

pkg/controller/admin/caches.go
```diff
@@ -0,0 +1,84 @@
...
+	Description string
+}
+
+var caches = map[string]*cacheItem{
+	"authorized_apps:": {"API keys", "Authentication caches for mobile devices"},
+	"apps:":            {"Mobile apps", "Registered mobile apps for the redirector service"},
+	"jwks:":            {"JWKs", "JSON web tokens"},
+	"realms:":          {"Realms", "Caches of realm data"},
+	"stats:":           {"Statistics", "API key, user, and realm statistic caches"},
+	"publickeys:":      {"Public keys", "PEM data from upstream key provider"},
```
whaught, line 36: should this be public_keys?

pkg/cache/in_memory.go
whaught: jk found the later impl that uses it

pkg/controller/admin/caches.go
```diff
@@ -0,0 +1,84 @@
...
+
+type cacheItem struct {
+	Name        string
+	Description string
+}
+
+var caches = map[string]*cacheItem{
+	"authorized_apps:": {"API keys", "Authentication caches for mobile devices"},
+	"apps:":            {"Mobile apps", "Registered mobile apps for the redirector service"},
+	"jwks:":            {"JWKs", "JSON web tokens"},
```
mikehelmick, line 33: description should be `JSON Web Key Set`

pkg/controller/admin/caches.go
```diff
@@ -0,0 +1,84 @@
...
+	"realms:":          {"Realms", "Caches of realm data"},
```
mikehelmick, line 34: probably should remove the word cache from all of these descriptions.

pkg/controller/admin/caches.go
sethvargo: The existing cache is `publickeys`. I could update them to all be `public_keys` pretty easily. Just kept it as-is for diffs sake.

pkg/controller/admin/caches.go
sethvargo: Done.

pkg/controller/admin/caches.go
sethvargo: Done.

pkg/controller/admin/caches.go
sethvargo: Ah it was only two places.

sethvargo:
@whaught @mikehelmick PTAnotherL

whaught (approved):

PR #846, opened 2020-10-16
Allow blank entry for URL.

flagxor:
Allow App URL to be blank on entry (rather than ignore blank).


sethvargo (approved):
/lgtm

mikehelmick:
This was unnecessary - AfterFind turns a nil pointer bdck into and ekpty string

PR #856, opened 2020-10-19
Fix errorcode conversion to string

jeremyfaller:


whaught (approved):

pkg/controller/user/importbatch.go
```diff
@@ -96,7 +96,7 @@ func (c *Controller) HandleImportBatch() http.Handler {
 
 		if err := batchErr.ErrorOrNil(); err != nil {
 			response.Error = err.Error()
-			response.ErrorCode = string(http.StatusInternalServerError)
+			response.ErrorCode = fmt.Sprintf("%d", http.StatusInternalServerError)
```
mikehelmick, line 99: should this be https://golang.org/pkg/net/http/#StatusText ?

jeremyfaller: Possibly, I don't know the API well. I was just fixing an error that running the tests with `go test ./...` found. I'm happy to address with further PRs.

Issue #857, opened 2020-10-19
Update to gorm v2


sethvargo:
### TL;DR
We're currently using [gorm v1](https://v1.gorm.io/). The latest is [gorm v2(https://gorm.io/), but there are pretty [substantial changes](https://gorm.io/docs/v2_release_note.html).

This issue attempts to highlight the changes we'd need to make in order to migrate to v2.

### Design

**Proposal**

1. Update all imports:

    ```text
    github.com/jinzhu/gorm -> gorm.io/gorm
    github.com/jinzhu/gorm/dialects/postgres -> ??
    ```

1. Update [gomigrate to v2](https://github.com/golang-migrate/migrate)

1. `DeletedAt` is now a custom field type instead of `*time.Time`. We need to replace all instances where we set `DeletedAt` to `nil` or `&time.Now()` to this new custom field (untested):

    ```diff
    -now := time.Now().UTC()
    -app.DeletedAt = &now
    +app.DeletedAt = gorm.DeletedAt{
    +  Time:  time.Now().UTC(),
    +  Valid: true,
    +}
    ```

1. `.Related` is no longer a thing. Instead you have to declare the association and then the target:

    ```diff
    -Related(&authApps).
    -Error; err != nil {
    +Association("AuthorizedApps").
    +Find(&authApps); err != nil {
    ```

1. Calling `.Error` is now even less intuitive. You no longer call `.Error()` on associations, but still call it on queries. That means:

    ```go
    db.Model(&User{}).First(&user).Error
    ```

    is valid, but

    ```go
    db.Model(&User{}).Association("Profile").First(&profile).Error
    ```

    because associations now return an error instead of the tx. It's actually worse than that though, because `.Error` is a _valid_ method to call on an `error`, meaning we'll need to do a full audit of the codebase since we typically wrap an error before returning.

1. `db.DB()` used to return the underlying db driver. It still does, but now it returns `(db, error)`, which is fine, but annoying to update everywhere. It also means we'll need to update signatures on several functions to return errors.

1. `gorm.IsRecordNotFoundError` is gone. The docs _claim_ you can use `errors.Is`, but I'm skeptical.

1. Gorm callbacks no longer include a scope. Instead you have to get the scope from the tx from the statement, all of which can be nil.

[...]

PR #858, opened 2020-10-19
Timebox 30min upgrading to gorm v2

sethvargo:
See GH-857

⚠️ 🛑 this is not designed to be merged. it's designed to be looked at. 🛑 ⚠️


sethvargo:
We talked about this in the group chat today and decided to carry on with gorm v1.

PR #860, opened 2020-10-20
e2e test script is guarded against prod resources

chaodaiG:
There are 2 scripts used for running e2e tests: `scripts/e2e-tests.sh` and `scripts/terraform.sh`, the former is a thin wrapper of the latter. Adding guard to the latter in this PR to avoid it from being used on prod resource

### DIFF
--- a/scripts/terraform.sh
+++ b/scripts/terraform.sh
@@ -42,7 +42,7 @@ readonly PROTECTED_DB_INSTANCE_NAMES="en-verification en-server"
 LIST_PROJECTED_DB="$(gcloud sql instances list --project=${PROJECT_ID} --filter="name=( ${PROTECTED_DB_INSTANCE_NAMES} )")"
 if [[ -n "${LIST_PROJECTED_DB}" ]]; then
   # The output will only exist when the database exist
-  echo "✋ Running this script if prohibited when database below exist:"
+  echo "✋ Running this script is prohibited when database below exist:"
   echo "${LIST_PROJECTED_DB}"
   exit 100
 fi
--- a/scripts/terraform.sh
+++ b/scripts/terraform.sh
@@ -23,6 +23,31 @@ if [[ -z "${PROJECT_ID:-}" ]]; then
   echo "✋ PROJECT_ID must be set"
 fi
 
+# Ensure not running on prod resources
+readonly PROTECTED_PROJECT_IDS=(
+  "apollo-server-273118"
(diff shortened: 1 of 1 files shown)


scripts/terraform.sh
```diff
@@ -23,6 +23,31 @@ if [[ -z "${PROJECT_ID:-}" ]]; then
...
+    echo "✋ Running this script on prod servers is prohibited"
+    exit 100
+  fi
+done
+
+readonly PROTECTED_DB_INSTANCE_NAMES="en-verification en-server"
+LIST_PROJECTED_DB="$(gcloud sql instances list --project=${PROJECT_ID} --filter="name=( ${PROTECTED_DB_INSTANCE_NAMES} )")"
+if [[ -n "${LIST_PROJECTED_DB}" ]]; then
+  # The output will only exist when the database exist
+  echo "✋ Running this script if prohibited when database below exist:"
```
whaught, line 45:
```suggestion
  echo "✋ Running this script is prohibited when database below exist:"
```

scripts/terraform.sh
```diff
@@ -23,6 +23,31 @@ if [[ -z "${PROJECT_ID:-}" ]]; then
...
+readonly PROTECTED_PROJECT_IDS=(
+  "apollo-server-273118"
+  "apollo-server-us"
+  "apollo-verification-us"
+  "encv-prod"
+  "encv-test"
+)
+for protected_project_id in ${PROTECTED_PROJECT_IDS[@]}; do
+  if [[ "${protected_project_id}" == "${PROJECT_ID}" ]]; then
+    echo "✋ Running this script on prod servers is prohibited"
```
whaught, line 36: we could mention. "this script destroys resources"

scripts/terraform.sh
```diff
@@ -23,6 +23,31 @@ if [[ -z "${PROJECT_ID:-}" ]]; then
   echo "✋ PROJECT_ID must be set"
 fi
 
+# Ensure not running on prod resources
+readonly PROTECTED_PROJECT_IDS=(
```
icco, line 27: Should we do the inverse and have an allow list instead?

chaodaiG: That would be much safer, but also means a larger list to maintain

icco: 👍

icco (commented):
/lgtm

PR #862, opened 2020-10-20
E2E test: update usage of gcloud to prevent breakage from future breaking API changes

chaodaiG:
`gcloud --filter` will change its `name=value` behavior to regex matching in future updates, and recommending `name<=value AND name>=value` if desire an equality checking. See: https://cloud.google.com/sdk/gcloud/reference/topic/filters

Updating the usage to prevent future breakages

### DIFF
--- a/scripts/terraform.sh
+++ b/scripts/terraform.sh
@@ -44,18 +44,17 @@ readonly PROTECTED_DB_INSTANCE_NAMES=(
   "en-verification"
   "en-server"
 )
-for protected_db_instance_name in ${PROTECTED_DB_INSTANCE_NAMES[@]}; do
-  # gcloud --filter will change its `name=value` behavior to regex matching, and recommending
-  # `name<=value AND name>=value` if desire an equality checking. See:
-  # https://cloud.google.com/sdk/gcloud/reference/topic/filters
-  LIST_PROTECTED_DB="$(gcloud sql instances list --project=${PROJECT_ID} --filter="(name <= ${protected_db_instance_name}) AND (name >= ${protected_db_instance_name})")"
-  if [[ -n "${LIST_PROTECTED_DB}" ]]; then
-    # The output will only exist when the database exist
-    echo "✋ Running this script is prohibited when database below exist:"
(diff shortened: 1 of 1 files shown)


yegle:
`key = (value)` is more readable it seems.

chaodaiG:
> `key = (value)` is more readable it seems.

Totally agree. It has the same warning when I use it too

yegle:
> > `key = (value)` is more readable it seems.
> 
> Totally agree. It has the same warning when I use it too

Bummer :-(

Reading the doc it seems like the intention was "to phase out support of using `=` as if it's `:`". I'd suggest to add a TODO to revert it back to simple `=` once that behavior is removed.

scripts/terraform.sh
```diff
@@ -40,15 +40,23 @@ for protected_project_id in ${PROTECTED_PROJECT_IDS[@]}; do
...
-fi
+readonly PROTECTED_DB_INSTANCE_NAMES=(
+  "en-verification"
+  "en-server"
+)
+for protected_db_instance_name in ${PROTECTED_DB_INSTANCE_NAMES[@]}; do
+  # gcloud --filter will change its `name=value` behavior to regex matching, and recommending
+  # `name<=value AND name>=value` if desire an equality checking. See:
+  # https://cloud.google.com/sdk/gcloud/reference/topic/filters
+  LIST_PROTECTED_DB="$(gcloud sql instances list --project=${PROJECT_ID} --filter="(name <= ${protected_db_instance_name}) AND (name >= ${protected_db_instance_name})")"
```
sethvargo, line 51: Two things:

1. If we stick with the filter, I'd rather `^...$` than use the ><.

2. I think it'd be more efficient to just list all of them for the project, then use grep instead of relying on gcloud output. That also saves us spawning gcloud N times.

sethvargo (changes requested):

scripts/terraform.sh
chaodaiG: Updated with ideas from 2

sethvargo (approved):

Issue #864, opened 2020-10-21
Set up cron job to backup database


jeremyfaller:
After we had the issues upgrading encv-test on Oct 20, we realized we do not have a backup of the database if the instance is deleted. We need a better system in place should the worst happen.

Related to post-mortem 171305680

PR #875, opened 2020-10-21
Backup the Cloud SQL database to GCS every 6 hours

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/864

**Nuances**

- Cloud SQL doesn't seem to support per-instance IAM, so we have to grant at the project level (yuck, but it's only viewer)

- I'm relying on bucket versioning instead of actually naming the exports with a datetime. This is mostly a design decision to avoid needing a middle service. Ideally Cloud Scheduler would support a template syntax, but it does not.

- The backups aren't compressed (for the same reason as above), but our last backup was 2MB. I'm not really worried about large databases for this project.

- Backup time is configurable via the cron notation.

**Release Note**

```release-note
Automatically backup the Cloud SQL database to GCS every 6 hours
```

/assign @jeremyfaller @icco

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -298,6 +298,10 @@ resource "google_storage_bucket" "backups" {
     enabled = true
   }
 
+  retention_policy {
+    retention_period = "2592000" // 30 days in seconds
+  }
+
   lifecycle_rule {
     action {
       type = "Delete"
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -285,6 +285,95 @@ resource "null_resource" "migrate" {
   ]
 }
 
+# Create a storage bucket where database backups will be housed.
+resource "google_storage_bucket" "backups" {
+  project  = var.project
+  name     = "${var.project}-backups"
+  location = var.storage_location
+
+  force_destroy               = true
+  uniform_bucket_level_access = true
+
+  versioning {
+    enabled = true
+  }
+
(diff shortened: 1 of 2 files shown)


mikehelmick (changes requested):
suggestion - we add GCS retention timeline on the bucket as well

mikehelmick (approved):

terraform/database.tf
```diff
@@ -285,6 +285,95 @@ resource "null_resource" "migrate" {
   ]
 }
 
+# Create a storage bucket where database backups will be housed.
+resource "google_storage_bucket" "backups" {
+  project  = var.project
+  name     = "${var.project}-backups"
+  location = var.storage_location
+
+  force_destroy               = true
```
jeremyfaller, line 294: feels safer to me to have this be false(the default).

icco (approved):
/lgtm

jeremyfaller (commented):
lgtm, I'll let @icco finish here.

terraform/database.tf
sethvargo: This tells terraform to delete the objects in the bucket before deleting the bucket. GCS doesn't let you delete a bucket with objects inside.

jeremyfaller: Yeah, I understood that. I felt it was kinda safer to require the bucket be empty.

PR #876, opened 2020-10-21
change log message when redirect doesn't redirect

mikehelmick:
## Proposed Changes

* change log message so as not to alarm operators

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

PR #881, opened 2020-10-22
Add test harness for headless browser testing

sethvargo:
After evaluation, I decided to use chromedp since it's a pure-go solution that should work headlessly on our existing CI infrastructure (fingers crossed).

This also adds a few new internal-only packages to help with testing:

- `internal/project` - project-wide helpers. Currently the only function is one that returns the filesystem root of the project, which is needed for telling the server where to find its assets. It will also be used by enx-redirector and any other filesystem-ey things.

- `internal/envstest` - globally shared test helpers. I'm going to start centralizing things here across all our test suites. This package also includes all the logic and helpers for spinning up a full in-memory UI server (with database and dependencies) for local testing.

- `internal/browser` - wrapper around chromedp. It's small, but I expect to add more helpers and things as we expand our testing.

**Release Note**

```release-note
Add test harness for headless browser testing
```

### DIFF
--- a/internal/browser/executor.go
+++ b/internal/browser/executor.go
@@ -1,0 +1,188 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package browser
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/chromedp/cdproto/cdp"
+	"github.com/chromedp/cdproto/emulation"
+	"github.com/chromedp/cdproto/network"
+	"github.com/chromedp/cdproto/page"
+	"github.com/chromedp/chromedp"
+)
+
+// defaultOptions are the default Chrome options.
+var defaultOptions = [...]chromedp.ExecAllocatorOption{
+	chromedp.NoFirstRun,
+	chromedp.NoDefaultBrowserCheck,
+
+	// After Puppeteer's default behavior.
+	chromedp.Flag("disable-background-networking", true),
+	chromedp.Flag("enable-features", "NetworkService,NetworkServiceInProcess"),
+	chromedp.Flag("disable-background-timer-throttling", true),
+	chromedp.Flag("disable-backgrounding-occluded-windows", true),
+	chromedp.Flag("disable-breakpad", true),
+	chromedp.Flag("disable-client-side-phishing-detection", true),
+	chromedp.Flag("disable-default-apps", true),
+	chromedp.Flag("disable-dev-shm-usage", true),
+	chromedp.Flag("disable-extensions", true),
+	chromedp.Flag("disable-features", "site-per-process,TranslateUI,BlinkGenPropertyTrees"),
+	chromedp.Flag("disable-hang-monitor", true),
+	chromedp.Flag("disable-ipc-flooding-protection", true),
+	chromedp.Flag("disable-popup-blocking", true),
+	chromedp.Flag("disable-prompt-on-repost", true),
+	chromedp.Flag("disable-renderer-backgrounding", true),
+	chromedp.Flag("disable-sync", true),
+	chromedp.Flag("force-color-profile", "srgb"),
+	chromedp.Flag("metrics-recording-only", true),
+	chromedp.Flag("safebrowsing-disable-auto-update", true),
+	chromedp.Flag("enable-automation", true),
+	chromedp.Flag("password-store", "basic"),
+	chromedp.Flag("use-mock-keychain", true),
+}
+
+// New creates a new headless browser context. Se NewFromOptions for usage.
+func New(tb testing.TB) context.Context {
+	tb.Helper()
+	opts := defaultOptions[:]
+	opts = append(opts, chromedp.Headless)
+	return NewFromOptions(tb, opts)
+}
+
+// NewHeadful creates a new browser context so you can actually watch the test.
+// This is for local debugging and will fail on CI where a browser isn't
+// actually available.
+func NewHeadful(tb testing.TB) context.Context {
+	tb.Helper()
+	return NewFromOptions(tb, defaultOptions[:])
+}
+
+// NewFromOptions creates a new browser instance. All future calls to `Run` must
+// use the context returned by this function!
+//
+// If this function returns successfully, a browser is running and ready to be
+// used. It's recommended that you wrap the returned context in a timeout.
+func NewFromOptions(tb testing.TB, opts []chromedp.ExecAllocatorOption) context.Context {
+	tb.Helper()
+
+	allocCtx, cancel := chromedp.NewExecAllocator(context.Background(), opts...)
+	tb.Cleanup(cancel)
+
+	taskCtx, cancel := chromedp.NewContext(allocCtx, chromedp.WithLogf(tb.Logf))
+	tb.Cleanup(cancel)
+
+	// Start browser
+	if err := chromedp.Run(taskCtx); err != nil {
+		tb.Fatal(err)
+	}
+
+	return taskCtx
+}
+
+// Screenshot captures a screenshot of the browser page in its current state.
+// This is useful for debugging a test failure. The dst will contain the
+// screenshot bytes in PNG format when the runner finishes.
+func Screenshot(dst *[]byte) chromedp.Action {
+	return chromedp.ActionFunc(func(ctx context.Context) error {
+		_, _, contentSize, err := page.GetLayoutMetrics().Do(ctx)
+		if err != nil {
+			return err
+		}
+
+		width, height := int64(math.Ceil(contentSize.Width)), int64(math.Ceil(contentSize.Height))
+
+		err = emulation.
+			SetDeviceMetricsOverride(width, height, 1, false).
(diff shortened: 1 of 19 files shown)


sethvargo:
Hmm... I guess chromedp needs Chrome 😄 

@chaodaiG any ideas of how we'd get chrome into our binary? There's a base image we could borrow from: https://github.com/chromedp/docker-headless-shell

mikehelmick:
For FB config - can we make that injected config from the runtime environment?

internal/project/root.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"path/filepath"
+	"runtime"
+)
+
+var _, self, _, _ = runtime.Caller(0)
+
+// Root returns the filepath to the root of this project.
+func Root() string {
```
jeremyfaller, line 26: This could be a global constant, like `self`, and the init generation code in the compiler will sort out the dependency chain.

internal/browser/executor.go
```diff
@@ -0,0 +1,123 @@
...
+	chromedp.Flag("use-mock-keychain", true),
+}
+
+// NewContext creates a new browser instance with the given options. Set
+// `headless` to false to watch the actual browser interaction. All future calls
+// to `Run` must use the context returned by this function!
+//
+// If this function returns successfully, a browser is running and ready to be
+// used. It's recommended that you wrap the returned function in a timeout.
+func NewContext(tb testing.TB, headless bool) context.Context {
```
jeremyfaller, line 63: This is a personal nit.

I hate functions that take random bools. It prevents readers of the calling code from understanding what is happening. I prefer to either hide this in `NewHeadlessContext` and `NewHeadedContext` or a type for the bool.

internal/envstest/random.go
```diff
@@ -0,0 +1,40 @@
...
+		tb.Fatal(err)
+	}
+	if n < length {
+		tb.Fatal(fmt.Errorf("insufficient bytes read: %v, expected %v", n, length))
+	}
+	return buf
+}
+
+// RandomString returns a random hex-encoded string of the given length.
+func RandomString(tb testing.TB, length int) string {
```
jeremyfaller, line 38: API is slightly clunky. `length` is the length of the random string, not the encoded string. Might want to document.

pkg/controller/home/home_test.go
```diff
@@ -0,0 +1,90 @@
...
+	taskCtx, done := context.WithTimeout(browserCtx, 30*time.Second)
+	defer done()
+
+	var code string
+	if err := chromedp.Run(taskCtx,
+		chromedp.Navigate(`http://`+harness.Server.Addr()),
+		Login("admin@example.com", "Password"),
+
+		// Post-login action is /home.
+		chromedp.WaitVisible(`body#home`, chromedp.ByQuery),
```
jeremyfaller, line 60: What if there's a a "verify with text message" screen? Not sure how `WaitVisible` handles it.

pkg/controller/home/home_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/google/exposure-notifications-verification-server/internal/browser"
+	"github.com/google/exposure-notifications-verification-server/internal/envstest"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+
+	"github.com/chromedp/chromedp"
+)
+
+func TestHome(t *testing.T) {
```
jeremyfaller, line 29: s/TestHome/TestLoginAndIssueCode/

pkg/controller/home/home_test.go
jeremyfaller: NVM, I see below.

pkg/database/database_util.go
```diff
@@ -34,13 +35,7 @@ var (
...
 )
 
-// NewTestDatabaseWithConfig creates a new database suitable for use in testing.
-// This should not be used outside of testing, but it is exposed in the main
-// package so it can be shared with other packages.
-//
-// All database tests can be skipped by running `go test -short` or by setting
-// the `SKIP_DATABASE_TESTS` environment variable.
-func NewTestDatabaseWithConfig(tb testing.TB) (*Database, *Config) {
+func NewTestDatabaseWithCacher(tb testing.TB, cacher cache.Cacher) (*Database, *Config) {
```
jeremyfaller, line 38: recreate some of the doc comment?

jeremyfaller (commented):
Modulo some nits, this LGTM.

chaodaiG:
> Hmm... I guess chromedp needs Chrome 😄
> 
> @chaodaiG any ideas of how we'd get chrome into our binary? There's a base image we could borrow from: https://github.com/chromedp/docker-headless-shell

How about add `docker-headless-shell` in prow test base image, and chromedp references it's entrypoint `/headless-shell/headless-shell` as ExecPath https://godoc.org/github.com/knq/chromedp#ExecPath?

internal/project/root.go
sethvargo: It'd be a global `var`, which would be mutable. AFAIK, you can't create a constant that uses `filepath.Join` since it wouldn't be constant across systems.

internal/envstest/random.go
sethvargo: It is the length of the encoded string (that's why I divide by 2).

sethvargo:
That could work, but it will make running tests locally harder. It would be better if we could have chrome or chrome-headless in `$PATH`.

pkg/controller/home/home_test.go
sethvargo: Yea, this is the first pass. I want to do some more thinking on how to "login" since it will be required for all journeys. I'm wondering if we can curate and inject a cookie instead of depending on firebase...

chaodaiG:
> @chaodaiG That could work, but it will make running tests locally harder. It would be better if we could have chrome or chrome-headless in `$PATH`.

Agreed, we can append `/headless-shell` to PATH  in dockerfile

internal/envstest/server.go
```diff
@@ -0,0 +1,208 @@
...
+	Cacher      cache.Cacher
+	KeyManager  keys.KeyManager
+	RateLimiter limiter.Store
+	Server      *server.Server
+}
+
+// NewServer creates a new test UI server instance. When this function returns,
+// a full UI server will be running locally on a random port. Cleanup is handled
+// automatically.
+func NewServer(tb testing.TB) *TestServerResponse {
```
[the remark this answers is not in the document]
sethvargo: It returns the server and everything needed to run it. We have to return things like the database because tests might need to insert or assert records.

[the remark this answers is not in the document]
sethvargo: I'm open to other names

internal/browser/executor.go
```diff
@@ -0,0 +1,123 @@
...
+	chromedp.Flag("password-store", "basic"),
+	chromedp.Flag("use-mock-keychain", true),
+}
+
+// NewContext creates a new browser instance with the given options. Set
+// `headless` to false to watch the actual browser interaction. All future calls
+// to `Run` must use the context returned by this function!
+//
+// If this function returns successfully, a browser is running and ready to be
+// used. It's recommended that you wrap the returned function in a timeout.
```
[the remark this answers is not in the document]
sethvargo: Done

internal/envstest/server.go
```diff
@@ -0,0 +1,208 @@
...
+	Database    *database.Database
+	Cacher      cache.Cacher
+	KeyManager  keys.KeyManager
+	RateLimiter limiter.Store
+}
+
+// NewServerConfig creates a new server configuration. It creates all the keys,
+// databases, and cacher, but does not actually start the server. All cleanup is
+// scheduled by t.Cleanup.
+func NewServerConfig(tb testing.TB) *ServerConfigResponse {
```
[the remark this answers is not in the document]
sethvargo: Done

internal/browser/executor.go
sethvargo: Done

pkg/controller/home/home_test.go
sethvargo: Done (picked a slightly different name)

pkg/database/database_util.go
sethvargo: Done

sethvargo:
@chaodaiG sounds good. Do you have any cycles to work on that today? I imagine we're going to have some trial and error to get it working against this PR.

internal/project/root.go
jeremyfaller: fair point. `Root` could still be a LHS.

chaodaiG:
> @chaodaiG sounds good. Do you have any cycles to work on that today? I imagine we're going to have some trial and error to get it working against this PR.

https://github.com/google/exposure-notifications-server/pull/1104
@sethvargo could you take a quick look? If there is no obvious problem I'll push the image for testing current PR

sethvargo:
Just approved

chaodaiG:
Image pushed

sethvargo:
Hmm @chaodaiG 

```
could not dial "ws://127.0.0.1:42897/devtools/browser/91b93c42-b53a-44d1-8b9c-7c822949108a": EOF
```

Different error this time. Do we need to do something to allow for that networking?

internal/project/root.go
sethvargo: What's an LHS?

sethvargo (commented):
@chaodaiG @jeremyfaller @whaught this is ready for review now

internal/project/root.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Package project defines global project helpers.
+package project
+
+import (
+	"path/filepath"
+	"runtime"
+)
+
+var _, self, _, _ = runtime.Caller(0)
```
chaodaiG, line 23: Would the following be more reliable?
```
git rev-parse --show-toplevel
```

chaodaiG (approved):
LGTM generally, left a comment

internal/project/root.go
sethvargo: That will only work if someone cloned the repo via git. This works regardless and doesn't require `git` to be present on the system.

internal/project/root.go
jeremyfaller: left hand side.

PR #883, opened 2020-10-22
thread test date all the way through

### LINKED ISSUE #882: Store and return testDate on verify code API
### TL;DR

Mobile apps would like to differentiate between testDate and symptomDate for purposes of prompting the user.

Currently this is collapsed to a single date that is stored and transported through the verify response, verification token, and verification certificate.

### Design

**Proposal**

1) add `TestDate` to the verification code struct
2) store this in the DB
3) when the verificaiton token is minted, expand the subject to contain `diagnosis.symptomdate.testdate`
4) verification token structure will remain the same - a single date called symptom date, that can be either.

/kind enhancement

mikehelmick:
Fixes #882

## Proposed Changes

* Add test date into verify API return
* Thread this all the way through the verification token

**Release Note**

```release-note
API CHANGE : /api/verify now returns the testDate in addition to symptom date if present. When the verification certificate is issue only one interval is inserted: symptomDate if present, testDate if not.
```

### DIFF
--- a/pkg/database/token.go
+++ b/pkg/database/token.go
@@ -98,7 +98,7 @@ func ParseSubject(sub string) (*Subject, error) {
 	}
 
 	var testDate *time.Time
-	if len(parts) == 3 {
+	if len(parts) == 3 && parts[2] != "" {
 		parsedDate, err := time.Parse("2006-01-02", parts[2])
 		if err != nil {
 			return nil, fmt.Errorf("subject contains invalid test date: %w", err)
--- a/pkg/database/token.go
+++ b/pkg/database/token.go
@@ -50,6 +50,7 @@ type Token struct {
 	TokenID     string `gorm:"type:varchar(200); unique_index"`
 	TestType    string `gorm:"type:varchar(20)"`
 	SymptomDate *time.Time
+	TestDate    *time.Time
 	Used        bool `gorm:"default:false"`
 	ExpiresAt   time.Time
 }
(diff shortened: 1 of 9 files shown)


mikehelmick:
/hold
for verification w/ clinical team

pkg/database/token.go
```diff
@@ -77,8 +85,8 @@ func (s *Subject) SymptomInterval() uint32 {
 
 func ParseSubject(sub string) (*Subject, error) {
 	parts := strings.Split(sub, ".")
-	if length := len(parts); length < 1 || length > 2 {
-		return nil, fmt.Errorf("subject must contain 2 parts, got: %v", length)
+	if length := len(parts); length < 2 || length > 3 {
```
sethvargo, line 88: Won't think break existing JWT tokens that have the old subject?

mikehelmick: the old if statement was wrong - it should have always been 2 parts before.

Why I didn't write `length != 2` the first time, we'll never know.

jeremyfaller: Not that I doubt your code, but this feels like a test.

sethvargo: ^ what Jeremy said

mikehelmick: funny enough - the 2 segment subject was the only thing under test before.
test coverage has been boosted to cover all scenarios.

jeremyfaller:
/lgtm

still hold, pending clinicians.

PR #884, opened 2020-10-22
Rename all e2e  test scripts to make people not want to run them locally

chaodaiG:
To prevent future misuse of these scripts, e.g. running them in prod


sethvargo (approved):
/lgtm

Issue #873, opened 2020-10-21
Looping redirect on fresh bootstrapped project


bschlaman:
### TL;DR
After provisioning ENVS for the first time using the "production" setup, logging in with the default user "super@example.com" results in looping redirects through several endpoints.

@whaught 

**Expected behavior**
Successful login

**Observed behavior**
Seemed to get 303 redirected through these endpoints until hitting rate limits (somewhere around 100 requests):
/admin
/login/manage-account?mode=verifyEmail
/login/select-realm

Before finally getting redirected to /logout

I can see these in the logs:
"realm does not exist in session"
"user email not verified"

In chrome, ERR_TOO_MANY_REDIRECTS also shows.  Clearing cookies did not resolve the problem, nor did testing from mobile or a new ip.

### Reproduction
Start with a fresh gcp project and follow the steps outlined in docs/production.md.
As stated, enter super@example.com into Firebase and enable Identity Platform.

Behavior can be seen at my server URL:
https://server-dyjepj3nfa-uc.a.run.app/
User: super@example.com
Pw: asdfasdf

**Environment**
- OS: windows & linux
- Browser: chrome: ERR_TOO_MANY_REDIRECTS
- Browser: ie: HTTP 405
- Version: chrome: 86.0.4240.75
- Version: ENVS: v13

bschlaman:
Should also mention I'm not using a load balancer or custom domain

sethvargo:
Can you share the logs (either copy-paste or screenshot) from around that time? If you can enable debug logs (set `LOG_DEBUG=true` as an envvar on the service) and run again, that'd be super helpful.

whaught:
Sorry not sure if I closed this prematurely. Could you try redeploying with #874 and verify that the scenario succeeds?

bschlaman:
Yes - done.  The problem is still there, but I think I found the issue.

bschlaman:
The problem is occurring because an admin is logging in without having verified the email, and with 0 realms present yet.  Here is the redirect flow:

POST /session
GET /login/select-realm (this first GET is from the login js script)
Redirect to “/admin”
 - First: requireVerified
Redirect to “/login/manage-account?mode=verifyEmail”
 - First: processFirewall -> controller.MissingRealm(w, r, h)
GET /login/select-realm (subsequent GETs to this endpoint are from MissingRealm)
(process repeats)

whaught:
[...]

PR #887, opened 2020-10-22
Fix verify email flow with no realm

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/873

## Proposed Changes

* Drops the realm required for verifying email
* Protect the template in the case realm is nil
* Make sure the email link works properly (html unescape)

**Release Note**

```release-note
Fix redirect loop when verifying email for an admin with no realm selected
```

### DIFF
--- a/cmd/server/assets/email/default_email_verify.html
+++ b/cmd/server/assets/email/default_email_verify.html
@@ -5,7 +5,7 @@ Hello,
 
 Click the link below to verify your email address for {{.RealmName}} on the COVID-19 exposure notifications verification server:
 
-{{.VerifyLink | unescape}}
+{{unescape .VerifyLink}}
 
 If you did not request verification of this email, please disregard this message.
 {{end}}
--- a/cmd/server/assets/email/default_email_verify.html
+++ b/cmd/server/assets/email/default_email_verify.html
@@ -5,7 +5,7 @@ Hello,
 
 Click the link below to verify your email address for {{.RealmName}} on the COVID-19 exposure notifications verification server:
 
-{{.VerifyLink}}
+{{.VerifyLink | unescape}}
 
 If you did not request verification of this email, please disregard this message.
 {{end}}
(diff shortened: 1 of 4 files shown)


cmd/server/assets/email/default_email_verify.html
```diff
@@ -5,7 +5,7 @@
 
 Click the link below to verify your email address for {{.RealmName}} on the COVID-19 exposure notifications verification server:
 
-{{.VerifyLink}}
+{{.VerifyLink | unescape}}
```
jeremyfaller, line 8: TIL about `|`.

internal/routes/server.go
```diff
@@ -169,13 +169,6 @@ func Server(ctx context.Context, cfg *config.ServerConfig, db *database.Database
 			sub.Handle("/login/change-password", loginController.HandleShowChangePassword()).Methods("GET")
 			sub.Handle("/login/change-password", loginController.HandleSubmitChangePassword()).Methods("POST")
 			sub.Handle("/account", loginController.HandleAccountSettings()).Methods("GET")
-
-			// Verifying email requires the user is logged in
-			sub = r.PathPrefix("").Subrouter()
-			sub.Use(requireAuth)
-			sub.Use(rateLimit)
```
jeremyfaller, line 176: still feels like we should keep the rate limiter?

jeremyfaller (approved):
/hold

hold until you decide on rate limiter.

lgtm

internal/routes/server.go
whaught: see line 164. This merges it with the above subrouter which still has rate limiting

pkg/render/render.go
```diff
@@ -128,13 +128,18 @@ func loadTemplates(tmpl *template.Template, root string) error {
 	})
 }
 
+func unescape(s string) template.HTML {
```
sethvargo, line 131: Nit: I'd prefer to name this `safeHTML` or something instead? It's possible we might want to do the same for JS or CSS in the future.

whaught: done. added a doc comment

jeremyfaller (approved):

sethvargo (approved):
/lgtm

PR #889, opened 2020-10-22
add help text around iOS ENX

mikehelmick:
## Proposed Changes

* Add help text around mobile app for iOS. Adding one when using ENX can break things.

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

Issue #871, opened 2020-10-21
Improve charts on realm stats


jeremyfaller:
Mike sent me some feedback on the realm stats yesterday. Specifically:

- It's not clear that clicking a modifys/selects a user in the per-realm stats
- It would be nice to see per-user-per-day stats numbers, possibly in the chart.
- possibly replace the overall line graph on a row click.
- Also, mike suggested reversing the order of the days.

I'm going to tweak the charts, and improve them.

PR #894, opened 2020-10-26
Allow app to disallow particular hostnames

whaught:
## Proposed Changes

* Allow server to block particular hostnames
    * Can be used to stop run.app from serving in favor of a custom domain

**Release Note**

```release-note
Allow blocking configured hostnames from serving eg. 'run.app'
```

### DIFF
--- a/pkg/controller/middleware/blockedhosts.go
+++ b/pkg/controller/middleware/blockedhosts.go
@@ -17,7 +17,6 @@ package middleware
 
 import (
 	"context"
-	"net"
 	"net/http"
 	"strings"
 
@@ -39,13 +38,8 @@ func CheckBlockedHosts(ctx context.Context, blockedHosts []string, h *render.Ren
 				next.ServeHTTP(w, r)
 			}
 
-			rHost, _, err := net.SplitHostPort(r.Host)
-			if err != nil { // No port or invalid address
-				rHost = r.Host
-			}
-
 			for _, host := range blockedHosts {
-				if strings.HasSuffix(rHost, host) {
+				if strings.HasSuffix(r.URL.Hostname(), host) {
 					logger.Warnf("received request from blocked domain %s", r.Host)
 					controller.Unauthorized(w, r, h)
 					return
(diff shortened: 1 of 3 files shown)


pkg/controller/middleware/blockedhosts.go
```diff
@@ -0,0 +1,51 @@
...
+	logger := logging.FromContext(ctx).Named("middleware.CheckBlockedHosts")
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if blockedHosts == nil {
+				next.ServeHTTP(w, r)
+			}
+
+			for _, host := range blockedHosts {
+				if strings.HasSuffix(r.Host, host) {
```
jeremyfaller, line 42: IIRC, `r.Host` can contain the port number. So, `foo.com:80` won't trigger if foo.com is listed. I'd strip the port here.

jeremyfaller (changes requested):

pkg/controller/middleware/blockedhosts.go
jeremyfaller: This has another problem in that if the host is myfoo.com, and foo.com is blocked, I will get unintentionally blocked.

icco:
See https://github.com/google/exposure-notifications-verification-server/issues/871

whaught:
/hold

All good points about breaking down hostname. This was a quick solution, but looks like folks are leaning towards doing this at the load-balancer level rather than application-level

yuriatgoogle (commented):
The current plan is to use the --ingress flag:
https://cloud.google.com/sdk/gcloud/reference/alpha/run/deploy#--ingress
once it's in GA.  Working to determine timeline for that.

sethvargo:
@whaught are we good to close this out since the `--ingress` flag seems to be coming soon?

whaught:
> @whaught are we good to close this out since the `--ingress` flag seems to be coming soon?

Yea this would be an interim solution if we can't figure out --ingress or load balancer. I assume that is coming shortly?

/close

PR #897, opened 2020-10-26
Pull CSRF token from metadata

sethvargo:
When I moved the javascript to be external, `{{.csrfToken}}` no longer renders.

**Release Note**

```release-note
Pull CSRF token from meta tags
```


sethvargo:
Got a bunch of connection reset errors 😦

mikehelmick (approved):

PR #899, opened 2020-10-27
Use text templates for emails (not-html)

whaught:
## Proposed Changes

* Switch emails to plaintext renderer

**Release Note**

```release-note
Switch emails to plaintext renderer
```

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -19,11 +19,12 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"html/template"
+	htmltemplate "html/template"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
+	texttemplate "text/template"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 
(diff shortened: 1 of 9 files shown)


sethvargo (changes requested):
I think we should be wrapping the content in `template.HTML` instead, but, even then, `template.HTML` is stripping the `=` from URLs.

pkg/render/render.go
```diff
@@ -116,24 +120,28 @@ func loadTemplates(tmpl *template.Template, root string) error {
...
-			return nil
+		if strings.HasSuffix(info.Name(), ".html") {
+			if _, err := tmpl.ParseFiles(path); err != nil {
+				return fmt.Errorf("failed to parse %s: %w", path, err)
+			}
 		}
 
-		if _, err := tmpl.ParseFiles(path); err != nil {
-			return fmt.Errorf("failed to parse %s: %w", path, err)
+		if strings.HasSuffix(info.Name(), ".txt") {
```
sethvargo, line 129: Clever

sethvargo (approved):
/lgtm

PR #904, opened 2020-10-27
fix email from string

whaught:
## Proposed Changes

* .FromEmail should be sent to the template as a string rather than a func


sethvargo (approved):
/lgtm

PR #907, opened 2020-10-27
Cleanup metrics in ratelimit middleware

### LINKED ISSUE #773: Deduplicate metrics

### LINKED ISSUE #906: Build tag missing from some metrics
### TL;DR
All metrics should have a label with the build version, it seems to have disappeared in some cases. Let's find out why.

**Expected behavior**

A label for build_id and build_version on all metrics.

**Observed behavior**

Both labels were missing in the /rate_limited.* metrics

yegle:
Fixes #773

Also appear to fix #906


pkg/ratelimit/limitware/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+const metricPrefix = observability.MetricRoot + "/ratelimit/limitware"
+
+var (
+	mRequest = stats.Int64(metricPrefix+"/request", "requests seen by middleware", stats.UnitDimensionless)
+)
+
+func init() {
+	enobservability.CollectViews(append(redis.ObservabilityMetricViews,
```
icco, line 32: Should we panic here if this fails?

pkg/ratelimit/limitware/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+package limitware
+
+import (
+	enobservability "github.com/google/exposure-notifications-server/pkg/observability"
+	"github.com/google/exposure-notifications-verification-server/pkg/observability"
+	"github.com/opencensus-integrations/redigo/redis"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
```
icco, line 22: We usually seperate this with newlines into three blocks env imports, en imports and third party imports

pkg/ratelimit/limitware/middleware.go
```diff
@@ -84,58 +73,10 @@ func NewMiddleware(ctx context.Context, s limiter.Store, f httplimit.KeyFunc, op
...
-	if err := view.Register(&view.View{
-		Name:        MetricPrefix + "/rate_limited_count",
-		Measure:     rl,
-		Aggregation: view.Count(),
-		TagKeys:     []tag.Key{},
-	}); err != nil {
-		return nil, fmt.Errorf("stat view registration failure: %w", err)
-	}
-
-	if err := view.Register(redis.ObservabilityMetricViews...); err != nil {
```
icco, line 127: Are these registered elswhere?

PR #909, opened 2020-10-27
Clean up deleted API keys after 1 week

### LINKED ISSUE #905: Add API Keys to cleanup job
### TL;DR

Our e2e test runner has created hundreds of API keys

API Keys should be deleted ~30 days after they have been disabled.

We may want to also consider a metric for only keeping the most recent X API keys of each type (50?)

These are outsized things that should only occur in tests

/kind enhancement

sethvargo:
Fixes #905

**Release Note**

```release-note
Clean up deleted API keys after 1 week
```


mikehelmick:
See comment in the bug around cleanup time - open to discussion

/hold

mikehelmick (approved):
/cancel hold

PR #913, opened 2020-10-27
Try and fix use case of burst capacity

icco:
## Proposed Changes

* Fixes metric based on conversation with @sethvargo 
* Found while writing docs with @mariliamelo 

**Release Note**

```release-note
Fixes overflow error in capacity metrics when burst quota is given.
```


sethvargo (approved):
/lgtm

PR #914, opened 2020-10-27
Validate and populate realm email templates

whaught:
## Proposed Changes

* Add validation to realm-email-templates before save
* Populate emails from these instead of default

Note: There is no UI to  set  these yet. That's coming next.


sethvargo (approved):
/lgtm

PR #915, opened 2020-10-28
Include email templates / useSystem in auditing

whaught:
## Proposed Changes

* Record changes to UseSystemEmail and templates to auditing


sethvargo (approved):
/lgtm

PR #918, opened 2020-10-28
Create alerts using MQL

yegle:
This alert is currently set up using a local-exec provisioner, due to
the lack of support to define alert policies using MQL in Terraform.

Part of #810

This is a revive of #831.


yegle:
cc @sethvargo PTAL. I think this still worth submitting as an intermediate workaround until the official MQL alert support is merged upstream.

sethvargo (approved):
/lgtm

PR #919, opened 2020-10-28
Allow metric exporter to finish uploading the last batch of metrics

### LINKED ISSUE #912: Container seem to exit ~immediately when SIGTERM is received
I've noticed we have several "failed to export metric" log lines, usually associate with server shutdown.

image

Per the Cloud Run container contract (https://cloud.google.com/run/docs/reference/container-contract#instance-shutdown) when killing a container a `SIGTERM` is sent first, and `SIGKILL` 10s later.

A 10s window should be sufficient to flush the remaining timeseries that has not be uploaded yet. Per the following query our 99pct latency of `CreateTimeSeries` call rarely exceeds 10s:

```
fetch generic_task
| metric 'custom.googleapis.com/opencensus/grpc.io/client/roundtrip_latency'
| filter
    (metric.grpc_client_method
     == 'google.monitoring.v3.MetricService/CreateTimeSeries')
| align delta(1m)
| every 1m
| group_by [],
    [value_roundtrip_latency_percentile:
       percentile(value.roundtrip_latency, 99)]
```

image

Did we make some mistake in our signal handling code that result in this issue?

yegle:
before the server is killed with SIGKILL.

See
https://github.com/google/exposure-notifications-verification-server/issues/912
for context.

Fixes #912


yegle:
@sethvargo Would this work?

sethvargo:
I think it works, but could maybe be simplified. I'm pretty sure you can just defer closing oe, right?

yegle:
Simplified and updated all binaries. PTAL.

icco (approved):
I'm surprised defer isn't doing the right thing here. That being said, lgtm.

For future reading @jeremyfaller I'd be interested in your thoughts on why defer doesn't do the right thing here.

/lgtm

cmd/adminapi/main.go
```diff
@@ -74,14 +74,23 @@ func realMain(ctx context.Context) error {
...
 	if err != nil {
+		cancel()
 		return fmt.Errorf("unable to create ObservabilityExporter provider: %w", err)
 	}
 	if err := oe.StartExporter(); err != nil {
+		cancel()
 		return fmt.Errorf("error initializing observability exporter: %w", err)
 	}
-	defer oe.Close()
+	defer func() {
```
sethvargo, line 89: Sorry, but I don't like this API or the repeatability. It's really too easy for someone to miss a `cancel` or misuse this. Can we push some of this logic into the observability package? Or can we just create a new context (not cancelable) and leave flow control timeouts up to Cloud Run?

```go
metricsCtx := context.Background()
oe, err := observability.NewFromEnv(metricCtx, oeConfig)
if err != nil {
  return ...
}
defer oe.Close()
```

That will still give the oe the cleanup signal and it will prevent the full function from stopping until the defer finishes (assuming `Close` is blocking).

Even better, we could drop the context parameter from the o11y entirely and just construct the context internally. In both cases, we're losing the logger that's attached to the provided context, so maybe we want `NewFromEnv` to look like this:

```go
NewFromEnv(ctx context.Context, config *Config) {
  logger := logger.FromContext(ctx)

  ctx = context.Background()
  ctx = logging.WithLogger(ctx, logger)

  // Existing stuffs
}
```

Then all the `main.go` functions are entirely unchanged.

sethvargo (changes requested):

cmd/adminapi/main.go
yegle: Updated with the simplified logic. The exporter will use its own ctx anyway so `observability.NewFromEnv` don't need to take a ctx.

sethvargo (approved):
/lgtm

sethvargo (approved):
/lgtm

PR #920, opened 2020-10-28
Return json if requested from realm-stats page.

### LINKED ISSUE #916: implement stats export
### TL;DR

Implement the ability to export CSV versions of export of the codes issued / codes claimed.

/kind enhancement
/assign @jeremyfaller

jeremyfaller:
Fixes #916

**Release Note**

```release-note
realm/stats now has support for returning json results if requested through the HTTP headers.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -45,14 +45,20 @@ func Back(w http.ResponseWriter, r *http.Request, h *render.Renderer) {
 	http.Redirect(w, r, r.Header.Get("Referer"), http.StatusSeeOther)
 }
 
+// getAcceptedTypes returns all content types accepted by a given http.Request.
+func getAcceptedTypes(r *http.Request) []string {
+	accept := strings.Split(r.Header.Get("Accept"), ",")
+	accept = append(accept, strings.Split(r.Header.Get("Content-Type"), ",")...)
+	return accept
+}
+
 // InternalError handles an internal error, returning the right response to the
 // client.
 func InternalError(w http.ResponseWriter, r *http.Request, h *render.Renderer, err error) {
 	logger := logging.FromContext(r.Context())
 	logger.Errorw("internal error", "error", err)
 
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -45,14 +45,20 @@ func Back(w http.ResponseWriter, r *http.Request, h *render.Renderer) {
 	http.Redirect(w, r, r.Header.Get("Referer"), http.StatusSeeOther)
 }
 
+// getAcceptedTypes returns all content types accepted by a given http.Request.
```
sethvargo, line 48: Nit: this should be `acceptedTypes` per Go style

pkg/controller/realmadmin/show.go
```diff
@@ -66,7 +66,15 @@ func (c *Controller) HandleShow() http.Handler {
 			return
 		}
 
-		c.renderShow(ctx, w, realm, stats, userStats)
+		if controller.AcceptsType(r, controller.ContentTypeJSON) {
```
sethvargo, line 69: If you plan on adding CSV, it'd probably be better to use a switch here:

```
switch {
case controller.AcceptsType(r, controller.ContentTypeJSON):
case controller.AcceptsType(r, controller.ContentTypeCSV):
default:
  // HTML
```

pkg/controller/realmadmin/show.go
```diff
@@ -66,7 +66,15 @@ func (c *Controller) HandleShow() http.Handler {
...
+			resp := struct {
```
sethvargo, line 70: This should either be a top-level struct documented in the `api` package or at least defined outside of the body so it's not re-allocated each time.

sethvargo (changes requested):

jeremyfaller:
Closing. I suck at reading requirements.

sethvargo:
Wait, I actually like this and think we should do it.

PR #925, opened 2020-10-28
change default allowed back dating to 27 days (from 14)

mikehelmick:
## Proposed Changes

* Default symptom age changed to 27 days from 14

This is to cover this use case

Someone has symptom onset 27 days ago, and didn't get their verification code until "today." In that case the oldest TEK on that person's phone would be +14 days from symptom onset and could still cause exposure notifications on someone else's phone.

**Release Note**

```release-note
Change default symptom age from 14 to 27 days.
```


whaught (approved):

greggles:
Thank you!

PR #931, opened 2020-10-29
add ZERO WIDTH NO-BREAK SPACE to our cut set

mikehelmick:
## Proposed Changes

* add unicode zero width no-break space to our cutset for certififcate issuer and audience

Workaround of https://github.com/golang/go/issues/42274

**Release Note**

```release-note
Fixes an issue where certain non-printable unicode characters would accepted as valid characters for the verification certificate issuer and audience.
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -49,6 +49,8 @@ var (
 	// callbackLock prevents multiple callbacks from being registered
 	// simultaneously because that's a data race in gorm.
 	callbackLock sync.Mutex
+
+	ExtraCutset = fmt.Sprintf("%v", '\uFEFF')
 )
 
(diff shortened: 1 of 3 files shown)


pkg/database/database.go
```diff
@@ -73,6 +75,10 @@ type Database struct {
 	statsCloser func()
 }
 
+func trim(s string) string {
```
sethvargo, line 78: We can do it later, but I feel like this should be a more global thing that we use everywhere.

sethvargo (approved):
/lgtm

pkg/database/database.go
mikehelmick: yep - fixing the immediate known badness

PR #935, opened 2020-10-29
Download stats in a card at the end?

whaught:
## Proposed Changes

* An alternative way to show download links for stats

exportbottom


jeremyfaller:
/hold
/lgtm

I am ok with whatever here. The only thought I had is that it's at the bottom of the page, and kinda sad down there. Might be better between the chart and table? I'll let Weston make the final decision unless someone else weighs in.

jeremyfaller:
/lgtm

I like it between

PR #943, opened 2020-10-30
Generalize the RecordLatency function to be reused later.

yegle:


pkg/observability/observability.go
```diff
@@ -151,3 +153,21 @@ func WithBuildInfo(octx context.Context) context.Context {
...
 }
+
+// RecordLatency calculate and record the latency.
+// Usage example:
+// func foo() {
+// 	 defer RecordLatency(ctx, metric, tag1, tag2)()
+//   // remaining of the function body.
+// }
+func RecordLatency(ctx context.Context, m *stats.Float64Measure, mutators ...*tag.Mutator) func() {
+	start := time.Now()
```
sethvargo, line 164: Can we make start a parameter to the function to avoid the closure?

```go
defer RecordLatency(ctx, time.Now(), metric, tag1, tag2)
```

yegle: Done

pkg/observability/observability.go
```diff
@@ -151,3 +153,20 @@ func WithBuildInfo(octx context.Context) context.Context {
...
 	return ctx
 }
+
+// RecordLatency calculate and record the latency.
+// Usage example:
+// func foo() {
+// 	 defer RecordLatency(ctx, time.Now(), metric, tag1, tag2)
+//   // remaining of the function body.
+// }
+func RecordLatency(ctx context.Context, start time.Time, m *stats.Float64Measure, mutators ...*tag.Mutator) {
```
sethvargo, line 163: Why use `*tag.Mutator` here only to deref later? If you make this `mutators ...tag.Mutator`, then we avoid the iteration and can pass it directly to `RecordWithTags`.

yegle: This is intended to use in a function as a defer function, and should appear at the top of the function. The recommended way to use this function is to define the needed mutator before deferring this function:

```
func foo() {
  var tag1, tag2 tag.Mutator
  defer RecordLatency(ctx, time.Now(), m, &tag1, &tag2)
  if condition {
    tag1 = xxx
    tag2 = yyy
 } else {
    tag1 = zzz
    tag2 = aaa
  }
  return
```

Capturing the tags by pointer will make it more useful.

sethvargo: Ah I see. Got it.

PR #944, opened 2020-10-30
Record latency in our API endpoint

yegle:
A caveat about the recorded latency: if there are multiple steps in an API call and it failed at step x, the latency recorded would be the total latency up to step x, including the latency of all previous steps.

Ideally we should record the latency of each steps, but the challenge part is to define a "step" that would the drill down useful. We can make a separate "phased latency" if we need it later.


yegle:
PR ready for review now.

Issue #933, opened 2020-10-29
Use super trim


mikehelmick:
Move the trim function from
  https://github.com/google/exposure-notifications-verification-server/pull/931

Into a common location 

use this instead of the standard strings.TrimSpace

/kind enhancement

sethvargo:
We may want to consider inverting the logic to an _allowed_ list of characters. There are many non-printable unicode characters and other printable unicode characters that we wouldn't want to allow. 👍 is valid unicode, but it's not valid in a certificate.

PR #948, opened 2020-11-02
Use super-trim all over

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/933

## Proposed Changes

* trim unicode zero-width no-break space everywhere
* add an option to trim space and non-printable
* Followup to @mikehelmick  https://github.com/google/exposure-notifications-verification-server/pull/931

### DIFF
--- a/internal/project/trim_test.go
+++ b/internal/project/trim_test.go
@@ -18,20 +18,20 @@ import (
 	"testing"
 )
 
-func TestTrimSpaceAndNonASCII_unicode(t *testing.T) {
+func TestTrimSpaceAndNonPrintable_unicode(t *testing.T) {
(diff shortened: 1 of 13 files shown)


internal/project/trim_test.go
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package project
+
+import (
+	"testing"
+)
+
+func TestTrimSpaceAndNonPrintable_unicode(t *testing.T) {
```
jeremyfaller, line 21: slightly more idiomatic is to combine these into a single function with testcases.

This is fine, just FYI.

jeremyfaller (approved):

PR #949, opened 2020-11-02
add debug logging to device APIs

### LINKED ISSUE #946: Better logging
Hello,
we are one of the users of the referential Verification (and Key) server. We're quite often facing problems that our users are unable to send us their keys and, reciprocally, we're unable to help him because we don't see almost anything useful in logs.

Imagine a case: user enters a verification code that was issued for him a few minutes ago and the upload fails, seeing "invalid code" message on the device. The problem is that even with our traffic (1.3M devices), we're often unable to find relevant lines in the log and even if we find them, we see nothing more than e.g. "verification code expired" (in a case when this should not be possible as the code was young enough).

I'd like to suggest you revise the logging in the whole Verification server (and especially the `verifyapi`) to show more useful information. I think it should be optional so no one would be forced to see that if he doesn't want to - logging level could solve that. I've noticed DEBUG logging level is not used in the Verification server almost at all...

An additional thing I'd love to see is to introduce a request tracing. A minimal form is that a reader of logs should be able to connect logging entries together to a single request to see the flow. An ideal case would be to have support for e.g. [opentracing](https://github.com/opentracing/opentracing-go) so the device would generate a pseudo-unique request ID for every request and this ID would be visible together with an error message to the user so he could report the ID to the support. Things would get much easier then... (and it still shouldn't jeopardize user privacy)

Please let me know what do you think about this. I'm willing to participate in the implementation if you'd like to. This is a real support-pain for us.

mikehelmick:
Fixes #946 

## Proposed Changes

* Add debug logging to device APIs

**Release Note**

```release-note
Added debug logging to device APIs
```

### DIFF
--- a/pkg/controller/certapi/certificate.go
+++ b/pkg/controller/certapi/certificate.go
@@ -86,6 +86,7 @@ func (c *Controller) HandleCertificate() http.Handler {
 		// Parse and validate the verification token.
 		tokenID, subject, err := c.validateToken(ctx, request.VerificationToken, allowedPublicKeys)
 		if err != nil {
+			c.logger.Debugw("verification token invalid", "error", err)
 			c.h.RenderJSON(w, http.StatusBadRequest, api.Error(err).WithCode(api.ErrTokenInvalid))
 			blame = observability.BlameClient
 			result = observability.ResultError("FAILED_TO_VALIDATE_TOKEN")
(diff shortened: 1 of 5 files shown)


pkg/controller/certapi/certificate.go
```diff
@@ -95,15 +96,17 @@ func (c *Controller) HandleCertificate() http.Handler {
...
 		if err != nil {
+			c.logger.Debugw("provided invalid hmac, not base64", "error", err)
 			c.h.RenderJSON(w, http.StatusBadRequest,
 				api.Errorf("exposure key HMAC is not a valid base64: %v", err).WithCode(api.ErrHMACInvalid))
 			blame = observability.BlameClient
 			result = observability.ResultError("FAILED_TO_DECODE_HMAC")
 			return
 		}
-		if len(hmacBytes) != 32 {
+		if l := len(hmacBytes); l != 32 {
```
sethvargo, line 106: nit: we should save 32 as a var here and use it later too

pkg/controller/verifyapi/verify.go
```diff
@@ -41,6 +42,7 @@ import (
 func (c *Controller) HandleVerify() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := observability.WithBuildInfo(r.Context())
+		ctx = logging.WithLogger(ctx, c.logger.Named("handleverify"))
```
sethvargo, line 45: Hmm - this should be:

```go
logger := c.logger.Named("HandleVerify")
```

and then update to use `logger` instead of `c.logger` in the rest of the method.

sethvargo (changes requested):

pkg/controller/verifyapi/verify.go
mikehelmick: removed - changed direction, don't need this

sethvargo (approved):
/lgtm

Issue #910, opened 2020-10-27
System admin - view mobile apps


mikehelmick:
### TL;DR

For coordinating EN Express (ENX) and understanding what should be happening with the ENX redirect service:

* Surface configured mobile apps to system admins

/kind enhancement

sethvargo:
Broken down by realm or just a total list?

mikehelmick:
Both views would be awesome

1) when viewing an individual realm, show this

2) view a list of all mobile apps configured in the system

PR #951, opened 2020-11-03
Show mobile apps in system admin  - Part 1

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/910

## Proposed Changes

* Gets the basic pages of mobile-apps into system admin
* Lists all mobile  apps

TODO
* currently action links / hrefs go back to realm admin
* Show realm name per app
* Might want pagination/filtering?

sysadminmobileapp

**Release Note**

```release-note
Show all mobile apps in system  admin
```

### DIFF
--- a/cmd/server/assets/admin/mobileapps/show.html
+++ b/cmd/server/assets/admin/mobileapps/show.html
@@ -1,0 +1,82 @@
+{{define "admin/mobileapps/show"}}
+
+{{$apps := .apps}}
+
+<!doctype html>
+<html lang="en">
+<head>
+  {{template "head" .}}
+</head>
+
+<body class="tab-content">
+  {{template "admin/navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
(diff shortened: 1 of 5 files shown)


cmd/server/assets/admin/mobileapps/show.html
```diff
@@ -0,0 +1,82 @@
...
+      <div class="card-body">
+
+        {{if .apps}}
+        <div class="table-responsive">
+          <table class="table table-bordered table-striped bg-white">
+            <thead>
+              <tr>
+                <th scope="col">App</th>
+                <th scope="col" width="95">Enabled</th>
+                <th scope="col" width="40"></th>
```
mikehelmick, line 28: I don't think we need enable/disable here 

the goal is to be able to see which mobile apps are configured for the redirect service

whaught: sure,  removed that column

mikehelmick (approved):

PR #952, opened 2020-11-03
Show realm name with mobile app

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/910

## Proposed Changes

* Merge  the  realm with mobile app, show the realm name in the list
    *  in-memory list merge

Next:  show as card in the system admin realm-edit page filtered to the realm

withrealm


cmd/server/assets/admin/mobileapps/show.html
```diff
@@ -24,6 +24,7 @@
             <thead>
               <tr>
                 <th scope="col">App</th>
+                <th scope="col">Realm</th>
```
sethvargo, line 27: This needs a width so App floats

pkg/controller/admin/mobile_apps.go
```diff
@@ -33,11 +38,30 @@ func (c *Controller) HandleMobileAppsShow() http.Handler {
 			return
 		}
 
-		c.renderShowMobileApps(ctx, w, apps)
+		realms, err := c.db.GetRealms()
```
sethvargo, line 41: I'm not a fan of this mutation happening in the controller layer (I think it belongs in the database layer). It might be worth timeboxing an hour or so with gorm to see if we can do this in the query somehow.

sethvargo (changes requested):

pkg/database/mobile_app.go
```diff
@@ -133,20 +133,30 @@ func (a *MobileApp) BeforeSave(tx *gorm.DB) error {
...
+		Rows()
+	if err != nil || rows == nil {
 		return nil, err
 	}
+	defer rows.Close()
+
+	apps := make([]*ExtendedMobileApp, 0)
+	for rows.Next() {
+		app := &ExtendedMobileApp{}
+		db.db.ScanRows(rows, &app)
```
sethvargo, line 156: Can this return an error?

pkg/database/mobile_app.go
```diff
@@ -133,20 +133,30 @@ func (a *MobileApp) BeforeSave(tx *gorm.DB) error {
...
+type ExtendedMobileApp struct {
+	MobileApp
+	Realm
+}
+
+// ListActiveAppsWithRealm finds all active mobile apps with their associated realm.
+func (db *Database) ListActiveAppsWithRealm() ([]*ExtendedMobileApp, error) {
+	rows, err := db.db.Table("mobile_apps").
+		Select("mobile_apps.*, realms.*").
+		Joins("left join realms on realms.id = mobile_apps.realm_id").
```
sethvargo, line 146: There's a subtle thing here... if a realm is deleted, but the associated mobile app is not, you'd have a NPE later.

sethvargo (approved):
/lgtm

pkg/database/mobile_app.go
whaught: I don't believe so. We always make an instance of ExtendedMobileApp and realm is inlined

PR #953, opened 2020-11-03
Add generic pagination functionality

sethvargo:
**Release Note**

```release-note
Add pagination helpers
```

### DIFF
--- a/pkg/controller/middleware/template.go
+++ b/pkg/controller/middleware/template.go
@@ -17,7 +17,6 @@ package middleware
 import (
 	"context"
 	"net/http"
-	"strings"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/buildinfo"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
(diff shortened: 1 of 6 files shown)


pkg/controller/middleware/template.go
```diff
@@ -40,9 +39,6 @@ func PopulateTemplateVariables(ctx context.Context, config *config.ServerConfig)
 			m["buildID"] = buildinfo.BuildID
 			m["buildTag"] = buildinfo.BuildTag
 
-			// Current page variables.
```
sethvargo, line 43: This was being overwritten by the other middleware, but I lost about 30min trying to debug why `currentPath` wasn't a string. Removing for the future.

pkg/database/pagination.go
```diff
@@ -0,0 +1,100 @@
...
+	if err := query.
+		Count(&total).
+		Error; err != nil {
+		return nil, err
+	}
+
+	// Calculate offset.
+	offset := uint64((page - 1) * limit)
+
+	// Always select one more than the limit so we can see if there are "more"
```
whaught, line 48:
```suggestion
	// Always select one more than the limit so we can see if there are "more"
```

Are we selecting an extra? We already did a full Count()

sethvargo: Ooops - this is from a previous design. Removed.

whaught (approved):

PR #958, opened 2020-11-03
Fix the alerts

yegle:
The incorrect one was deleted in
https://github.com/google/exposure-notifications-verification-server/pull/957


sethvargo (approved):
/lgtm

PR #968, opened 2020-11-03
Define probers alert in MQL.

### LINKED ISSUE #810: Convert our alerts to MQL

yegle:
Fixes #810


sethvargo (approved):
/lgtm

PR #973, opened 2020-11-03
[TESTONLY]

chaodaiG:


scripts/presubmit.sh
```diff
@@ -86,6 +86,9 @@ if [ -n "${OUT}" ]; then
   exit 1
 fi
 
+while true; do docker ps >>"${ARTIFACTS}/docker.out"; done &
```
sethvargo, line 89: Let's sleep for like 2-3s here just so we don't super-spew.

sethvargo, line 89: Can we make this per-run (like is there a build ID or something we can use)?

chaodaiG: This actually saves to a sub-directory of where the job log is located, it's pretty well separated from run to run

sethvargo:
/close 

we fixed it

PR #977, opened 2020-11-04
Single page for user index/search

whaught:
## Proposed Changes

* Keep index/search as one page with the search as an optional query param
* This seems simpler to maintain - @sethvargo is there a specific reason to split?


sethvargo:
Using `SearchUsers` on each page render is expensive since it adds the `WHERE ILIKE` clauses. That's why I made it separate. 

I also thought we might want a different view for searching in the future, instead of just index.

whaught:
> Using `SearchUsers` on each page render is expensive since it adds the `WHERE ILIKE` clauses. That's why I made it separate.
> 
> I also thought we might want a different view for searching in the future, instead of just index.

https://github.com/whaught/exposure-notifications-verification-server/blob/b31560801375627568050961b887550cbd8f39a3/pkg/database/realm.go#L807

We omit `WHERE ILIKE` for empty queries, but I suppose we could make the results page more interesting. However I think merging them might be simpler for now

sethvargo:
I think the difference is whether we want to _search_ or _filter_. Moving this into the index makes it a _filter_, but having it as its own page and function makes it a _search_ (where we could contextualize more information in the future).

sethvargo (commented):
Okay, I've come around to it being a single page, but let's rebase and use the pattern on db.ListUsers with scopes instead.

whaught:
/close

I can pull in gorm scopes, but I'ma trash this branch and do it over

Issue #975, opened 2020-11-03
System-admin level user managment


whaught:
### Currently
* We manage and search users within a realm scope
* Delete button to remove from current realm

### Proposal
* Mirror the users page / search in system admin with realm unscoped
* Option to remove from individual realms or all realms
* Option to 'disable' the user's login (in firebase auth rather than our User table)

### Why
Makes it easy to find a particular user in the system or disable an abusive user system-wide

PR #978, opened 2020-11-04
Create disable-user action

whaught:

## Proposed Changes

* API to disable a user's login

Next PR will integrate this into a UI for System admin

**Release Note**

```release-note
Button for system admins to disable login for a user
```


cmd/server/assets/users/show.html
```diff
@@ -37,15 +38,67 @@ <h1>{{$user.Name}}</h1>
...
+          {{else}}
+          <div class="card-text mb-3">Disabled</div>
+          {{end}}
         </div>
 
         <a href="/users/{{$user.ID}}/reset-password" data-method="POST" class="btn btn-primary btn-block">Send password reset</a>
       </div>
     </div>
 
+    {{if $currentUser.Admin}}
```
sethvargo, line 59: I'm -1 on including admin-related actions outside of the admin console. I think we should only have these pages for realm admins and system admin operations need to take place on the system admin console.

internal/auth/firebase.go
```diff
@@ -163,6 +163,23 @@ func (f *firebaseAuth) CreateUser(ctx context.Context, name, email, pass string,
 	return true, nil
 }
 
+// DisableUser prevents a user from logging in by disabling their account.
+func (f *firebaseAuth) DisableUser(ctx context.Context, email string) error {
```
sethvargo, line 167: Why do we need to disable the user in firebase versus doing a local disable in the database?

internal/routes/server.go
```diff
@@ -283,6 +283,13 @@ func Server(
 		userSub.Handle("/{id}", userController.HandleUpdate()).Methods("PATCH")
 		userSub.Handle("/{id}", userController.HandleDelete()).Methods("DELETE")
 		userSub.Handle("/{id}/reset-password", userController.HandleResetPassword()).Methods("POST")
+
+		userAdminSub := r.PathPrefix("/users").Subrouter()
```
sethvargo, line 287: I think this should be a system admin-only route.

sethvargo (changes requested):

internal/auth/firebase.go
whaught: if there were an abusive user this stops all access, it's also flippable so you could disable and then enable without having to re-join all their realms

sethvargo: We can set `DeletedAt` on the user in the database and achieve the same effect (without removing them from all the realms). I'm worried about a situation where the developer's firebase is attached to multiple systems. I dont think they'd expect disabling a user here to disable them in other systems.

whaught:
/close

I'll revisit pieces of this, but going forward with just delete instead of disable

PR #980, opened 2020-11-05
Split admin and regular users in system-admin. Allow delete user

whaught:
Issue #https://github.com/google/exposure-notifications-verification-server/issues/975

## Proposed Changes

* Creates a new users tab in system-admin
    * Keeps system-admin list and regular users separate
* Allow system admin to delete a user
* Searching and pagination

withsearch

**Release Note**

```release-note
Show regular users across realms in system-admin. Allow deletion.
```

### DIFF
--- a/pkg/controller/admin/superusers.go
+++ b/pkg/controller/admin/superusers.go
@@ -1,0 +1,221 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package admin
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/internal/auth"
+	"github.com/google/exposure-notifications-verification-server/internal/project"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/gorilla/mux"
+)
+
+// HandleSuperUsersIndex renders the list of system admins.
+func (c *Controller) HandleSuperUsersIndex() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		admins, err := c.db.ListSystemAdmins()
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		m := controller.TemplateMapFromContext(ctx)
+		m["admins"] = admins
+		c.h.RenderHTML(w, "admin/superusers/index", m)
+	})
+}
+
+// HandleSuperUsersCreate creates a new system admin.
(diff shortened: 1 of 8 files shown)


pkg/controller/admin/superusers.go
```diff
@@ -0,0 +1,221 @@
...
+		// Requested form, stop processing.
+		if r.Method == http.MethodGet {
+			var user database.User
+			c.renderNewUser(ctx, w, &user)
+			return
+		}
+
+		var form FormData
+		err := controller.BindForm(w, r, &form)
+		email := project.TrimSpace(form.Email)
```
sethvargo, line 78: This trimming and validation should happen in the model via a `BeforeSave`, not the controller.

pkg/controller/admin/users.go
```diff
@@ -15,125 +15,32 @@
...
 
-	"github.com/google/exposure-notifications-verification-server/internal/auth"
-	"github.com/google/exposure-notifications-verification-server/internal/project"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/gorilla/mux"
 )
 
-// HandleUsersIndex renders the list of system admins.
+// HandleUsersIndex renders the list of all non-system-admin usders.
```
sethvargo, line 25:
```suggestion
// HandleUsersIndex renders the list of all non-system-admin users.
```

pkg/database/user.go
```diff
@@ -223,6 +223,29 @@ func (u *User) Stats(db *Database, realmID uint, start, stop time.Time) ([]*User
 	return stats, nil
 }
 
+// DeleteUser deletes the user entry.
+func (db *Database) DeleteUser(u *User) error {
+	return db.db.Delete(u).Error
+}
+
+// ListUsers returns a list of all users sorted by name.
```
sethvargo, line 231: Maybe a warning about using this vs `realm.ListUsers`?

pkg/database/user.go
```diff
@@ -223,6 +223,29 @@ func (u *User) Stats(db *Database, realmID uint, start, stop time.Time) ([]*User
...
+func (db *Database) ListUsers() ([]*User, error) {
+	var users []*User
+	if err := db.db.
+		Model(&User{}).
+		Where("admin IS FALSE").
```
sethvargo, line 236: Why exclude admins?

pkg/database/user.go
```diff
@@ -223,6 +223,29 @@ func (u *User) Stats(db *Database, realmID uint, start, stop time.Time) ([]*User
...
+		Order("LOWER(name) ASC").
+		Find(&users).
```
sethvargo, line 238: Pagination?

sethvargo (changes requested):
This is looking good. I'd like us to avoid introducing a new term though. "super user" should be "system admin" everywhere, including routing, to avoid confusion.

pkg/database/user.go
whaught: There are two pages: Admin users and Users. The delete action deletes the user here, but the admin users delete button just de-privileges the user by removing the admin bit.
I'm considering merging things or moving actions to a sub-page together, but for now I'm trying to keep them separate

pkg/database/user.go
whaught: That's coming next!

pkg/controller/admin/superusers.go
whaught: That does also happen in BeforeSave, but FindUserByEmail needs the trim

pkg/database/user.go
whaught: aaand there it is

pkg/controller/admin/superusers.go
sethvargo: Why not trim in `FindUserByEmail` then?

PR #981, opened 2020-11-05
Optionally mirror dependent Docker Hub images

### LINKED ISSUE #959: Do something about new Docker Hub rate limits
Tests are sporadically failing due to new rate limits.

/assign

sethvargo:
Also allow customizing the image used in CI via an environment variable.

Fixes #959

**Release Note**

```release-note
Add Terraform module to optionally mirror dependent Docker Hub images for tests
```

/assign @chaodaiG

### DIFF
--- a/pkg/database/database_util.go
+++ b/pkg/database/database_util.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -63,11 +64,14 @@ func NewTestDatabaseWithCacher(tb testing.TB, cacher cache.Cacher) (*Database, *
 		tb.Fatalf("failed to create Docker pool: %s", err)
 	}
 
+	// Determine the container image to use.
+	repo, tag := postgresRepo(tb)
+
 	// Start the container.
 	dbname, username, password := "en-verification-server", "my-username", "abcd1234"
 	container, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository: "postgres",
-		Tag:        "12-alpine",
+		Repository: repo,
+		Tag:        tag,
 		Env: []string{
 			"LANG=C",
 			"POSTGRES_DB=" + dbname,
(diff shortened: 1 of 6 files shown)


sethvargo:
@chaodaiG to use this, we'll need to:

1. Choose a project to host the mirror (could be any of them)
1. Run the Terraform there and grant the Prow service account read permissions
1. Set the `CI_POSTGRES_IMAGE` environment variable to point to the Artifact Registry image instead

Any thoughts on which project we should use? I already ran the Terraform against one for testing, so we could use that one. It doesn't have to be the same project as Prow.

chaodaiG:
> @chaodaiG to use this, we'll need to:
> 
> 1. Choose a project to host the mirror (could be any of them)
> 2. Run the Terraform there and grant the Prow service account read permissions
> 3. Set the `CI_POSTGRES_IMAGE` environment variable to point to the Artifact Registry image instead
> 
> Any thoughts on which project we should use? I already ran the Terraform against one for testing, so we could use that one. It doesn't have to be the same project as Prow.

The GCP projects used for testing are not relevant in presubmit tests right now, the only cluster used for presubmit tests is the prow build cluster. So I would prefer let prow build cluster host these images. 

> 2. Run the Terraform there and grant the Prow service account read permissions

The prow service account used for test is the default compute engine service account for the project, I can grant it permission

sethvargo:
> So I would prefer let prow build cluster host these images.

That would involve a decent amount of setup on the prow projects.

> The prow service account used for test is the default compute engine service account for the project, I can grant it permission

See the PR in the -infra repo (I CCed you).

chaodaiG:
I see your point,  and agreed. we can use one of the test project for this, let's use `
apollo-boskos-key-terraform-02`

sethvargo:
@chaodaiG do we run the risk of that project being auto-deleted or cleaned up? I'd rather use one of our more static projects.

chaodaiG:
@sethvargo , actually you're right, there is a risk of cloud run service being cleaned up, but it's controlled by us. I wouldn't imagine the images being cleaned up though.

For static project, we don't have a different static project yet other than the prow build cluster, if desired we can set up another one, but it'll take some time for a reason.

PR #982, opened 2020-11-05
Move string trims out of controller

whaught:
## Proposed Changes

* Trims occur in BeforeSave for the user email/name
* Trim in the email template
* Trim in FindByEmail
* Trim search queries
* No more trim in the controller layer


sethvargo (approved):
/lgtm

PR #984, opened 2020-11-05
Rename Admin to SystemAdmin in users table

sethvargo:
Just a simple rename, but now that we're merging some functionality, I want to be clear when we say "system admin". This has also confused our partners in the past.

**Release Note**

```release-note
Rename users `admin` to `system_admin` for clarity.
```

/assign @whaught

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -247,7 +247,7 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 					logger.Debugw("adding user to default realm", "user", u.ID)
 
 					u.AddRealm(defaultRealm)
-					if u.Admin {
+					if u.SystemAdmin {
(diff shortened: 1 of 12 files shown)


pkg/database/migrations.go
```diff
@@ -1596,6 +1596,46 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+				return nil
+			},
+		},
+		{
+			ID: "00065-RenameUserAdminToSystemAdmin",
+			Migrate: func(tx *gorm.DB) error {
+				sqls := []string{
+					`
+					DO $$
+					BEGIN
```
sethvargo, line 1608: AutoMigrate striking again 😦

whaught (approved):
I like this - it always felt ambiguous with system/realm admin

PR #985, opened 2020-11-05
Show pills for realm/system admin

whaught:

## Proposed Changes

* Show pills for system/realm admins in the admin console
* Now shows all users for the Users page
    * Later I'll fold in the System admin stuff and remove that tab

pills

### DIFF
--- a/cmd/server/assets/admin/users/index.html
+++ b/cmd/server/assets/admin/users/index.html
@@ -44,7 +44,14 @@
             <tbody>
             {{range $users}}
               <tr>
-                <td>{{.Name}}</td>
+                <td>
+                  {{.Name}}
+                  {{if .Admin}}
+                  <span class="ml-1 badge badge-pill badge-primary">System admin</span>
+                  {{else if .IsRealmAdmin}}
+                  <span class="ml-1 badge badge-pill badge-secondary">Realm admin</span>
+                  {{end}}
+                </td>
                 <td>{{.Email}}</td>
                 <td class="text-center">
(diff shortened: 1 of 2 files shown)


cmd/server/assets/admin/users/index.html
```diff
@@ -44,7 +44,14 @@
             <tbody>
             {{range $users}}
               <tr>
-                <td>{{.Name}}</td>
+                <td>
+                  {{.Name}}
+                  {{if .Admin}}
+                  <span class="ml-1 badge badge-pill badge-primary">System admin</span>
+                  {{else if .IsRealmAdmin}}
```
sethvargo, line 51: I don't think this is if/else, a single user could be both

whaught: Idk if that's useful. A System Admin is likely both (and can join any realm as admin), so it seems a little redundant to show both.

mikehelmick: I think it's useful to show which realms a person is admin and member of

whaught: We will, but in this list it would be hard to show all the realm names etc.
This will become a link to a detail page (that looks more similar to the My Account page with a list of realms)

mikehelmick: my vote is the same as seth - show both pills in that case

whaught: well that's 2-1 then. Both pills shown.

mikehelmick (approved):

PR #990, opened 2020-11-05
Link to system admin from realm select

whaught:
## Proposed Changes

* Link directly to system-admin page from realm-select
   * Note that if the admin is not a member of any realms, we go straight to admin already. https://github.com/google/exposure-notifications-verification-server/blob/f76a3cea09d53d076a47a204ce9c948e183c8f35/pkg/controller/login/select.go#L50

* Remove "launch admin" from navigation if already there

straighttoadmin

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -91,30 +91,28 @@ aria-expanded="false" aria-label="Toggle navigation">
       {{if .currentRealm }}
       {{if .currentUser.CanAdminRealm .currentRealm.ID}}
       <h6 class="dropdown-header">Manage realm</h6>
-      <a class="dropdown-item" href="/apikeys">API keys</a>
-      <a class="dropdown-item" href="/mobile-apps">Mobile apps</a>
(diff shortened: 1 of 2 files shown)


cmd/server/assets/header.html
```diff
@@ -103,10 +103,12 @@ <h6 class="dropdown-header">Manage realm</h6>
       {{end}}
 
       {{if .currentUser.SystemAdmin}}
+      {{if not (.currentPath.IsDir "/admin")}}
```
mikehelmick, line 106: I don't see any reason to remove this - we don't do it w/ other pages

whaught: maybe "Launch" just feels funny to me when you're already there

mikehelmick: what if we link to specific pages here instead?

image

whaught: How about we just highlight the active nav item (see most recent commit)

mikehelmick (approved):

PR #992, opened 2020-11-06
Ensure emails are sanitized

whaught:
## Proposed Changes

* CodeQL doesn't like the email strings since they are user-input. Check they are sanitized.

Note: we are using email content-type: plain in our email header.


mikehelmick:
Do we need to sanitize the email body or the email address? or both?

whaught:
> Do we need to sanitize the email body or the email address? or both?

I think the email address is the only user-content, but this will make it hard to forget as we change the emailer. We also allow realm admins to override the template which may set the body.

mikehelmick (approved):

PR #995, opened 2020-11-06
System admin user details

whaught:

## Proposed Changes

* Show user details from the system-admin view
      * Hybrid of the realm user details page and the my-account page 
           * Card for basic info
           * Card to show realm membership / admin

details

There is more work to do here
1) Let system admin jump to realm-admin view (with stats)
2) Deduplicate a little bit
3) Remove-user-from-realm link

**Release Note**

```release-note
System admin user-details page
```


cmd/server/assets/admin/users/show.html
```diff
@@ -0,0 +1,69 @@
...
+
+<body id="users-show" class="tab-content">
+  {{template "admin/navbar" .}}
+
+  <main role="main" class="container">
+    {{template "flash" .}}
+
+    <h1>{{$user.Name}}</h1>
+    <p>
+      Here is information about the user.
```
jeremyfaller, line 19: Maybe leave this paragraph out?

cmd/server/assets/admin/users/show.html
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+
+        <h6 class="card-title">Email</h6>
+        <div class="mb-3 mt-n2">
+          {{$user.Email}}
+        </div>
+
+        <h6 class="card-title">Password</h6>
+        <div class="card-text mb-3 mt-n2">Password was last changed <span class="text-info">{{$user.PasswordAgeString}}</span>
+          ago</div>
```
jeremyfaller, line 37: s/ago/ago./   ????

cmd/server/assets/admin/users/show.html
```diff
@@ -0,0 +1,69 @@
...
+          {{range $realm := $user.Realms}}
+          <li class="list-group-item">
+            {{$realm.Name}}
+
+            {{range $admin := $user.AdminRealms}}
+            {{if eq $admin.ID $realm.ID}}
+            <span class="badge badge-primary">Admin</span>
+            {{end}}
+            {{end}}
+          </li>
```
jeremyfaller, line 59: I know you'll end up doing this, but do you want to link to things like disabling the user? Adding them additional realms, etc?

jeremyfaller (approved):
/hold
hold only if you want to address some of these comment

cmd/server/assets/admin/users/show.html
whaught: for some reason it looks funny (this is the only true sentence under the headers)

cmd/server/assets/admin/users/show.html
whaught: That is all coming soon, but I want to start a new PR for it. The trouble is that the regular link assumes currentUser and passes a realm - we need to pass both a user and a realm for the system-admin case

cmd/server/assets/admin/users/show.html
whaught: Done

cmd/server/assets/admin/users/show.html
whaught: I think having an explanatory subtitle is the pattern for all of the other pages. This one is pretty lame though - wish we had more to say.
I'm going to leave it for consistency

pkg/controller/admin/users_list.go
```diff
@@ -56,6 +56,30 @@ func (c *Controller) HandleUsersIndex() http.Handler {
...
+func (c *Controller) HandleUserShow() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		vars := mux.Vars(r)
+
+		// Pull the user from the id.
+		user, err := c.db.FindUser(vars["id"])
+		if err != nil {
+			if database.IsNotFound(err) {
+				controller.Unauthorized(w, r, c.h)
```
sethvargo, line 69: This _probably_ doesn't make as much sense here, since this is guarded by the system admin. We should probably return a not-found or something.

whaught: agreed

cmd/server/assets/admin/users/show.html
```diff
@@ -0,0 +1,72 @@
...
+
+    <h1>{{$user.Name}}</h1>
+    <p>
+      Here is information about the user.
+    </p>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Details</div>
+      <div class="card-body">
+        <h6 class="card-title">Name</h6>
```
sethvargo, line 25: Check the `<hr>` + simplified margins on https://github.com/google/exposure-notifications-verification-server/blob/main/cmd/server/assets/admin/realms/edit.html#L43 if you want to use that here.

whaught: hows that?

sethvargo (approved):
/lgtm

Issue #991, opened 2020-11-06
system admin: list which realms have abuse prevention on


mikehelmick:
### TL;DR

* list which realms have abuse prevention on
* showing the current quota settings is a nice to have

/kind enhancement
/assign @whaught

PR #997, opened 2020-11-06
Make it easy to select and admin as realm admin

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/991

## Proposed Changes

* We have a system-admin realm edit page, but most functionality of it is in realm admin. Rather than duplicate, allow the user to easily select the realm and admin as realm-admin.

wrench
(there's a nice hover tooltip there too)


cmd/server/assets/admin/realms/edit.html
```diff
@@ -24,11 +24,17 @@ <h1>Edit {{$realm.Name}}</h1>
     <div class="card mb-3 shadow-sm">
       <div class="card-header">Details</div>
       <div class="card-body">
-        <div class="alert alert-warning" class="mb-4">
+        <div class="alert alert-warning mb-4">
           Only a subset of realm fields are editable after creation. Work with
           the realm administrator to update these values.
         </div>
 
+        {{if $currentUser.CanAdminRealm $realm.ID}}
```
sethvargo, line 32: I think I'd prefer to have this in the header as an icon with a tooltip. If the user is not a member of the realm, the action should be disabled with helpful text telling them to join first. Otherwise, the tooltip should say "edit this realm in the realm view" or something.

sethvargo (commented):
I'm a bit concerned about having multiple paths to the same place.

cmd/server/assets/admin/realms/edit.html
whaught: Done

sethvargo (approved):
/lgtm

PR #999, opened 2020-11-06
Use scopes for filtering, update CSS

sethvargo:
Uses scopes for filtering instead of sentinel values. Also updates system admin UX to match realm pages.

**Release Note**

```release-note
Update system admin UX to match realm pages.
```


sethvargo:
screenshot-20201106-124115@2x
screenshot-20201106-124117@2x
screenshot-20201106-124123@2x
screenshot-20201106-124146@2x

whaught (approved):

PR #1000, opened 2020-11-06
Limit number of non-deleted KMS signing keys on a realm

### LINKED ISSUE #993: quota: limit KMS keys
### TL;DR

As a cost protection measure, limit the number of non-deleted KMS signing keys a realm can have

3 should be enough?

/kind enhancement

sethvargo:
Fixes #993

**Release Note**

```release-note
Limit number of non-deleted KMS signing keys on a realm (default: 5, configurable)
```

/assign @mikehelmick 

I went with 5 instead of 3 because the cost differential is ~$0.12 and I'd rather never be in the way of the end-user.

### DIFF
--- a/pkg/database/config.go
+++ b/pkg/database/config.go
@@ -55,6 +55,11 @@ type Config struct {
 	// created on.
 	CertificateSigningKeyRing string `env:"CERTIFICATE_SIGNING_KEYRING"`
 
+	// MaxCertificateSigningKeyVersions is the maximum number of certificate
+	// signing key versions per realm. This is enforced at the database layer, not
+	// the upstream KMS.
+	MaxCertificateSigningKeyVersions int64 `env:"MAX_CERTIFICATE_SIGNING_KEY_VERSIONS, default=5"`
+
 	// EncryptionKey is the reference to an encryption/decryption key to use when
 	// for application-layer encryption before values are persisted to the
 	// database.
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -1162,6 +1162,24 @@ func (r *Realm) CreateSigningKeyVersion(ctx context.Context, db *Database) (stri
 		return "", fmt.Errorf("missing key name")
 	}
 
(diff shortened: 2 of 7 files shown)


mikehelmick (changes requested):
remove bin files

pkg/database/realm.go
```diff
@@ -1162,6 +1162,23 @@ func (r *Realm) CreateSigningKeyVersion(ctx context.Context, db *Database) (stri
...
+		Table("signing_keys").
+		Where("realm_id = ?", r.ID).
+		Count(&count).
+		Error; err != nil {
+		if !IsNotFound(err) {
+			return "", fmt.Errorf("failed to count existing signing keys: %w", err)
+		}
+	}
+	if max := db.config.MaxCertificateSigningKeyVersions; count >= max {
+		return "", fmt.Errorf("too many active signing keys (maximum: %d)", max)
```
mikehelmick, line 1179: I don't think "active" is right here, because only one can be active.

pkg/database/config.go
```diff
@@ -55,6 +55,11 @@ type Config struct {
 	// created on.
 	CertificateSigningKeyRing string `env:"CERTIFICATE_SIGNING_KEYRING"`
 
+	// MaxCertificateSigningKeyVersions is the maximum number of certificate
+	// signing key versions per realm. This is enforced at the database layer, not
+	// the upstream KMS.
+	MaxCertificateSigningKeyVersions int64 `env:"MAX_CERTIFICATE_SIGNING_KEY_VERSIONS, default=5"`
```
mikehelmick, line 61: can we add a warning if a realm already has > 5

pkg/database/realm.go
sethvargo: Fixed

pkg/database/config.go
sethvargo: Added to the UI

sethvargo:
@mikehelmick PTAnotherL

mikehelmick (approved):

PR #1002, opened 2020-11-06
Ensure bits don't overflow

sethvargo:
All our systems are 64-bit, so uint == uint64. However, on 32-bit systems, we would overflow.

**Release Note**

```release-note
Ensure bits don't overflow when running on 32-bit systems
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/bits"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


whaught (approved):

pkg/database/database.go
```diff
@@ -315,6 +316,11 @@ func callbackIncrementMetric(ctx context.Context, m *stats.Int64Measure, table s
 					_ = scope.Err(fmt.Errorf("failed to parse realm_id: %w", err))
 					return
 				}
+
+				if raw >= 1<<bits.UintSize-1 {
```
jeremyfaller, line 320: https://golang.org/pkg/math/#pkg-constants

sethvargo: Right, that tells me the max 32 and 64-bit uints, but it doesn't tell me what the max value of `uint` is on the system, since `int` and `uint` vary based on the architecture (32 vs 64 bit)

PR #1006, opened 2020-11-06
Switch user search over to scopes

whaught:
## Proposed Changes

* Uses scopes for realm-user query
* Merges the index/search pages

re-do of https://github.com/google/exposure-notifications-verification-server/pull/977


sethvargo (approved):
/lgtm

PR #1007, opened 2020-11-06
Allow admins to add/remove users from other realms

### LINKED ISSUE #1003: Delete-from-realm link in system-admin user page
### Summary
We recently added a user-details page to system admin which shows a list of realms. Let's add in a link to 'remove from realm' similar to the my-account page.

leaverealm

Currently we have:
`/admin/realms/{{$realm.ID}}/leave`
and
`/users/{{.ID}} data-method="DELETE"`

One assumes currentUser and specifies realmId, the other assumes currentRealm and specifies userId.
It would be nice to not assume anything for the /admin path.

We'd want to add the link here: https://github.com/google/exposure-notifications-verification-server/blob/86df37ac5548473b62ca0be36663242812572d90/cmd/server/assets/admin/users/show.html#L55

sethvargo:
Fixes #1003

**Release Note**

```release-note
Allow admins to add/remove users from other realms
```

### DIFF
--- a/cmd/server/assets/admin/users/show.html
+++ b/cmd/server/assets/admin/users/show.html
@@ -47,24 +47,30 @@
       </div>
     </div>
 
-    <div class="card mb-3 shadow-sm">
-      <div class="card">
+    {{if $user.Realms}}
+      <div class="card mb-3 shadow-sm">
         <div class="card-header">Member of realms</div>
         <ul class="list-group list-group-flush">
           {{range $realm := $user.Realms}}
-          <li class="list-group-item">
-            {{$realm.Name}}
+          <li class="list-group-item d-flex justify-content-between">
+            <div>
+              {{$realm.Name}}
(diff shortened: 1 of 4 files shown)


cmd/server/assets/admin/users/show.html
```diff
@@ -47,24 +47,30 @@ <h6 class="mb-2">System admin</h6>
       </div>
     </div>
 
-    <div class="card mb-3 shadow-sm">
-      <div class="card">
+    {{if $user.Realms}}
```
whaught, line 50: Alternatively leave the card and
```
{{else}
This user is not a member of any realms
```

might make the UI more consistent / clear

whaught (commented):
/lgtm
/hold depending on how you feel about my other comment

cmd/server/assets/admin/users/show.html
sethvargo: Done

whaught (approved):

PR #1011, opened 2020-11-07
E2e improvements

mikehelmick:
## Proposed Changes

* Fix race condition in e2e-runner
* Actually generate an HMAC secret (instead of using 00000...)
* make both e2e runners run "at the same time" some of the time

**Release Note**

```release-note
e2e-runner improvements, fixes race condition
```

### DIFF
--- a/terraform/service_e2e_runner.tf
+++ b/terraform/service_e2e_runner.tf
@@ -190,7 +190,7 @@ resource "google_cloud_run_service_iam_member" "e2e-runner-invoker" {
 resource "google_cloud_scheduler_job" "e2e-default-workflow" {
   name             = "e2e-default-workflow"
   region           = var.cloudscheduler_location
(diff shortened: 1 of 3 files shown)


terraform/service_e2e_runner.tf
```diff
@@ -190,7 +190,7 @@ resource "google_cloud_run_service_iam_member" "e2e-runner-invoker" {
 resource "google_cloud_scheduler_job" "e2e-default-workflow" {
   name             = "e2e-default-workflow"
   region           = var.cloudscheduler_location
-  schedule         = "0,10,20,30,40,50 * * * *"
+  schedule         = "0,10,20,30,40,50,55 * * * *"
```
yegle, line 193: */5 ?

mikehelmick: the idea was that they both run at the time at 0 and 55 and run on 5 minute offsets the rest of the time.

PR #1012, opened 2020-11-07
Add unique request_id to context and logger

### LINKED ISSUE #1009: Add request id middleware and add request id to logs
### TL;DR
In a single HTTP request handling code, there will be multiple log lines produced from various places.

It would be great if there's a unique "request id" for each incoming request, and then we can search for this string if we want to get all logged lines for that request.

sethvargo:
Fixes #1009

**Release Note**

```release-note
Add unique request_id for logs
```

### DIFF
--- a/pkg/controller/healthz.go
+++ b/pkg/controller/healthz.go
@@ -20,10 +20,9 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
-
-	"github.com/google/exposure-notifications-server/pkg/logging"
 
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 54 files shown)


pkg/controller/healthz.go
```diff
@@ -20,23 +20,23 @@ import (
...
 )
 
 var (
 	rl = rate.NewLimiter(rate.Every(time.Minute), 1)
 )
 
-func HandleHealthz(hctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
-	logger := logging.FromContext(hctx).Named("healthz")
-
+func HandleHealthz(ctx context.Context, cfg *database.Config, h *render.Renderer) http.Handler {
```
yegle, line 34: The `ctx` here is not actually used right? In the body of the returned function, `ctx` is from `r.Context()`.

sethvargo: Yea, removed. Still confused about this data race though...

pkg/controller/middleware/apikey.go
```diff
@@ -111,7 +109,7 @@ func RequireAPIKey(ctx context.Context, cacher cache.Cacher, db *database.Databa
 			// Save the authorized app on the context.
 			ctx = controller.WithAuthorizedApp(ctx, &authApp)
 			ctx = controller.WithRealm(ctx, &realm)
-			*r = *r.WithContext(ctx)
+			r = r.Clone(ctx)
```
mikehelmick, line 112: I think these should go back to r.WithContext since the deep copy isn't necessary

mikehelmick (approved):

PR #1015, opened 2020-11-09
Fix missing commas

sethvargo:
/cc @yegle 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/terraform/alerting/verification-server/alerts.tf
+++ b/terraform/alerting/verification-server/alerts.tf
@@ -47,8 +47,9 @@ resource "google_monitoring_alert_policy" "RealmTokenRemainingCapacityLow" {
     mime_type = "text/markdown"
   }
   notification_channels = [
-    google_monitoring_notification_channel.email.id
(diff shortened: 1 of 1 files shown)


terraform/alerting/verification-server/alerts.tf
```diff
@@ -268,11 +273,11 @@ EOT
   }
 
   notification_channels = [
-    google_monitoring_notification_channel.email.id
+    google_monitoring_notification_channel.email.id,
   ]
+
   depends_on = [
-    null_resource.manual-step-to-enable-workspace
+    null_resource.manual-step-to-enable-workspace,
```
sethvargo, line 280: This one in particular is actual failing the run

chaodaiG:
Curious why it was not caught by the smoke test this morning https://prow.k8s.io/view/gcs/oss-prow/logs/ci-en-verification-server-terraform-smoke/1325810403162198016

sethvargo:
@chaodaiG because everything is pinned to the old release branch.

PR #1017, opened 2020-11-09
Only show 5 next/prev pages

sethvargo:
10 rolls off the page when there's a lot of pages.

**Release Note**

```release-note
Only show 5 next/prev pages in pagination.
```

### DIFF
--- a/pkg/database/pagination.go
+++ b/pkg/database/pagination.go
@@ -21,6 +21,12 @@ import (
 	"github.com/jinzhu/gorm"
 )
 
+const (
+	// maxPrevPages and maxNextPages control the "window" to show.
+	maxPrevPages = 5
+	maxNextPages = 5
+)
+
 // Paginate is a helper that paginates a gorm query into the given result. In
 // addition to reflecting into the provided result, it returns a pagination
 // struct.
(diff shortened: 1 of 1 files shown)


whaught (approved):

pkg/database/pagination.go
```diff
@@ -66,13 +66,13 @@ func PaginateFn(query *gorm.DB, page, limit uint64, populateFn func(query *gorm.
 		remaining := total - current
 		nextPages = uint64(math.Ceil(float64(remaining) / float64(limit)))
 		if nextPages > 10 {
```
whaught, line 68: a `const` might be good

does this need to be `nextPages > 5`

sethvargo:
@whaught good call on constants and that check. Fixed. PTAnotherL.

whaught (approved):

pkg/database/pagination.go
```diff
@@ -14,6 +14,12 @@
 
 package database
 
+const (
```
whaught, line 17: I think go wants this below the imports

sethvargo: Oh yea, fixed 😄

whaught (approved):

PR #1018, opened 2020-11-09
Allow searching API keys

sethvargo:
/assign @whaught 

This became obviously necessary on the e2e test project where there are thousands of keys waiting to be cleaned up.

**Release Note**

```release-note
Allow searching API keys
```


whaught (approved):
/lgtm

I think we have search for all of our Lists now

PR #1025, opened 2020-11-09
Show event log across realms in system-admin

whaught:
## Proposed Changes

* Adds the event log to system admin scoped across all realms

TODO: next PR will add in a `Realm` column

systemaudit

**Release Note**

```release-note
Show event logs in system-admin
```

### DIFF
--- a/pkg/database/audit_entry.go
+++ b/pkg/database/audit_entry.go
@@ -16,6 +16,8 @@ package database
 
 import (
 	"time"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/pagination"
 )
 
 // AuditEntry represents an event in the system. These records are purged after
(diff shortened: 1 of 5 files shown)


pkg/database/audit_entry.go
```diff
@@ -75,3 +77,28 @@ func (db *Database) PurgeAuditEntries(maxAge time.Duration) (int64, error) {
 		Delete(&AuditEntry{})
 	return result.RowsAffected, result.Error
 }
+
+// ListAudits returns the list audit events which match the given criteria.
+// Warning: This list may be large. Use Realm.Audits() to get users scoped to a realm.
+func (db *Database) ListAudits(p *pagination.PageParams, scopes ...Scope) ([]*AuditEntry, *pagination.Paginator, error) {
```
sethvargo, line 83: Something we should do (eventually): `realm.ListAudits` should call `db.ListAudits` with an additional scope that limits it to the realm id. And then all the other realm methods + db methods.

sethvargo (approved):
/lgtm

pkg/database/audit_entry.go
sethvargo: @whaught ^

PR #1027, opened 2020-11-09
Move realm db lists to ref system-wide queries

whaught:
## Proposed Changes

* The realm.ListUsers and realm.ListEvents queries are duplicated in the system scope. This makes the reference each other for reduced duplication/complexity

### DIFF
--- a/pkg/database/audit_entry.go
+++ b/pkg/database/audit_entry.go
@@ -78,9 +78,9 @@ func (db *Database) PurgeAuditEntries(maxAge time.Duration) (int64, error) {
 	return result.RowsAffected, result.Error
 }
 
-// ListSystemAudits returns the list audit events which match the given criteria.
+// ListAllAudits returns the list audit events which match the given criteria.
 // Warning: This list may be large. Use Realm.Audits() to get users scoped to a realm.
-func (db *Database) ListSystemAudits(p *pagination.PageParams, scopes ...Scope) ([]*AuditEntry, *pagination.Paginator, error) {
+func (db *Database) ListAllAudits(p *pagination.PageParams, scopes ...Scope) ([]*AuditEntry, *pagination.Paginator, error) {
 	return db.ListAudits(nil, p, scopes...)
 }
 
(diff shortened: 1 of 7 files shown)


pkg/database/audit_entry.go
```diff
@@ -78,16 +78,25 @@ func (db *Database) PurgeAuditEntries(maxAge time.Duration) (int64, error) {
...
+// ListAudits returns the list audit events which match the given criteria.
+func (db *Database) ListAudits(r *Realm, p *pagination.PageParams, scopes ...Scope) ([]*AuditEntry, *pagination.Paginator, error) {
 	var entries []*AuditEntry
 
 	query := db.db.
 		Model(&AuditEntry{}).
 		Scopes(scopes...).
 		Order("created_at DESC")
 
+	if r != nil {
```
sethvargo, line 96: Ah, I didn't explain myself much 😄 

In `realm.go`, I think we want to do this:

```go
func (r *Realm) ListAudits(db *Database, p *pagination.PageParams, scopes ...Scope) (whatever) {
  scopes = scopes.Append(func(db *gorm.DB) *gorm.DB {
    return db.Where("audit_entries.realm_id = ?", r.ID)
  })

  return db.ListAudits(p, scopes...)
}
```

Does that make sense?

sethvargo (changes requested):

pkg/database/realm.go
```diff
@@ -583,27 +583,11 @@ func (r *Realm) EmailProvider(db *Database) (email.Provider, error) {
...
-
-	paginator, err := Paginate(query, &entries, p.Page, p.Limit)
-	if err != nil {
-		if IsNotFound(err) {
-			return entries, nil, nil
-		}
-		return nil, nil, err
-	}
+	scopes = append(scopes, func(db *gorm.DB) *gorm.DB {
+		return db.Where("realm_id = ?", r.ID)
```
sethvargo, line 587:
```suggestion
		return db.Where("audit_entries.realm_id = ?", r.ID)
```

sethvargo (approved):
/lgtm
/hold

I'm worried about `realm.Users()`  conflicting with a `Users` property on the realm. It's probably fine, but just wanted to flag that we can't have a public function and public property with the same name.

sethvargo (approved):
/lgtm

PR #1034, opened 2020-11-10
Remove metrics descriptor definitions in Terraform.

yegle:
Metric descriptors cannot be updated, and can only be destroyed &
recreated. This is problematic because the running job is still
exporting metrics data point while the metric descriptors are being
destroyed and recreated.

We should always use the metric definition in the Go code as source of
truth.

/cc @sethvargo


sethvargo (approved):
/lgtm

PR #1039, opened 2020-11-10
Make the magnifying-glass search icon clickable everywhere

whaught:
## Proposed Changes

* Where we are allowing searches we have a magnifying-glass icon that looks like a button... make it clickable so it does what you might expect.
* Adds one for the system-admin users search


mikehelmick (approved):

sethvargo (commented):
Hmm... I don't like this. It's weird to have the `submit` button _before_ the form. It's also the first tab-index, which is unlikely to be what people expect.

PR #1040, opened 2020-11-10
All / System events filter + link from realms

whaught:
## Proposed Changes

* Links from the system-admin realm page to event logs for that realm
* Shows filter as a removable badge
* Adds a filter for "All events" and "System events"

systemfilter


sethvargo (changes requested):
I like the idea, but I think I'd rather force people to join the realm and view the realm's event page. I also wonder if we should change the system events page to be _only_ system events.

whaught:
> I like the idea, but I think I'd rather force people to join the realm and view the realm's event page. I also wonder if we should change the system events page to be _only_ system events.

I think this is pretty cool - we've sort of done this with mobile_apps and users shows both all and admin! Added in a filter for only system events

pkg/controller/admin/events.go
```diff
@@ -29,6 +30,9 @@ const (
 
 	// QueryToSearch is the query key for an ending time.
 	QueryToSearch = "to"
+
+	// QueryRealmIDSearch is the query key to filter by realmID.
+	QueryRealmIDSearch = "realmid"
```
sethvargo, line 35:
```suggestion
	QueryRealmIDSearch = "realm_id"
```

Maybe?

pkg/controller/admin/events.go
```diff
@@ -53,16 +60,30 @@ func (c *Controller) HandleEventsShow() http.Handler {
 			return
 		}
 
-		c.renderEvents(ctx, w, events, paginator, from, to)
+		var realm *database.Realm
+		if realmID == "0" {
```
sethvargo, line 64:
```suggestion
		if realmID == "0" || realmID == "" {
```

pkg/database/realm.go
```diff
@@ -587,10 +587,7 @@ func (r *Realm) EmailProvider(db *Database) (email.Provider, error) {
 
 // ListAudits returns the list audit events which match the given criteria.
 func (r *Realm) ListAudits(db *Database, p *pagination.PageParams, scopes ...Scope) ([]*AuditEntry, *pagination.Paginator, error) {
-	scopes = append(scopes, func(db *gorm.DB) *gorm.DB {
-		return db.Where("audit_entries.realm_id = ?", r.ID)
-	})
-
+	scopes = append(scopes, WithAuditRealmID(fmt.Sprint(r.ID)))
```
sethvargo, line 590:
```suggestion
	scopes = append(scopes, WithAuditRealmID(strconv.FormatUint(r.ID, 10)))
```

pkg/database/scopes.go
```diff
@@ -69,6 +69,17 @@ func WithAuditSearch(from, to string) Scope {
 	}
 }
 
+// WithAuditRealmID returns a scope that adds querying for Audit events by realm.
+func WithAuditRealmID(r string) Scope {
```
sethvargo, line 73: I'm not a huge fan of this accepting a string. I'd rather accept an interface{} or uint.

sethvargo (changes requested):

sethvargo (approved):
/lgtm

PR #1050, opened 2020-11-10
Example test

sethvargo:
🛑  DNM
/hold

/cc @jeremyfaller

### DIFF
--- a/internal/routes/server.go
+++ b/internal/routes/server.go
@@ -18,7 +18,10 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"path/filepath"
+	"testing"
+	"time"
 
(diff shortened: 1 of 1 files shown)


internal/routes/server.go
```diff
@@ -383,3 +386,38 @@ func Server(
...
 }
+
+func AdminRoutes(TODO) *mux.Router {
+	r := mux.NewRouter()
+	r.Use(someMiddleware) // will need to figure out this plumbing
+
+	adminSub := r.PathPrefix("/admin").Subrouter()
+	r.Use(someMiddleware) // will need to figure out this plumbing
+
+	adminController := admin.New(ctx, cfg, cacher, db, authProvider, limiterStore, h) // or maybe take this as a param
```
sethvargo, line 397: Alternatively, the admincontroller package could define its own routes

PR #1051, opened 2020-11-10
Copy adjustments for consistency and specificity

greggles:
## Proposed Changes

* On the code status page there are two expirations shown for folks using texts. This PR gives precise labels to the 2 different expiration times.
* On the reporting page, a graph title was inaccurate (the graph shows codes issued and claimed) and one graph includes API-issued codes while the other doesn't. It seems good to explicitly state that.

**Release Note**

NONE


greggles:
I'm not sure why the two old commits (10ff07d and 226b9dc) are showing up in this PR. I tried merging the upstream prior to beginning this work. Since they were already merged they don't show up in the diffs, so I think it's OK. Happy to try to fix that if you can advise me on what to do.

mikehelmick:
/ok-to-test
/lgtm
/approve

We use squash merges, so don't worry about the stray commits.

thanks!

Issue #1056, opened 2020-11-11
Export users to CSV


whaught:
### Export users to CSV
We currently accept a .csv file for creating users, we should create a link to easily export a realm to a .CSV
This can help move folks from custom servers to EN Express (or any other realm)

PR #1057, opened 2020-11-11
Export users to CSV

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1056

## Proposed Changes

* Create a link to export realm users to CSV
* just `email, usernamae` does not preserve realm admin status or other attributes

**Release Note**

```release-note
Link to download realm users as CSV
```


pkg/controller/user/export.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		users, _, err := realm.ListUsers(c.db, pageParams, []database.Scope{}...)
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		m := controller.TemplateMapFromContext(ctx)
+		m["users"] = users
+		b, err := c.h.RenderEmail("users/export", m)
```
sethvargo, line 47: I'd prefer to use go's built-in CSV package and render directly into `w` instead of the interim template. It's also weird this is `RenderEmail`.

sethvargo (changes requested):

pkg/controller/user/export.go
whaught: I've moved it to its own RenderCSV function, but I'll take a look at go's CSV package

pkg/controller/user/export.go
```diff
@@ -44,14 +44,9 @@ func (c *Controller) HandleExport() http.Handler {
 
 		m := controller.TemplateMapFromContext(ctx)
 		m["users"] = users
-		b, err := c.h.RenderEmail("users/export", m)
-		if err != nil {
+		if err := c.h.RenderCSV(w, "users/export", m); err != nil {
```
sethvargo, line 47: https://golang.org/pkg/encoding/csv/

```go
w.Header.Set(...)
csvw := csv.NewWriter(w)
csvw.Write(...)
```

cmd/server/assets/users/import.html
```diff
@@ -67,6 +67,17 @@ <h1>Import users</h1>
...
       </div>
     </div>
+
+    <div class="card mb-3 shadow-sm">
+      <div class="card-header">Import</div>
+      <div class="card-body">
+        <p>
+          Use the following link to export a list of users.
+        </p>
+        <a class="card=link" href="/realm/users/export.csv">export.csv</a>
```
jeremyfaller, line 77: `target="_blank"` ??

jeremyfaller (approved):
/hold

cmd/server/assets/users/import.html
whaught: Done. That said the content-disposition header should cause the browser to download rather than try to display this content.

PR #1063, opened 2020-11-11
Skip invitations checkbox for bulk-import

whaught:
## Proposed Changes

* Allows the user to skip sending password-reset invitations when bulk-importing users
* This would be used, for example, if a state moving to ENX wants to import firebase auth (and doesn't want to send reset-passwords)

**Release Note**

```release-note
Option to skip password-reset invitations on bulk user import
```

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -54,7 +54,7 @@ type Provider interface {
 	CreateUser(ctx context.Context, name, email, pass string, sendInvite bool, composer InviteUserEmailFunc) (bool, error)
 
 	// SendResetPasswordEmail resets the given user's password. If the user does not exist,
-	// the underlying provider determines whether its an error or perhaps upserts
+	// the underlying provider determines whether it's an error or perhaps upserts
 	// the account.
 	SendResetPasswordEmail(ctx context.Context, email string, composer ResetPasswordEmailFunc) error
 
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -51,10 +51,10 @@ type Provider interface {
 
 	// CreateUser creates a user in the auth provider. If pass is "", the provider
 	// creates and uses a random password.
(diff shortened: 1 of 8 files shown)


internal/auth/auth.go
```diff
@@ -51,10 +51,10 @@ type Provider interface {
 
 	// CreateUser creates a user in the auth provider. If pass is "", the provider
 	// creates and uses a random password.
-	CreateUser(ctx context.Context, name, email, pass string, composer InviteUserEmailFunc) (bool, error)
+	CreateUser(ctx context.Context, name, email, pass string, sendInvite bool, composer InviteUserEmailFunc) (bool, error)
 
 	// SendResetPasswordEmail resets the given user's password. If the user does not exist,
-	// the underlying provider determines whether itls an error or perhaps upserts
+	// the underlying provider determines whether its an error or perhaps upserts
```
jeremyfaller, line 57: it's

internal/auth/firebase.go
```diff
@@ -145,18 +145,20 @@ func (f *firebaseAuth) CreateUser(ctx context.Context, name, email, pass string,
...
-		if err := f.firebaseInternal.SendNewUserInvitation(ctx, email); err != nil {
-			return true, fmt.Errorf("failed to send new user invitation firebase email: %w", err)
+			if err := emailer(ctx, inviteLink); err != nil {
+				return true, fmt.Errorf("failed to send new user invitation email: %w", err)
+			}
+		} else {
+			if err := f.firebaseInternal.SendNewUserInvitation(ctx, email); err != nil {
+				return true, fmt.Errorf("failed to send new user invitation firebase email: %w", err)
+			}
 		}
```
jeremyfaller, line 162: feels like we should error here.

jeremyfaller (approved):
/hold

jeremyfaller (approved):

PR #1066, opened 2020-11-11
Add log-based metric to monitor OpenCensus stackdriver export.

### LINKED ISSUE #1065: Add an alert on OpenCensus metric export errors
See https://github.com/google/exposure-notifications-verification-server/issues/1064.

We should have an alert on the metric export issue.

It has to be a log based alert since we can't rely on OpenCensus on this one.

yegle:
Also created an alert if the error rate is exceeding 0.1/s.

Fixes #1065


sethvargo (approved):
/lgtm

PR #1067, opened 2020-11-11
Incremental e2e tests: use latest service for running tests

chaodaiG:
Just discovered that incremental e2e test had been creating images and migrating db, but didn't really deploy new services, so the test had been testing the old services from last month. This PR should fix it

/cc @sethvargo @hueich


sethvargo:
Note this will increase the test time significantly though

chaodaiG:
Ack, unfortunately images are built twice

Issue #1061, opened 2020-11-11
Bulk code issuance


mikehelmick:
### TL;DR

Write a command line tool that process a CSV file in the format of

```
phoneNumber,testDate
```

i.e. 

```
+12065555555,2020-11-10
+13155555555
```

### Design

Needs an admin API key, admin API endpoint, and request rate.
The tool should respect 429s and backoff to quota reset points as well.

/kind enhancement

whaught:
/assign

Taking this one. After some discussion, I think folks are okay with attempting to do this not as a CLI tool, but as a javascript one. This will make it easier to ship and maintain, and lowers the bar to easy use. The downside is we won't be able to write back to files and that sort of thing.

whaught:
I'm going to /close this as done. There may be more to add with error handling and additional fields, but I want to hear that user-feedback before doing more

PR #1069, opened 2020-11-12
Bulk issue

whaught:

## Proposed Changes

* Create a page to read a CSV file and bulk issue codes
* Generates a private-key ("retry code") in javascript for the client
    * Hashes all phone#s with the private-key to a UUID for tracking. This prevents re-upload of the same phone twice.
    * Optionally saves the retry code as a cookie
* Stores start-at-line N for easy resume

bulky2

TODO:
* Batch API that takes []IssueRequest
* Pay attention to status conflict and insufficient quota
* Show errors as a table instead of dumping as a toast
    * Potentially try to find malformed input before starting the upload
* Configurable delay for rate-limiting's sake?
* Consider if we need more fields for the CSV
* Consider if we should support JSON (seems unlikely)


mikehelmick:
Couple of questions

1) Do we have to work about any browser side timeouts / limitations?
2) Think about how we report individual line items errors back to the users (i.e. line 27 contains a land line)
3) retry stratergy?

whaught:
> Couple of questions
> 
> 1. Do we have to work about any browser side timeouts / limitations?
Well each is a separate request (there's no batch API so it's just doing it one-by-one). Each should be quick but the whole thing won't time-out. There is a progress bar and it is cancellable. 

> 2. Think about how we report individual line items errors back to the users (i.e. line 27 contains a land line)
True. We can put it in alert, but the bigger question might be around returning back a list of the failed ones (so they can be retried or corrected)

> 3. retry strategy?
Also we could introduce rate-limiting per the bug.

In general I just wanted to show this to open up a discussion as an alternative to the the CLI idea: it's more integrated and easier to use, but doesn't upload the CSV as a file in a way where pii storage becomes a concern.

sethvargo:
Architecturally, I think it would be better to background this:

1. Create a new table `batch_codes(id INTEGER, data TEXT, user_id INTEGER, realm_id INTEGER, status ENUM, timestamps...)`
1. On submission, insert the entire data into the database
1. New "worker" that pulls from that table and does work in batches (similar to cleanup)
1. Cloud Scheduler invokes worker frequently
1. Landing page to check the status of batch codes

sethvargo (changes requested):
I don't think this architecture will work. Most notably, those AJAX requests will count against rate limiting quota, so the max CSV size is rather limited.

whaught:
> Architecturally, I think it would be better to background this:
> 
> 1. Create a new table `batch_codes(id INTEGER, data TEXT, user_id INTEGER, realm_id INTEGER, status ENUM, timestamps...)`
> 2. On submission, insert the entire data into the database
> 3. New "worker" that pulls from that table and does work in batches (similar to cleanup)
> 4. Cloud Scheduler invokes worker frequently
> 5. Landing page to check the status of batch codes

The change I was going to make is create a batch issue API that takes more than one request.
The question about the above architecture is what to do with phone#s - we don't want to store them.

mikehelmick:
Currently storing phone numbers is a no-go for us - we've made that statement over and over again with jurisdictions that have already launched.

perhaps a live design conversation is in order for this

cmd/server/assets/home.html
```diff
@@ -35,13 +35,21 @@
...
+    <h1>Create verification codes
+    </h1>
     <p>
       Complete the following form to issue a single-use token to verify a
       patient. Do not submit this form until you are prepared to relay
       the code to the patient.
     </p>
 
+    <p>
+      <a href="/home/bulkissue" class="mr-3 text-secondary" data-toggle="tooltip" title="Bulk import users">
```
sethvargo, line 47:
```suggestion
      <a href="/home/bulk-issue" class="mr-3 text-secondary" data-toggle="tooltip" title="Bulk issue codes">
```

cmd/server/assets/codestatus/bulkissue.html
```diff
@@ -41,6 +45,19 @@
           This form will read the CSV contents and call the issue API one-by-one.
         </p>
         <form id="form">
+          <div class="form-label-group">
+            <input type="text" class="form-control" id="retry-code" placeholder="Retry code" required>
```
sethvargo, line 49: From a security perspective, I think it'd be better for us to generate this and instruct the user to save it.

whaught: We generate a value for the field in javascript below. I want to ensure the value does not come from the server. The field needs to be editable so the user could replace it with an already-save code?

internal/routes/server.go
```diff
@@ -229,6 +229,7 @@ func Server(
 		// API for creating new verification codes. Called via AJAX.
 		issueapiController := issueapi.New(ctx, cfg, db, limiterStore, h)
 		sub.Handle("/issue", issueapiController.HandleIssue()).Methods("POST")
+		sub.Handle("/bulk-issue", issueapiController.HandleBulkIssue()).Methods("GET")
```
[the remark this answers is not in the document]
whaught: This doesn't use anything that can't be done using the regular issue one-by-one (it's just faster), so it's permission structure doesn't necessarily need to be any different. My fear here would be that folks over-privilege health authorities or ELR people (who are now able to mess with the realm's settings and users).

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+      }
+      $retryCode.val(randomString);
+
+      let totalCreated = 0;
+      let upload = readFile();
+
+      $table.hide();
+
+      {{if $hasSMSConfig}}
+      if (typeof (FileReader) == "undefined") {
```
sethvargo, line 136:
```suggestion
      if (typeof (FileReader) === "undefined") {
```

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+      {{end}}
+
+      $csv.change(function(file) {
+        let fileName = file.target.files[0].name;
+        $fileLabel.html(fileName);
+        $import.prop('disabled', false);
+      });
+
+      $cancel.on('click', function(event) {
+        upload.cancel();
```
sethvargo, line 150: `event.PreventDefault()`

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+        flash.error('Canceled batch upload.');
+      });
+
+      $newCode.on('click', function(event) {
+        $retryCode.val(genCode());
```
sethvargo, line 155: `event.PreventDefault()`

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+              }
+
+              let request = {};
+              let cols = rows[i].split(',');
+              request["phone"] = cols[0].trim();
+              request["symptomDate"] = (cols.length > 1) ? cols[1].trim() : "";
+              let hs  = String(CryptoJS.HmacSHA256(request["phone"], retryCode)).substr(0,36)
+              request["uuid"] = hs.substr(0,8)+ '-' + hs.substr(9,4) + '-' + hs.substr(13,4) + '-' + hs.substr(17,4) + '-' + hs.substr(21,12);
+
+              let row = "<tr><td>" + request["phone"] + "</td><td>" + request["symptomDate"] + "</td></tr>";
```
sethvargo, line 206: This is vulnerable to XSS if the CSV contains HTML. Consider building these objects with jQuery and `.html` and `.text` instead.

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+        },
+      });
+    }
+
+    // generates a random 16 digit alphanumeric code
+    function genCode() {
+      return Math.random().toString(36).substr(2, 8) + Math.random().toString(36).substr(2, 8);
+    }
+
+    function setCookie(cname, cvalue, exdays) {
```
sethvargo, line 264: I don't know if this works in all browsers.

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+      var d = new Date();
+      d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
+      var expires = "expires="+d.toUTCString();
```
sethvargo, line 267: I believe cookies are local time, not UTC time

cmd/server/assets/codes/bulkissue.html
```diff
@@ -0,0 +1,289 @@
...
+      document.cookie = cname + "=" + cvalue + ";" + expires;
+    }
+
+    function getCookie(cname) {
```
sethvargo, line 271: We should move cookie functions into `application.js`. We should probably move all of this into application.js

internal/routes/server.go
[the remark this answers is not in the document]
sethvargo: Can we meet in the middle and make bulk-uploading a realm setting that a realm admin has to enable before it can be used?

pkg/controller/issueapi/bulk_issue.go
```diff
@@ -0,0 +1,44 @@
...
+
+package issueapi
+
+import (
+	"net/http"
+
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+)
+
+func (c *Controller) HandleBulkIssue() http.Handler {
```
sethvargo, line 23: Nit: comment

pkg/controller/issueapi/bulk_issue.go
```diff
@@ -0,0 +1,44 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		realm := controller.RealmFromContext(ctx)
+		if realm == nil {
+			controller.MissingRealm(w, r, c.h)
+			return
+		}
+
+		hasSMSConfig, err := realm.HasSMSConfig(c.db)
```
sethvargo, line 33: This check doesn't match what's in the template.

pkg/controller/issueapi/bulk_issue.go
```diff
@@ -0,0 +1,44 @@
...
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
```
sethvargo, line 35: Probably not the right error if we keep this

pkg/controller/issueapi/bulk_issue.go
```diff
@@ -0,0 +1,44 @@
...
+			return
+		}
+
+		m := controller.TemplateMapFromContext(ctx)
+		m["hasSMSConfig"] = hasSMSConfig
+		m.Title("Bulk issue codes")
+		c.h.RenderHTML(w, "code/issue/bulk", m)
```
sethvargo, line 42:
```suggestion
		c.h.RenderHTML(w, "code/issue-bulk", m)
```

pkg/controller/issueapi/bulk_issue.go
whaught: This matches single-code upload?

https://github.com/google/exposure-notifications-verification-server/blob/c6a8d8b9297fa533f9f91300bcfd4064096e1512/pkg/controller/codes/issue.go#L42

sethvargo: The template has checks for if an SMSConfig is defined, but as-written, it would be impossible for that code in the template to execute because this would error out first.

sethvargo: Oops nvm. I'm wrong.

whaught: realm.HasSMSConfig gives true/false (no error if NotFound)

cmd/server/assets/codes/bulkissue.html
whaught: I think UTC is what we want here

cmd/server/assets/codes/bulkissue.html
whaught: moved

cmd/server/assets/codes/bulkissue.html
whaught: Isn't it more the server's responsibility to ensure escaping? Anyone could make a malicious client

whaught: escaped anyway

whaught:
/unhold

I realize the commit list is a little nuts here, but I think this is ready for another review pass. Some of the non-happy-path TODO items I intend to tackle next, but ideally in a new PR.

cmd/server/assets/codes/bulk-issue.html
```diff
@@ -0,0 +1,302 @@
...
+          No SMS provider is configured for this realm.
+          Please contact a realm administrator to enable this feature.
+        </div>
+        {{end}}
+
+        <p>
+          Use this form to bulk issue codes.
+        </p>
+
+        <pre>Example file contents:
```
sethvargo, line 39: Pre is whitespace sensitive. This looks kinda weird:

sethvargo: x

cmd/server/assets/codes/bulk-issue.html
```diff
@@ -0,0 +1,302 @@
...
+          </code>
+        </pre>
+
+        <p>
+          This form will data and issue a code for each provided entry.
+          Phone numbers are expected to be in <a href="https://www.twilio.com/docs/glossary/what-e164" target="_blank">E.164</a> format.
+        </p>
+        <form id="form">
+          <div class="form-label-group input-group">
+            <input type="text" class="form-control" id="retry-code" placeholder="Retry code" required>
```
sethvargo, line 51: Add class `text-monospace`

cmd/server/assets/codes/bulk-issue.html
```diff
@@ -0,0 +1,302 @@
...
+            +12065555555,2020-11-10
+          </code>
+        </pre>
+
+        <p>
+          This form will data and issue a code for each provided entry.
+          Phone numbers are expected to be in <a href="https://www.twilio.com/docs/glossary/what-e164" target="_blank">E.164</a> format.
+        </p>
+        <form id="form">
+          <div class="form-label-group input-group">
```
sethvargo, line 50: Why is this form element so large? It's like 3x higher than our other textboxes.

cmd/server/assets/codes/bulk-issue.html
```diff
@@ -0,0 +1,302 @@
...
+            <input type="text" class="form-control" id="start-at" placeholder="Start at line" value="0">
+            <label for="start-at">Start at line</label>
+          </div>
+          <small class="form-text text-muted pb-4">
+            Parsing the CSV will start at this line. Begin at 0 for a new file.
+            This can be used to resume a canceled or partial upload.
+          </small>
+
+          <div class="custom-file">
```
sethvargo, line 80: This order feels backward. I think I should select the file, then pick options, then start the upload.

cmd/server/assets/codes/bulk-issue.html
```diff
@@ -0,0 +1,302 @@
...
+            <input type="file" class="custom-file-input" id="csv" accept=".csv" disabled required>
+            <label class="custom-file-label" for="csv" id="file-label">Select a CSV file</label>
+          </div>
+          <button class="btn btn-primary" type="submit" id="import" disabled>Issue codes</button>
+          <button class="btn btn-danger" id="cancel" disabled>Cancel</button>
```
sethvargo, line 85: Don't display cancel until upload has started?

cmd/server/assets/codes/issue.html
```diff
@@ -36,6 +36,14 @@
     {{end}}
 
     <h1>Create verification code</h1>
+    {{if $currentRealm.AllowBulkUpload}}
```
sethvargo, line 39: I'd rather have this in the main nav header instead of here.

cmd/server/assets/codes/bulk-issue.html
whaught: I think its the floating labels?

sethvargo: I'll take a look.

PR #1072, opened 2020-11-12
Fill out assertions for Users search test

whaught:
## Proposed Changes

1. Navigate to `New System Admin` page
2. Create an admin user
3. Submit and land on `Index` page
4. Search for newly created user by email
5. Assert row shown
4. Search for newly created user by name with a filter
5. Assert row shown
6. Search for non-existing user
7. Assert no rows shown

**Release Note**

```release-note
Integration test assertions for new system admin
```

### DIFF
--- a/pkg/controller/admin/users_operations.go
+++ b/pkg/controller/admin/users_operations.go
@@ -78,7 +78,7 @@ func (c *Controller) HandleSystemAdminCreate() http.Handler {
 
 			// User does not exist, create a new one.
 			user = &database.User{
-				Name:  form.Email,
+				Name:  form.Name,
 				Email: form.Email,
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/controller/admin/users_operations.go
```diff
@@ -78,7 +78,7 @@ func (c *Controller) HandleSystemAdminCreate() http.Handler {
 
 			// User does not exist, create a new one.
 			user = &database.User{
-				Name:  form.Email,
+				Name:  form.Name,
```
whaught, line 81: This is a heinous bug. Good thing we are testing.

jeremyfaller: yikes.

jeremyfaller (approved):

pkg/controller/admin/users_operations.go
sethvargo: Clicky clicky for-the-win!

PR #1077, opened 2020-11-12
Assert system-admin mobile apps search

whaught:
## Proposed Changes

* Creates a mobile app in the DB
* Asserts search can find it by name
* Asserts search for notexisting gets no results


sethvargo (approved):
/lgtm

PR #1085, opened 2020-11-13
Test system-admin audit event search

whaught:
## Proposed Changes

* Creates an audit event with a known time
* Searches for events in that time frame and asserts the one
* Searches and hour before and asserts no results

**Release Note**

```release-note
Integration test for system-admin events page
```


mikehelmick:
see test failures

mikehelmick (approved):

PR #1086, opened 2020-11-13
Accept client UUID for issued-code tracking

whaught:

## Proposed Changes

* Allow client to provide the tracking UUID for an issued code
    * Thus will allow us to pre-allocate these for the bulk-upload scenario which will help with tracking / retryability
* Backwards-compatible, if not provided the server will generate the UUID.
* Redo the UUID index as UNIQUE for enforcement

**Release Note**

```release-note
Allow client-provided UUID for issuing a code
Drop and re-create UUID index as UNIQUE
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -62,6 +62,8 @@ const (
 	ErrTokenExpired = "token_expired"
 	// ErrHMACInvalid indicates that the HMAC that is being signed is invalid (wrong length)
 	ErrHMACInvalid = "hmac_invalid"
+	// ErrUUIDAlreadyExists indicates that the UUID has already been used for an issued code.
+	ErrUUIDAlreadyExists = "uuid_already_exists"
 )
 
 // ErrorReturn defines the common error type.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -175,6 +175,10 @@ type IssueCodeRequest struct {
 	// (using the default of 0) are all valid. 0 is considered to be UTC.
 	TZOffset float32 `json:"tzOffset"`
 	Phone    string  `json:"phone"`
+
+	// Optional: UUID is a handle which allows the issuer to track status
+	// of the issued verification code. If omitted the server will generate the UUID.
+	UUID string `json:"uuid"`
 }
 
 // IssueCodeResponse defines the response type for IssueCodeRequest.
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -192,7 +192,7 @@ func (c *Controller) HandleIssue() http.Handler {
 			return
 		}
 
-		// Set up parallel arrays to leverage the observability reporting and connect the parse / valdiation errors
+		// Set up parallel arrays to leverage the observability reporting and connect the parse / validation errors
 		// to the correct date.
 		parsedDates := make([]*time.Time, 2)
 		input := []string{request.SymptomDate, request.TestDate}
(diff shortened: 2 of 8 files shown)


pkg/otp/code.go
```diff
@@ -133,10 +135,14 @@ func (o *Request) Issue(ctx context.Context, retryCount uint) (string, string, s
 			LongExpiresAt: o.LongExpiresAt,
 			IssuingUserID: issuingUserID,
 			IssuingAppID:  issuingAppID,
+			UUID:          o.UUID,
 		}
 		// If a verification code already exists, it will fail to save, and we retry.
-		if err := o.DB.SaveVerificationCode(&verificationCode, o.MaxSymptomAge); err != nil {
```
whaught, line 138: I think this is a bug. This `err` is scoped to this if statement, but when we break we `return "", "", "", err` which is unset in that scope

pkg/database/migrations.go
```diff
@@ -1639,6 +1638,35 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00066-AddVerCodeUUIDUniqueIndexe",
+			Migrate: func(tx *gorm.DB) error {
+				sqls := []string{
+					`DROP INDEX IF EXISTS idx_vercode_uuid`,
```
mikehelmick, line 1645: this could be expensive - should be noted in release notes

pkg/controller/issueapi/issue.go
```diff
@@ -290,14 +290,21 @@ func (c *Controller) HandleIssue() http.Handler {
...
 
 		code, longCode, uuid, err := codeRequest.Issue(ctx, c.config.GetCollisionRetryCount())
 		if err != nil {
 			logger.Errorw("failed to issue code", "error", err)
-			c.h.RenderJSON(w, http.StatusInternalServerError, api.Errorf("failed to generate otp code, please try again"))
 			blame = observability.BlameServer
 			result = observability.ResultError("FAILED_TO_ISSUE_CODE")
+
+			// GormV1 doesn't have a good way to match db errors
+			if strings.Contains(err.Error(), database.VercodeUUIDUniqueIndex) {
```
mikehelmick, line 303: I think you want to handle this error up one level in `otp.Request.Issue`

otherwise we'll retry the code creation multiple times before popping back up here

and that particular flavor of error is probably not BlamServer?

mikehelmick (changes requested):

pkg/controller/issueapi/issue.go
whaught: This is also handled in otp.Request.Issue, this is just translating it to be 409 conflict

pkg/database/migrations.go
whaught: Added to the notes

pkg/controller/issueapi/issue.go
mikehelmick: missed that - was looking at the code at head

pkg/api/api.go
```diff
@@ -175,6 +175,10 @@ type IssueCodeRequest struct {
 	// (using the default of 0) are all valid. 0 is considered to be UTC.
 	TZOffset float32 `json:"tzOffset"`
 	Phone    string  `json:"phone"`
+
+	// Optional: UUID is a handle which allows the issuer to track status
+	// of the issued verification code. If omitted the server will generate the UUID.
+	UUID string `json:"uuid"`
```
mikehelmick, line 181: can you update the API docs as well - trying to keep those in sync

mikehelmick: including the new return code!

mikehelmick (commented):
/lgtm
/hold

you can decide if you want to do api doc in a followup

mikehelmick (approved):

whaught:
> /lgtm
> /hold
> 
> you can decide if you want to do api doc in a followup

Updated with the doc.
/unhold

mikehelmick (approved):

PR #1088, opened 2020-11-13
Adding service definition and SLO definition

yuriatgoogle:
Fixes #

## Proposed Changes

* added service definition in slos.tf
* added availability SLO in slos.tf
*

**Release Note**

```release-note
Added service definition and availability SLO definition
```


terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,49 @@
...
+  # the basics
+  service = google_monitoring_custom_service.verification-server.service_id
+  slo_id = "availability-slo"
+  display_name = "99.9% of requests are successful over rolling 28 days"
+  project = var.monitoring-host-project
+
+  # the SLI
+  request_based_sli {
+    good_total_ratio {
+      good_service_filter = join(" AND ", [
```
yegle, line 31: Use multiline string literal, there's no need to join here:

good_service_filter = <--EOT
metric.type="loadbalancing.googleapis.com/https/request_count" AND
...
EOT

terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,49 @@
...
+        "metric.type=\"loadbalancing.googleapis.com/https/request_count\"",
```
yegle, line 32: It looks like this depends on GCLB? If so please optionally enable it based on whether var.https-forwarding-rule is defined or not. See `alerts.tf` for reference.

terraform/alerting/verification-server/alerts.tf
```diff
@@ -326,4 +326,4 @@ EOT
     null_resource.manual-step-to-enable-workspace,
     google_logging_metric.stackdriver_export_error_count
   ]
-}
+}
```
yegle, line 329: Revert this change.

terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,49 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "google_monitoring_custom_service" "verification-server" {
+  service_id = "verification-server"
+  display_name = "Verification Server"
+  project = var.monitoring-host-project
+}
+
+resource "google_monitoring_slo" "availability-slo" {
```
yegle, line 21: Will there be a notification if we are out of SLO?

yuriatgoogle: Yes, I'll add an alert in a subsequent PR.

terraform/alerting/verification-server/slos.tf
yuriatgoogle: Done.

terraform/alerting/verification-server/slos.tf
yuriatgoogle: Done - I think?

terraform/alerting/verification-server/alerts.tf
yuriatgoogle: Done, I think?

yuriatgoogle:
I've tried to address all comments.

terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,53 @@
...
+  # the basics
+  service = google_monitoring_custom_service.verification-server.service_id
+  slo_id = "availability-slo"
+  display_name = "99.9% of requests are successful over rolling 28 days"
+  project = var.monitoring-host-project
+  # create if GCLB is used
+  depends_on = [
+    var.https-forwarding-rule
```
yegle, line 29: Use count to optionally enable this resource:

```
count = var.https-forwarding-rule == "" ? 0 : 1
```

terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,53 @@
...
+        metric.type="loadbalancing.googleapis.com/https/request_count"
+        resource.type="https_lb_rule"
+        resource.label.backend_name="apiserver"
+        metric.label.response_code_class=200
+      EOT
+      bad_service_filter = <<-EOT
+        metric.type="loadbalancing.googleapis.com/https/request_count"
+        resource.type="https_lb_rule"
+        resource.label.backend_name="apiserver"
+        metric.label.response_code_class=500
```
yegle, line 45: What if we accidentally rejected all requests with 429 status code?

I think there's some nuance here if we only look at 200 and 500 code. I'm fine with only check 200/500 but am a bit concern that it doesn't give us the whole picture.

terraform/alerting/verification-server/slos.tf
```diff
@@ -0,0 +1,53 @@
...
+      EOT
+    }
+  }
+
+  # the goal
+  goal = 0.999
```
yegle, line 51: Let's start with 0.995 first. Most of our dependencies have SLA of 0.995 (Cloud Run, GCLB etc.)

yegle (approved):

yegle:
Oops I didn't realize approval will auto merge. Please address the depends_on in a follow up PR.

terraform/alerting/verification-server/slos.tf
yuriatgoogle: Done - I think?

terraform/alerting/verification-server/slos.tf
yuriatgoogle: Done.

terraform/alerting/verification-server/slos.tf
yuriatgoogle: I agree that this is an imperfect place to start. Let's see what alerts look like once this is in place.

PR #1096, opened 2020-11-16
Cleanup codes/ paths

sethvargo:
Also a few nits:

- Fixes the HTML nesting on code lookup form
- Redirects a good UUID in the wrong format/casing to the correct URL

**Release Note**

```release-note
NONE
```

The /home -> /code was never deployed, so we don't need to do a /code -> /codes redirect.


whaught (approved):
Looks good - I realized this later that 'codes' should be plural. If you're up for it can we also rename the folder "codestatus" to "codes"
/hold

whaught (approved):

whaught (approved):

PR #1097, opened 2020-11-16
Move /codes html folder and template names

whaught:
## Proposed Changes

* Make the template names consistent with their serving paths
* Move /codestatus html asset folder to /codes


sethvargo (approved):
/lgtm

PR #1098, opened 2020-11-16
Skip twilio for fake phones in dev mode

whaught:
## Proposed Changes

* Expose DevMode to IssueConfig
* Detect phones in the [fictional range](https://en.wikipedia.org/wiki/555_(telephone_number)) and just log when in dev mode

I'm trying to make a fake CSV to test some bulk-issue scenarios.

**Release Note**

```release-note
Allow no-op for SMS from fictional +1XXX55501XX numbers in dev mode for testing
```

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -310,7 +310,10 @@ func (c *Controller) HandleIssue() http.Handler {
 		}
 
 		if request.Phone != "" && smsProvider != nil {
-			if err := func() error {
+			// +1XXX55501XX is reserved for fictitious use. Just log for dev mode.
+			if c.config.DevMode() && request.Phone[:2] == "+1" && request.Phone[5:5] == "55501" {
+				logger.Infow("dev mode example number detected", "phone", request.Phone)
+			} else if err := func() error {
 				defer observability.RecordLatency(ctx, time.Now(), mSMSLatencyMs, &blame, &result)
 				message := realm.BuildSMSText(code, longCode, c.config.GetENXRedirectDomain())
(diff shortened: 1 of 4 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -310,7 +310,10 @@ func (c *Controller) HandleIssue() http.Handler {
 		}
 
 		if request.Phone != "" && smsProvider != nil {
-			if err := func() error {
+			// +1XXX55501XX is reserved for fictitious use. Just log for dev mode.
+			if c.config.DevMode() && request.Phone[:2] == "+1" && request.Phone[5:5] == "55501" {
```
mikehelmick, line 314: Is there a panic if I put in an invalid phone number

`+1123`

mikehelmick: I would move this to a helper / testable method

whaught: good points. hows that?

sethvargo:
I'm not a huge fan of magical numbers beyond what Twilio already provides. Can you use the Twilio magic numbers instead: https://www.twilio.com/blog/2018/04/twilio-test-credentials-magic-numbers.html

whaught:
> I'm not a huge fan of magical numbers beyond what Twilio already provides. Can you use the Twilio magic numbers instead: https://www.twilio.com/blog/2018/04/twilio-test-credentials-magic-numbers.html

I'll use those for testing. My only counterpoint here is that this is sms-provider agnostic with the US reserved magic number (however twilio is our only supported provider, so this works for me).

/close

PR #1107, opened 2020-11-17
Add i18n framework

### LINKED ISSUE #1083: i18n
### TL;DR

Prepare key strings in the issue code UI (not the admin UI) for internationalization.

No translation or language settings just yet.

/kind enhancement
/assign @sethvargo

sethvargo:
Adds a few default translations for demonstration

Fixes #1083

**Release Note**

```release-note
Add internationalization framework (no supported translations yet)
```


cmd/server/assets/codes/issue.html
```diff
@@ -35,7 +35,7 @@
       </div>
     {{end}}
 
-    <h1>Create verification code</h1>
+    <h1>{{t $.locale "create-verification-code"}}</h1>
```
sethvargo, line 38: I'm not in love with having to pass in the locale, but template functions aren't mutable after creation.

pkg/render/render.go
```diff
@@ -151,6 +152,21 @@ func selectedIf(v bool) htmltemplate.HTML {
 	return ""
 }
 
+// translate accepts a message printer (populated by middleware) and prints the
+// translated text for the given key. If the printer is nil, an error is
+// returned.
+func translate(l *gotext.Locale, key string, vars ...interface{}) (string, error) {
```
sethvargo, line 158: This is important because the default behavior is to just return the empty string, which is a pretty difficult-to-debug failure. I'd like the template execution to fail on missing i18n keys.

internal/i18n/locales/en/default.po
```diff
@@ -0,0 +1,16 @@
+msgid ""
+msgstr ""
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+msgid "create-verification-code"
```
sethvargo, line 9: I'd like to come up with a better grouping for these. I thought about `header.create-verification-code` or something to make it clearer.

internal/i18n/locales/es/default.po
```diff
@@ -0,0 +1,16 @@
...
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+msgid "create-verification-code"
+msgstr "Crear código de verificación"
+
+msgid "diagnosis"
+msgstr "Diagnóstico"
+
+msgid "dates"
+msgstr "Fechas"
```
whaught, line 16: nice spanish

internal/i18n/locales/en/default.po
```diff
@@ -0,0 +1,16 @@
+msgid ""
+msgstr ""
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
```
whaught, line 7: what is this Plural-Forms doing?

sethvargo: Nothing right now. But if you ever had a sentence like:

> There were %d bananas

You'd want to define:

- `=0` "There were no bananas"
- `=1` "There was 1 banana"
- `>1` "There were %d bananas"

That's what plurals do. It's part of the gettext standard: https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

PR #1109, opened 2020-11-17
More bulk upload js

whaught:
## Proposed Changes

* Fix issue with the total count
* Fix the flashing progress bar
* The results table is still a little ugly and possibly confusing - I intend to keep improving error handling etc


cmd/server/assets/codes/bulk-issue.html
```diff
@@ -114,6 +114,8 @@
   </main>
 
   <script type="text/javascript">
+    var total = 0;
```
sethvargo, line 117: Nit: `let`

sethvargo (approved):
/lgtm

PR #1110, opened 2020-11-17
Reverse dates for user stats

whaught:
## Proposed Changes

* These dates show up backwards - reverse the data array to match what we do with realm stats

### DIFF
--- a/cmd/server/assets/users/show.html
+++ b/cmd/server/assets/users/show.html
@@ -93,7 +93,6 @@
 
     function drawChart() {
       let arr = [
-        ['Date', 'Codes issued'],
         {{range $stat := $stats}}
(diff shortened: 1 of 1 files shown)


cmd/server/assets/users/show.html
```diff
@@ -92,20 +92,24 @@ <h6 class="card-title">Realm admin</h6>
     google.charts.setOnLoadCallback(drawChart)
 
     function drawChart() {
-      var data = google.visualization.arrayToDataTable([
+      let arr = [
         ['Date', 'Codes issued'],
```
jeremyfaller: You need to delete this line

whaught: refresh!

jeremyfaller (approved):

sethvargo (approved):
/lgtm

PR #1111, opened 2020-11-17
Sort stats dates ascending on the server

whaught:
## Proposed Changes

* We currently javascript-reverse the stats to render charts properly - this instead sorts ascending from the server


sethvargo (changes requested):
/hold

This will super break the modeler: https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/controller/modeler/modeler.go#L127

whaught:
/close

table and modeler do want DESC, so we have to reverse something

PR #1114, opened 2020-11-17
Switch var to let for function-scoped javascript

whaught:
## Proposed Changes

* Should be functionally identical, but scopes down function-level js


sethvargo (approved):
/lgtm

mikehelmick (approved):

PR #1116, opened 2020-11-17
All / system events filter for system admin

whaught:
## Proposed Changes

* Trying to pull out the uncontroversial part of https://github.com/google/exposure-notifications-verification-server/pull/1040
* Adds a filter for all events or system-only events to system-admin

filter

**Release Note**

```release-note
Filter system-events in system-admin event page
```


whaught:
/close

I think we're going to go with the original one. yay

PR #1118, opened 2020-11-17
Remove RealmTokenRemainingCapacityLow alert

yegle:
This alert was initially added so that we can notify realm admin about
impending quota exhausion. We now have all the tools for the realm admin
to monitor and add burst quota, and the system admin (who holds the
pager) should no longer be involved to handle this situation.


sethvargo:
/lgtm

Tests are failing though

PR #1125, opened 2020-11-17
Stop bulk upload if we see certain permanent errors or throttling

whaught:

## Proposed Changes

* Stop at 429 Too Many Requests which we give when the user runs out of codes
    * Stop a few permanent errors for good measure

**Release Note**

```release-note
Cancel remaining bulk upload when throttled by the server
```


sethvargo (approved):
/lgtm

PR #1128, opened 2020-11-18
Throttle bulk code issuance using server ratelimiter config

whaught:

## Proposed Changes

* Look for the retry-after header on error response
* if found, sleep as directed

**Release Note**

```release-note
Throttle batches for bulk code issuance
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -336,7 +336,7 @@ You can expect the following responses from this API:
 
 -   `412` - The client requested a precondition that cannot be satisfied.
 
--   `429` - The client is rate limited. Check the `X-Retry-After` header to
+-   `429` - The client is rate limited. Check the `Retry-After` header to
     determine when to retry the request. Clients can also monitor the
     `X-RateLimit-Remaining` header that's returned with all responses to
     determine their rate limit and rate limit expiration.
--- a/cmd/server/assets/codes/bulk-issue.html
+++ b/cmd/server/assets/codes/bulk-issue.html
@@ -115,9 +115,6 @@
 
   <script type="text/javascript">
     let total = 0;
-    let throttle = {{.throttleMs}}; // milliseconds per request.
-
-    // Common error codes which should cancel the whole upload.
     let stopUploadingCodes = [
       '429', // too many requests
       '403', // forbidden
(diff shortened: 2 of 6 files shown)


sethvargo:
> Alternative: instead of spreading load evenly we could burst up to remaining tokens and then sleep until the interval. Would make things faster for small CSVs that are under the count/interval. That's a little trickier because we'd actually need to peek into the ratelimiter (not just the config)

I would prefer this approach, but you don't need to peek into the limiter, look at the response headers. They'll include all the data you need 👍

whaught:
> > Alternative: instead of spreading load evenly we could burst up to remaining tokens and then sleep until the interval. Would make things faster for small CSVs that are under the count/interval. That's a little trickier because we'd actually need to peek into the ratelimiter (not just the config)
> 
> I would prefer this approach, but you don't need to peek into the limiter, look at the response headers. They'll include all the data you need 👍

I just saw that we have `X-Retry-After` in the api doc, which is what I need for this

mikehelmick:
> > > Alternative: instead of spreading load evenly we could burst up to remaining tokens and then sleep until the interval. Would make things faster for small CSVs that are under the count/interval. That's a little trickier because we'd actually need to peek into the ratelimiter (not just the config)
> > 
> > 
> > I would prefer this approach, but you don't need to peek into the limiter, look at the response headers. They'll include all the data you need 👍
> 
> I just saw that we have `X-Retry-After` in the api doc, which is what I need for this

one consideration with the consuming all of the realm's quota - it makes it so that online code distribution won't work during this. probably not a big deal

docs/api.md
```diff
@@ -336,7 +336,7 @@ You can expect the following responses from this API:
 
 -   `412` - The client requested a precondition that cannot be satisfied.
 
--   `429` - The client is rate limited. Check the `X-Retry-After` header to
```
whaught, line 339: There is no `x-` for retry-after
https://github.com/sethvargo/go-limiter/blob/86d882305d36ae8c59138bc9a93b3599f0b70936/httplimit/middleware.go#L28

sethvargo: Why would I remember my own code? 😝

sethvargo (approved):
/lgtm

PR #1129, opened 2020-11-18
Keep fake success on no-user

whaught:
## Proposed Changes

* For security reasons we don't want to differentiate between user not-found and when we really send and email. A malicious user could use this to trawl account-existence.
* I think we accidentally "fixed" this in https://github.com/google/exposure-notifications-verification-server/pull/902


jeremyfaller (approved):

sethvargo:
FWIW, I adamantly disagree with this "security" practice, but fiiiine 😄

PR #1130, opened 2020-11-18
More resilient recaptcha

whaught:
## Proposed Changes

* Load recaptcha in jquery
* Use the recaptcha callback submit to ensure captcha solved on login
* Reset recaptcha on unhandled failures
* Fixes an 'error' misnamed 'err'
* Capitalization of SMS pin heading


sethvargo (approved):
/lgtm

Issue #1131, opened 2020-11-18
Cannot enable MFA without being a member of at least one realm


sethvargo:
### TL;DR
As a system admin, I want to enable MFA. If I'm not a member of a realm, attempting to register a phone number results in "Please select a realm to continue."

![screenshot-20201118-104548](https://user-images.githubusercontent.com/408570/99552725-564cd180-298b-11eb-8914-b98ee3193300.gif)

PR #1133, opened 2020-11-18
Don't require realm for sms registration

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1131

## Proposed Changes

* Guard places where we display realm in SMS registration and show system defaults

### DIFF
--- a/cmd/server/assets/login/register-phone.html
+++ b/cmd/server/assets/login/register-phone.html
@@ -106,9 +106,9 @@
       let phone = document.querySelector('#phone');
       let iti = window.intlTelInput(phone, {
         nationalMode: true,
-        {{-if .currentRealm}}{{-if $currentRealm.SMSCountry}}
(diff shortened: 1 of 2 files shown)


cmd/server/assets/login/register-phone.html
```diff
@@ -108,9 +106,9 @@
       let phone = document.querySelector('#phone');
       let iti = window.intlTelInput(phone, {
         nationalMode: true,
-        {{- if $currentRealm.SMSCountry }}
-        initialCountry: '{{$currentRealm.SMSCountry}}',
-        {{- end }}
+        {{-if .currentRealm}}{{-if $currentRealm.SMSCountry}}
```
sethvargo, line 109: `$currentRealm` is no longer defined

sethvargo (approved):
/lgtm

Issue #606, opened 2020-09-21
Delete User if removed from all realms


whaught:
### TL;DR
Currently deleting a user removes the current realm from their User entry

**Proposal**
If a user has no realms and has not been modified for ~30 days (configurable), the system should delete their database entry.

The cleanup job should **not** do this in real-time: when the last realm is removed, an admin may be adding them to another one.

**Alternatives considered**
Instead of deleting we could remove/scramble User PII data. We would do this if the record is important for maintaining any sort of auditability.

sethvargo:
I think we should talk about this more in standup/chat. I'm worried about auditing requirements and deleting users.

whaught:
Sure thing - I think it's worth discussing one way or the other.

mikehelmick:
re: auditing - are you suggesting we just keep the user records around longer?

the user-stats is the only thing we would possibly lose, but I don't think those are accessible once the user has been removed from the realm already.

sethvargo:
I'm thinking about the requirements we got from certain states.

mikehelmick:
Nothing actionable there yet.

sethvargo:
Do we still want to do this? I brought this up to @whaught in chat, but I'm worried about a situation where someone is using a shared Firebase account (e.g. shared among multiple login systems) and they want to disable that person's access to ENVS. It feels weird that ENVS would disable their overall account in that parent authentication system.

If we integrated directly with SAML, for example, I wouldnt want ENVS to delete the user from my internal IDP.

whaught:
I agree with your point about the auth entry. I hadn't considered sharing auth with this system and others. That said we still might want to delete the ENVS User entry after a time.

PR #1135, opened 2020-11-18
Purge old realmless users

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/606

## Proposed Changes

* Delete users who have been removed from all realms, are not a system admin, and are older than USER_PURGE_MAX_AGE

**Release Note**

```release-note
Cleanup for old users who have no realms and have aged out. This clears their DB information, but not their auth.
```


whaught:
I tried to do this as

```
Joins("LEFT JOIN user_realms ON users.id = user_realms.user_id").
Where("AND user_realms.user_id IS NULL").
```

but that didn't seem to work

pkg/database/user.go
```diff
@@ -330,6 +330,21 @@ func (db *Database) DeleteUser(u *User, actor Auditable) error {
...
+// PurgeUsers will delete users who are not a system admin, not a member of any realms
+// and have not been modified before the expiry time.
+func (db *Database) PurgeUsers(maxAge time.Duration) (int64, error) {
+	if maxAge > 0 {
+		maxAge = -1 * maxAge
+	}
+	deleteBefore := time.Now().UTC().Add(maxAge)
+	// Delete users who were created/updated before the expiry time.
+	rtn := db.db.Unscoped().
+		Where("users.system_admin = false AND users.created_at < ? AND users.updated_at < ?", deleteBefore, deleteBefore).
```
sethvargo, line 342: I don't think this does what you think it does? 

- Realm membership changes don't change `updated_at`, so you could end up immediately deleting someone
- That column is updated anytime the direct user record changes, we can happen at login with the revoke check

sethvargo (commented):
/hold

pkg/database/user.go
whaught: The latter isn't a problem - it's okay to not-delete users who are logging in to keep fresh.

I suppose we can ensure users is bumped on realm-remove

sethvargo: > I suppose we can ensure users is bumped on realm-remove

I couldn't figure out a good way to do that. If we can "touch" the update_at on the user when realm membership is changed, I think this is fine (maybe add a comment).

whaught: I've added `TestRemoveRealmUpdatesTime` and I can't seem to repro your claim that modifying the realm relationship _isn't_ bumping time

pkg/database/user_test.go
```diff
@@ -97,6 +97,115 @@ func TestUserLifecycle(t *testing.T) {
...
+		Name:        "Dr Delete",
+		SystemAdmin: true,
+	}
+
+	if err := db.SaveUser(&user, System); err != nil {
+		t.Fatalf("error creating user: %v", err)
+	}
+	expectExists(t, db, user.ID)
+
+	db.PurgeUsers(time.Duration(0)) // is admin
```
sethvargo, line 117: Check error

pkg/database/user.go
sethvargo: Interesting... what happens if the user comes serialized from the cache?

mikehelmick:
not sure if we can use it - but firebase stores the last login date of a user.

pkg/database/user.go
whaught: I believe the reason is that we have wrapped calls with SaveUser here:
https://github.com/google/exposure-notifications-verification-server/blob/f76a3cea09d53d076a47a204ce9c948e183c8f35/pkg/database/user.go#L337

which always saves both the User entry and the associations (even if the table fields haven't changed)

sethvargo (approved):
/lgtm
/hold 

Can you update the release note to warn about this and note that it does _not_ delete them from the upstream auth provider?

sethvargo:
Hmm - you might need to rebase @whaught

PR #1136, opened 2020-11-18
Spanish feedback

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/internal/i18n/locales/es/default.po
+++ b/internal/i18n/locales/es/default.po
@@ -26,7 +26,7 @@ msgid "nav.bulk-issue-codes"
 msgstr "Crear códigos en lote"
 
 msgid "nav.check-code-status"
-msgstr "Verificar estado código"
+msgstr "Estado de código"
 
 msgid "nav.api-keys"
 msgstr "Claves API"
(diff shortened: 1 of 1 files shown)


internal/i18n/locales/es/default.po
```diff
@@ -38,7 +38,7 @@ msgid "nav.event-log"
 msgstr "Registro de eventos"
 
 msgid "nav.signing-keys"
-msgstr "Firma de claves"
+msgstr "Clave de firma"
```
rafaeltello, line 41: Sorry, I should have said "Claves" in the feedback (plural).

rafaeltello: Also, I found a couple of sources that use "Llaves de firmado". Let me check with a couple of industry friends from Mexico to get a sense of which one is more common (both are correct).

rafaeltello: Ok, after checking with a couple of Mexican security folks, they tell me "Llaves para firmar" is actually the most widely used term (like in https://www.php.net/manual/es/book.gnupg.php ). Part of the problem is that in Mexico, "clave" (key/code/clef) is often used interchangeably with "contraseña" (password).
That being said, "Claves para firmar" will be understood as well though.

sethvargo: Awesome - thanks for the background. I imagine we'll enlist you in the future :)

whaught (approved):

Issue #1139, opened 2020-11-18
Add maintenance mode


sethvargo:
System setting only available to system admins. If set, add a banner to all pages. Can be used to annotate an outage or upcoming maintenance across all realms.

- [ ] Database setting
- [ ] HTML banner
- [ ] Return 429s in `InternalError` handler when maint is active (maybe just for issueapi)

sethvargo:
Maybe envvar instead of database, but db might be better to sync across the 3 services...

PR #1140, opened 2020-11-18
Add system admin message text and maintenance bit

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1139

## Proposed Changes

* Adding DB fields for a system message and maintenance bit

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1674,7 +1674,9 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				logger.Debugw("db migrations: adding allow_bulk_upload to realms")
 
 				sqls := []string{
-					`ALTER TABLE realms ADD COLUMN IF NOT EXISTS allow_bulk_upload bool DEFAULT false`,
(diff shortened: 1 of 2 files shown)


sethvargo (changes requested):
This needs to be at a higher level than the realm, since this applies to all realms. I think we need to do an envvar or new table.

whaught:
> This needs to be at a higher level than the realm, since this applies to all realms. I think we need to do an envvar or new table.

I was going to write it to all realms. A new table seems like a lot. This had the advantage of being recyclable for generic messages the system might need to give out in the future (thus the separate bool).

That said if this is just a quick thing for the postgres upgrade, envvar is way easier to write

sethvargo:
The problem with scoping it to the realm is that now a realm needs to be in scope during any operation to check if maintenance mode is active. I think envvar is probably fine for this.

whaught:
> The problem with scoping it to the realm is that now a realm needs to be in scope during any operation to check if maintenance mode is active. I think envvar is probably fine for this.

See https://github.com/google/exposure-notifications-verification-server/pull/1142

PR #1141, opened 2020-11-18
Switch how we make this bool column

whaught:
## Proposed Changes

I'm not sure how postgres creates a column with default set, but our other columns are initialized like this


sethvargo (approved):
/lgtm

PR #1142, opened 2020-11-18
Maintenance mode env var

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1139

## Proposed Changes

* Creates an environment variable for maintenance mode which rejects api/issue requests
* random nits

**Release Note**

```release-note
Environment variable for maintenance mode
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -100,6 +100,8 @@ Possible error code responses. New error codes may be added in future releases.
 | `code_not_found`        | 400         | No    | The server has no record of that code. |
 | `invalid_test_type`     | 400         | No    | The client sent an accept of an unrecgonized test type |
 | `missing_date`          | 400         | No    | The realm requires either a test or symptom date, but none was provided. |
+| `uuid_already_exists`   | 409         | No    | The UUID has already been used for an issued code |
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -152,7 +154,7 @@ Possible error code responses. New error codes may be added in future releases.
 | `token_invalid`         | 400         | No    | The provided token is invalid, or already used to generate a certificate |
 | `token_expired`         | 400         | No    | Code invalid or used, user may need to obtain a new code. |
 | `hmac_invalid`          | 400         | No    | The `ekeyhmac` field, when base64 decoded is not the right size (32 bytes) |
-| `uuid_already_exists`   | 409         | No    | The UUID has already been used for an issued code |
```
sethvargo, line 155: Did you mean to delete this?

sethvargo (approved):
/lgtm
/hold

It might be good to make it a string so we can customize the message, but a bool is probably fine.

docs/api.md
whaught: it's in the wrong section, I moved it up. That's a response for Issue

whaught:
/hold cancel
I'm going to leave bit for now, we can customize a message later if we need to

PR #1146, opened 2020-11-19
Update register phone catcha

whaught:
## Proposed Changes

* Fires the recaptcha on form submit
    * This makes the 2 minute timeout less likely to expire if you load the page and sit there
    * Already did this for the login page
* Reset the captcha on failure / cancel
* Reset the captcha on successful enrollment
    * Gives a new captcha token if you're enrolling more than one phone
* Store MFA prompted when we decide it's not needed for the realm
* Clear MFA prompted on realm select (if its a different realm)

### DIFF
--- a/cmd/server/assets/login/register-phone.html
+++ b/cmd/server/assets/login/register-phone.html
@@ -63,25 +63,12 @@
           </div>
 
           {{template "login/pindiv" .}}
-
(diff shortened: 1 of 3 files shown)


sethvargo (approved):
/lgtm

PR #1147, opened 2020-11-19
Share js retry logic

whaught:
## Proposed Changes

* Adds retry logic to user-import and moves the js to application.js

### DIFF
--- a/cmd/server/assets/static/js/application.js
+++ b/cmd/server/assets/static/js/application.js
@@ -253,7 +253,6 @@ $(function() {
   window.flash = flash;
 });
 
-
 function setCookie(cname, cvalue, exdays) {
   let d = new Date();
   d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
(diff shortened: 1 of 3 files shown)


jeremyfaller (approved):

cmd/server/assets/static/js/application.js
```diff
@@ -275,3 +274,38 @@ function getCookie(cname) {
   }
   return "";
 }
+
+// Common error codes which should cancel the whole upload.
+let stopUploadingCodes = [
```
sethvargo, line 279: Nit: `const`

sethvargo, line 279: It'd be a larger refactor, but should we get these from the server somehow?

whaught: That's a good idea. I also want to include the detailed error const eg "maintenance_mode" rather than just HTTP code. I'll save that all for another PR and do them together

jeremyfaller (approved):

PR #1149, opened 2020-11-19
Have the bulk uploader pay attention to our enum values, version js

whaught:
## Proposed Changes

* Check the error enum codes too
* Append the server build string to our `application.js?{{.buildID}}` to cause clients to not use outdated cached versions.


sethvargo (approved):
/lgtm

Issue #1153, opened 2020-11-20
Exclude e2e from audit logs


sethvargo:
The system event log is quite littered with actions by the e2e-runner, making it _incredibly_ difficult to discern actual behavior for the data. Should we exclude the e2e-runner from events/audits? We can either prevent creating them or prevent showing them.

@whaught @chaodaiG thoughts?

<img width="743" alt="screenshot-20201120-100808@2x" src="https://user-images.githubusercontent.com/408570/99815433-4c0c0e00-2b18-11eb-96eb-dc486570b218.png">

chaodaiG:
Agree that these should not be shown, hiding them is preferable to me

whaught:
I don't want to skip logging per-se (I don't want there to exist a way to bypass), but I think the ability to mark and AppId as 'test' and have a static Auditable SystemTest instead of just System. Then we can implement a filter in the logs to not show these by default with a checkbox to show them.

Thoughts?

sethvargo:
I think that's fine, but I'm curious why add `IsTest` to auditable? We can just filter audits where the principle is SystemTest (which feels cleaner)

whaught:
Because we can use SystemTest for the things where System was the auditable, but parts of the the test eg. https://github.com/google/exposure-notifications-verification-server/blob/d2c677772dd08b06c1f7e61506370cae2218c28c/pkg/clients/e2e.go#L93
Also are taking actions that will go to the realm log. This allows us to filter any kind of test. If PWC wants to build external tests in the future, this all makes it simple for anything to be marked as test.

sethvargo:
My pref would be to just use SystemTest for now. We could also build a system-admin-only portal to generate API keys that are tagged as SystemTest that PWC could use for their testing.

PR #1155, opened 2020-11-20
Filter test-data from event logs

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1153

## Proposed Changes

* Adds a bit to authorized_app and audit_entry to mark data as test data
* Creates a SystemTest audit actor
* Updates e2e test to use the above
* No support for test User, MobileApp, or Realm (we can do that later if needed).
* Adds a checkbox to event log in system admin to include test events (off by default)

**Release Note**

```release-note
Mark e2e test audit events as 'test' and filter them out by default
```

### DIFF
--- a/cmd/server/assets/admin/events/index.html
+++ b/cmd/server/assets/admin/events/index.html
@@ -36,6 +36,7 @@
         <form method="GET" action="/admin/events" id="search-form">
           <!--
             hidden query to filter on test events values are:
+               "?include_test=false" // Filters out test events. Default if omitted.
                "?include_test=only" // shows only test events
                "?include_test=all" // shows all test and non-test events
            -->
--- a/cmd/server/assets/admin/events/index.html
+++ b/cmd/server/assets/admin/events/index.html
@@ -60,7 +60,7 @@
           </div>
           <div class="form-check">
             <input type="checkbox" class="form-check-input" name="include_test" id="include-test" value="true" {{if .include_test}}{{if eq .include_test "true"}}checked{{end}}{{end}}>
-            <label class="form-check-label" for="includ-test">Include test</label>
+            <label class="form-check-label" for="include-test">Include test</label>
           </div>
         </form>
       </div>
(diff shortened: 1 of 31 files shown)


pkg/database/authorized_app.go
```diff
@@ -81,6 +81,10 @@ type AuthorizedApp struct {
 
 	// APIKeyType is the API key type.
 	APIKeyType APIKeyType `gorm:"column:api_key_type; type:integer; not null;"`
+
+	// TestApp marks this as a test app which may be used to ignore or filter
+	// data (such as audit logs) from this class of app.
+	TestApp bool `gorm:"type:boolean; default: false"`
```
sethvargo, line 87: Nit trailing `;`

sethvargo (commented):
This is definitely verbose. TBH, my original plan was to just change the scope query to exclude things that started with `e2e-*`

cmd/server/assets/admin/events/index.html
```diff
@@ -58,6 +58,10 @@ <h1>System event log</h1>
               </button>
             </div>
           </div>
+          <div class="form-check">
```
sethvargo, line 61: I think this should be a hidden URL param - no need to edit the form.

pkg/database/scopes.go
```diff
@@ -76,6 +76,16 @@ func WithAuditRealmID(r uint) Scope {
 	}
 }
 
+// WithIncludeTest returns a scope that adds querying for Audit events originating from test.
```
sethvargo: It might be better to invert this for the scope given default bool values

cmd/server/assets/admin/events/index.html
whaught: done... but that does seem a little hidden

whaught: left the hidden input in always, so it shows up in the URL on search. At least that's find-able if a bit obscure

PR #1166, opened 2020-11-21
Password sentinel for system admin too

whaught:
## Proposed Changes

* Don't give back passwords
* @sethvargo  did this for realm in https://github.com/google/exposure-notifications-verification-server/pull/1163

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -169,15 +169,20 @@ func translate(l *gotext.Locale, key string, vars ...interface{}) (string, error
 
 func templateFuncs() htmltemplate.FuncMap {
 	return map[string]interface{}{
-		"joinStrings":    strings.Join,
-		"trimSpace":      project.TrimSpace,
-		"stringContains": strings.Contains,
-		"toLower":        strings.ToLower,
-		"toUpper":        strings.ToUpper,
-		"safeHTML":       safeHTML,
-		"selectedIf":     selectedIf,
-		"t":              translate,
+		"joinStrings":      strings.Join,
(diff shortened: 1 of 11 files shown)


sethvargo (changes requested):
I think you need to plumb the password sentinel through the admin controller?

whaught:
> I think you need to plumb the password sentinel through the admin controller?

oops I missed part of this... done

pkg/render/render.go
```diff
@@ -169,17 +169,22 @@ func translate(l *gotext.Locale, key string, vars ...interface{}) (string, error
...
-		"t":              translate,
+		"joinStrings":      strings.Join,
+		"trimSpace":        project.TrimSpace,
+		"stringContains":   strings.Contains,
+		"toLower":          strings.ToLower,
+		"toUpper":          strings.ToUpper,
+		"safeHTML":         safeHTML,
+		"selectedIf":       selectedIf,
+		"t":                translate,
+		"passwordSentinel": pwdSentinel,
```
sethvargo, line 180: I think this will bring it in as a func, not a variable (e.g. `passwordSentinel` vs `.passwordSentinel`) - all templates need to be updated.

sethvargo (changes requested):

pkg/render/render.go
whaught: ... I did that

sethvargo (approved):
/lgtm

pkg/render/render.go
sethvargo: Weird - I think GitHub was only showing me a partial diff.

Issue #1168, opened 2020-11-21
Require password re-validation or MFA to reset current user's password


sethvargo:
### TL;DR
The current user needs to provide something at reset time (either MFA or current password).

whaught:
This makes a lot of sense for us to do. Firebase sort-of does this by requiring _recent_ auth. Practically that means that if you haven't logged in in the last two minutes, our password change fails and kicks you out jarringly back to the sign-in page. Requiring reauthenticate right before change will eliminate the mystery failure (and is more secure).

PR #1171, opened 2020-11-21
Always reauthenticate before password change

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1168

## Proposed Changes

* Share the authentication code with login
* Always have the user reauthenticate before changing password
* This eliminates issues with Firebase login-recency requirement and is more secure (eg. shared machines).
* Start moving some out stuff to application.js where it can be better cached


cmd/server/assets/login/_loginscripts.html
```diff
@@ -155,67 +155,289 @@
...
       $retyped.addClass("text-muted");
+      $retyped.removeClass("text-danger");
     }
 
     {{end}}
     return valid;
   }
 {{end}}
+
+{{define "login/login-js"}}
```
sethvargo, line 235: Is it feasible to move this to application.js while we're moving things around? Anything that depends on server-side information we can populate in the HTML with hidden fields and then pull them in the js.

sethvargo (commented):
/lgtm
/hold 
for one comment

cmd/server/assets/login/_loginscripts.html
whaught: move password complexity into application.js and moved server-side stuff into a js struct to use as a param.

still looking at the rest, but I get a little concerned by the complexity of this refactor

sethvargo (approved):
/lgtm

PR #1179, opened 2020-11-24
Try using a single database instance

sethvargo:
Like https://github.com/google/exposure-notifications-server/pull/1197, but for this repo.

**Release Note**

```release-note
NONE
```


sethvargo:
Benchmarks on my Mac:

```text
$ go test -count=1 ./pkg/database/...
before: 6.38s user 4.49s system 15% cpu 1:10.91 total
after:  3.03s user 1.94s system 37% cpu 11.408 total
```

```text
$ make test-acc
before: 82.16s user 25.41s system 103% cpu 1:44.14 total
after:  69.93s user 20.64s system 281% cpu 32.165 total
```

Interestingly, the new tests are 3-8x faster, but use ~2.5x the CPU. I'm guessing this is because there's a shared database pool, more tests can run in parallel (using more CPU).

sethvargo:
Ah, these tests won't pass until we merge and deploy https://github.com/GoogleCloudPlatform/oss-test-infra/pull/626. `WITH (FORCE)` is a postgres 13 thing, but our envvars are still requesting Postgres 12. 

Good news is that cleanup is the _last_ step, meaning it's very likely this shaved a full minute off of CI runtime, but we won't know for certain until that prow config is merged and deployed.

/cc @chaodaiG

whaught (approved):

PR #1181, opened 2020-11-24
Capture ctx pointer when recording latency.

### LINKED ISSUE #1180: Ream id is missing from some metric
Take the following MQL as an example:

```
fetch generic_task
| metric
    'custom.googleapis.com/opencensus/en-verification-server/api/issue/request_count'
| align rate(1m)
| every 1m
| group_by [metric.realm],
    [value_request_count_aggregate: aggregate(value.request_count)]
```

If zoom out to 1 months of history data, we can see prior to Nov 10 the group by works as intended. We had a gap between Nov 10 and Nov 11, presumably due to https://github.com/google/exposure-notifications-verification-server/issues/1064. But after the data is back, all of them have realm label containing empty string.

yegle:
The realm id is usually added to ctx much later in the function body. We
need the defer of observability.RecordLatency() at the beginning of the
function body to accurately record the latency.

Fixes #1180

### DIFF
--- a/pkg/clients/e2e.go
+++ b/pkg/clients/e2e.go
@@ -77,7 +77,7 @@ func RunEndToEnd(ctx context.Context, config *config.E2ETestConfig) error {
 	// Parameterize observability.RecordLatency()
 	recordLatency := func(ctx context.Context, start time.Time, step string) {
 		stepMutator := tag.Upsert(stepTagKey, step)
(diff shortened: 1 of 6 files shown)


sethvargo (approved):
/lgtm

I don't love pointers to contexts, but this is really clean. Good work.

PR #1183, opened 2020-11-24
Mark e2e and testing events as test

### LINKED ISSUE #1153: Exclude e2e from audit logs

sethvargo:
Closes #1155
Fixes #1153

**Release Note**

```release-note
Mark e2e and testing events as "test" to filter them out from audit entries.
```

### DIFF
--- a/cmd/server/assets/admin/events/index.html
+++ b/cmd/server/assets/admin/events/index.html
@@ -33,14 +33,8 @@
       </div>
 
       <div class="card-body">
-        <form method="GET" action="/admin/events" id="search-form">
-          <!--
-            hidden query to filter on test events values are:
-               "?include_test=false" // Filters out test events. Default if omitted.
-               "?include_test=only" // shows only test events
-               "?include_test=all" // shows all test and non-test events
-           -->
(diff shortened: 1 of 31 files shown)


cmd/server/assets/admin/events/index.html
```diff
@@ -33,7 +33,8 @@ <h1>System event log</h1>
       </div>
 
       <div class="card-body">
-        <form method="GET" action="/admin/events" id="search-form">
+        <form method="GET" id="search-form">
+          <!-- There's a hidden query param to filter on test `include_test=true` -->
```
whaught, line 37: Note: golang html template doesn't render comments.

That may or may not be desirable. Then again we also have golang `{{/* comment */}}` comments, so that seems like a weird choice. Ask me how I feel about golang templating sometime.

whaught (approved):
/lgtm

PR #1185, opened 2020-11-24
Lookup realm before passing in ID

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/security/code-scanning/7

**Release Note**

```release-note
Lookup realm before passing in ID
```

/assign @whaught

### DIFF
--- a/pkg/controller/admin/events.go
+++ b/pkg/controller/admin/events.go
@@ -68,20 +68,21 @@ func (c *Controller) HandleEventsShow() http.Handler {
 		case "":
 			// All events
 		case "0":
-			scopes = append(scopes, database.WithAuditRealmID(0))
 			realm = &database.Realm{
 				Model: gorm.Model{ID: 0},
 				Name:  "System",
 			}
 		default:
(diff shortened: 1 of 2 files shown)


pkg/controller/admin/events.go
```diff
@@ -74,11 +74,9 @@ func (c *Controller) HandleEventsShow() http.Handler {
 				Name:  "System",
 			}
 		default:
-			if id, err := strconv.ParseUint(realmID, 10, 64); err != nil {
-				scopes = append(scopes, database.WithAuditRealmID(uint(id)))
-			}
-
-			if realm, err = c.db.FindRealm(realmID); err != nil {
+			scopes = append(scopes, database.WithAuditRealmID(realmID))
```
whaught, line 77:
```
realm, err = c.db.FindRealm(realmID)
scopes = append(scopes, database.WithAuditRealmID(realm.ID))
```

I think this would be the same without needing interface{} params necessarily

sethvargo: Yea, but that adds another db call and I think it leaks the existence of a realm

whaught (approved):

Issue #1092, opened 2020-11-13
Sync android apps to realm mobile apps


mikehelmick:
### TL;DR

Automatically import Google ENX Application configuration into the verification server IFF there is a realm that matches the ISO code.

Google will publish a file of public mobile apps @
https://www.gstatic.com/exposurenotifications/apps.json

The format is:

```
{
  "apps": [
    {
      "region": "US-CA",
      "is_enx": true,
      "android_target": {
        "namespace": "android_app",
        "package_name": "APK NAME",
        "sha256_cert_fingerprints": "FINGERPRINT",
      }
    }
  ]
}
```

### Design

**Proposal**

- Create a new job for importing this periodically 
- If a realm exists for the specified region code
    - ensure that an android app exists w/ the associated package name
        - if not, create it
        - if yes, ensure the sha256 is in the list (this is a merge, there may be dev fingerprints there)

**Alternatives considered**

This is currently manual and error prone, we're aiming to to better.

/kind enhancement

PR #1191, opened 2020-11-24
Sync android apps from json file

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1092

## Proposed Changes

* Unmarshal the json file for the list of android apps
* Structs to hold the result
* Look up/cache realms
* Look up/cache mobile-apps per realm
* If the file app doesn't exist, create it.

**Release Note**

```release-note
New endpoint to sync android apps to realms from a configurable .json file URL.
This will be hooked up to a cron job in the future.
```


pkg/controller/appsync/appsync.go
```diff
@@ -45,7 +46,23 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
-		// TODO(whaught): implement this
-		controller.InternalError(w, r, c.h, errors.New("not implemented"))
+		resp, err := http.Get(c.config.AppSyncURL)
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
```
sethvargo, line 56: We need to add more diligence and resilience here. Check how we do some of the downloading elsewhere. We should put a max size reader on this too.

pkg/controller/appsync/response.go
```diff
@@ -0,0 +1,31 @@
...
+type App struct {
+	Region        string `json:"region"`
+	IsEnx         bool   `json:"is_enx"`
+	AndroidTarget `json:"android_target"`
+}
+
+type AndroidTarget struct {
+	Namespace              string `json:"namespace"`
+	PackageName            string `json:"package_name"`
+	SHA256CertFingerprints string `json:"sha256_cert_fingerprints"`
```
sethvargo, line 30: Should any of these be omitempty?

pkg/config/appsync_server_config.go
```diff
@@ -41,7 +42,9 @@ type AppSyncConfig struct {
 	RateLimit uint64 `env:"RATE_LIMIT,default=60"`
 
 	// AppSync config
-	AppSyncURL string `env:"APP_SYNC_URL, default=https://www.gstatic.com/exposurenotifications/apps.json"`
```
sethvargo, line 44: Should we consider a world where app syncs are actually a database table and it's possible to sync from multiple URLs?

pkg/config/appsync_server_config.go
```diff
@@ -41,7 +42,9 @@ type AppSyncConfig struct {
...
+	AppSyncURL    string        `env:"APP_SYNC_URL, default=https://www.gstatic.com/exposurenotifications/apps.json"`
```
sethvargo, line 45: Is this a sane default? Should it be required and have no default?

pkg/config/appsync_server_config.go
```diff
@@ -41,7 +42,9 @@ type AppSyncConfig struct {
...
+	FileSizeLimit int64         `env:"APP_SYNC_SIZE_LIMIT, default=10240"`
```
sethvargo, line 46: Is this bytes, num records? Lets add a suffix to make it clearer.

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
 func (c *Controller) HandleSync(ctx context.Context) http.Handler {
+	type AppSyncResult struct {
+		OK     bool    `json:"ok"`
+		Errors []error `json:"errors,omitempty"`
+	}
+
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// TODO(whaught): implement this
-		controller.InternalError(w, r, c.h, errors.New("not implemented"))
+		logger := logging.FromContext(ctx).Named("appsync.HandleSync")
```
sethvargo, line 58: Pull the logger from `r.Context()` so we get the request ID

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
 
 // HandleSync performs the logic to sync mobile apps.
 func (c *Controller) HandleSync(ctx context.Context) http.Handler {
```
sethvargo, line 51: Don't pass in a context here unless you need it.

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+		var merr *multierror.Error
+		realms := map[string]*database.Realm{}
+		appsByRealm := map[uint][]*database.MobileApp{}
+
+		for _, app := range apps.Apps {
+			realm, has := realms[app.Region]
+			if !has {
+				realm, err := c.db.FindRealmByRegion(app.Region)
+				if err != nil {
+					merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm: %w", err))
```
sethvargo, line 89:
```suggestion
					merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm %s: %w", app.Region, err))
```

or something more informative

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+		}
+
+		var merr *multierror.Error
+		realms := map[string]*database.Realm{}
+		appsByRealm := map[uint][]*database.MobileApp{}
+
+		for _, app := range apps.Apps {
+			realm, has := realms[app.Region]
+			if !has {
+				realm, err := c.db.FindRealmByRegion(app.Region)
```
sethvargo, line 87: Do you mean to shadow `realm` here?

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+				}
+				realms[app.Region] = realm
+			}
+
+			realmApps, has := appsByRealm[realm.ID]
+			if !has {
+				// Only Android supported for sync.
+				realmApps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
+				if err != nil {
+					merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm: %w", err))
```
sethvargo, line 100:
```suggestion
					merr = multierror.Append(merr, fmt.Errorf("unable to list apps for realm %s: %w", realm.ID err))
```

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+					continue
+				}
+				appsByRealm[realm.ID] = realmApps
+			}
+
+			has = false
```
sethvargo, line 106: I'm not sure I follow what this is doing?

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+					merr = multierror.Append(merr, fmt.Errorf("failed saving mobile app: %v", err))
+					continue
+				}
+			}
+		}
+
+		// If there are any errors, return them
+		if merr != nil {
+			if errs := merr.WrappedErrors(); len(errs) > 0 {
+				logger.Errorw("failed to sync apps", "errors", errs)
```
sethvargo, line 137: You're returning an error to the client, so you shouldn't need to log here.

pkg/controller/appsync/appsync.go
```diff
@@ -44,8 +49,98 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+
+		client := http.Client{Timeout: c.config.Timeout}
+		resp, err := client.Get(c.config.AppSyncURL)
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(io.LimitReader(resp.Body, c.config.FileSizeLimit))
```
sethvargo, line 68: I think it'd be better if we didn't have this intermediate []byte data structure:

```go
var apps AppsResponse
if err := json.NewDecoder(io.LimitReader(resp.Body, c.config.FileSizeLimit).Decode(&apps); err != nil {
  // ...
}
```

pkg/controller/appsync/appsync.go
whaught: Sorry I'm kind of working in the open here (this is not ready). I'm trying to match the given JSON app to the existing `MobileApp` (per realm for android). At the moment I'm matching on the SHA in that list, but trying to understand what `package_name` maps to how how to set Name, URL, and AppID

pkg/controller/appsync/appsync.go
```diff
@@ -43,9 +46,93 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+					merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm %s: %w", app.Region, err))
+					continue
+				}
+				realms[app.Region] = realm
+			}
+
+			realmApps, has := appsByRealm[realm.ID]
+			if !has {
+				// Only Android supported for sync.
+				realmApps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
```
sethvargo, line 91: A good follow-up here would be to refactor ListActiveAppsByOS into `ListActiveApps` and a scope 😄

pkg/controller/appsync/appsync.go
flagxor: So I probably did a useless thing in having the 'target' field mirror the content that will end up at https://www.example.com/.well-known/assetlinks.json:
https://developer.android.com/training/app-links/verify-site-associations

But presumably you care more about mapping to existing config fields.

package_name == AppID
URL === https://play.google.com/store/apps/details?id=<package_name>
e.g. https://play.google.com/store/apps/details?id=gov.co.cdphe.exposurenotifications
If you'd prefer, I can add a field for the URL explicitly in the generated json?

For Name, I'm unsure, we have the app name for various locales, so I could give you the default locale, but I had assumed we didn't need to check these match, as the PHA might configure the device key once and then rename their app and not want to keep both in sync? WDYT?

whaught: I've added in that with the a default name "[package_name] Android App", but it's not my favorite.
The more that's in the published file, the less the server needs to guess / hardcode, so that's my preference.

pkg/config/appsync_server_config.go
whaught: that's a good point. Left it with no default and kill processing immediately if none is configured.

pkg/config/appsync_server_config.go
mikehelmick: I think the single URL is fine.

pkg/controller/appsync/appsync.go
```diff
@@ -16,16 +16,22 @@
...
 
+	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
+	"github.com/hashicorp/go-multierror"
 )
 
+const playStoreLink = `https://play.google.com/store/apps/details?id=`
```
mikehelmick, line 33: are there international play store links that we need to consider?
@flagxor

pkg/controller/appsync/appsync.go
```diff
@@ -43,9 +49,95 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
-		// TODO(whaught): implement this
-		controller.InternalError(w, r, c.h, errors.New("not implemented"))
+		logger := logging.FromContext(r.Context()).Named("appsync.HandleSync")
+
+		if c.config.AppSyncURL == "" {
+			controller.InternalError(w, r, c.h, errors.New("no APP_SYNC_URL configured"))
+			return
+		}
+
+		client := http.Client{Timeout: c.config.Timeout}
```
mikehelmick, line 66: should we move this to clients
https://github.com/google/exposure-notifications-verification-server/tree/main/pkg/clients

pkg/controller/appsync/appsync.go
```diff
@@ -43,9 +49,95 @@ func New(config *config.AppSyncConfig, db *database.Database, h *render.Renderer
...
+		if err := json.NewDecoder(io.LimitReader(resp.Body, c.config.FileSizeLimitBytes)).Decode(&apps); err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		var merr *multierror.Error
+		realms := map[string]*database.Realm{}
+		appsByRealm := map[uint][]*database.MobileApp{}
+
+		for _, app := range apps.Apps {
```
mikehelmick, line 85: and extract this to a function to make it testable

whaught: I even added a test!

internal/project/random.go
```diff
@@ -0,0 +1,31 @@
...
+	"crypto/rand"
+	"crypto/sha256"
+	"fmt"
+)
+
+// RandomString generates a random string of 32 characters in length
+func RandomString() (string, error) {
+	b := make([]byte, 512)
+	if _, err := rand.Read(b[:]); err != nil {
+		return "", fmt.Errorf("failed to generate random: %v", err)
```
sethvargo, line 28:
```suggestion
		return "", fmt.Errorf("failed to generate random: %w", err)
```

pkg/config/appsync_server_config.go
```diff
@@ -55,10 +58,12 @@ func NewAppSyncConfig(ctx context.Context) (*AppSyncConfig, error) {
 }
 
 func (c *AppSyncConfig) Validate() error {
-	if url, err := url.Parse(c.AppSyncURL); err != nil {
-		return fmt.Errorf("unable to parse APP_SYNC_URL: %v", err)
-	} else if url == nil {
-		return errors.New("expecting a value for APP_SYNC_URL")
+	if c.AppSyncURL != "" {
```
sethvargo, line 61: Nit: It'd be less nesting if we checked the negative case and returned `nil`:

```go
if c.AppSyncURL == "" {
  return nil
}

if url, err := ...
```

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
-		// TODO(whaught): implement this
-		controller.InternalError(w, r, c.h, errors.New("not implemented"))
+		apps, err := clients.AppSync(c.config.AppSyncURL, c.config.Timeout, c.config.FileSizeLimitBytes)
+		if err != nil {
+			controller.InternalError(w, r, c.h, err)
+			return
+		}
+
+		// If there are any errors, return them
+		if merr := c.syncApps(r.Context(), apps); merr != nil {
```
sethvargo, line 47: For consistency, can you pull the `ctx` out at the top of the HandlerFunc?

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+					Errors: errs,
+				})
+				return
+			}
+		}
+		c.h.RenderJSON(w, http.StatusOK, &AppSyncResult{OK: true})
 	})
 }
+
+func (c *Controller) syncApps(ctx context.Context, apps *clients.AppsResponse) *multierror.Error {
```
sethvargo, line 60: doc

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+
+	realms := map[string]*database.Realm{}
+	appsByRealm := map[uint][]*database.MobileApp{}
+
+	for _, app := range apps.Apps {
+		realm, has := realms[app.Region]
+		if !has { // Find this apps region and cache it in our realms map
+			realm, err = c.db.FindRealmByRegion(app.Region)
+			if err != nil {
+				merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm %s: %w", app.Region, err))
```
sethvargo, line 73:
```suggestion
				merr = multierror.Append(merr, fmt.Errorf("unable to lookup realm for region %q: %w", app.Region, err))
```

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+		if !has { // Find all of the apps for this realm and cache that list in our appByRealmMap
+			realmApps, err = c.db.ListActiveApps(realm.ID, database.WithAppOS(database.OSTypeAndroid))
+			if err != nil {
+				merr = multierror.Append(merr, fmt.Errorf("unable to list apps for realm %d: %w", realm.ID, err))
+				continue
+			}
+			appsByRealm[realm.ID] = realmApps
+		}
+
+		has = false // Find out if this realm's applist already has an app with this fingerprint.
```
sethvargo, line 89: It might be clearer to pull this into a helper? I don't like that we're shadowing `has` so many times.

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+				has = true
+			}
+			if a.Name == generateAppName(app) {
+				hasGeneratedName = true
+			}
+		}
+
+		// Didn't find an app. make one.
+		if !has {
+			logger.Infof("App not found during sync. Adding app %#v", app)
```
sethvargo, line 102:
```suggestion
			logger.Infow("App not found during sync, adding", "app", app)
```

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+
+			name := generateAppName(app)
+			if hasGeneratedName { // add a random string to names on collision
+				s, _ := project.RandomString()
```
sethvargo, line 106: Handle this error

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+				name += " " + s[:8]
```
sethvargo, line 107: Should the project helper take an extra param of "length"?

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+			}
+
+			newApp := &database.MobileApp{
+				Name:    name,
+				RealmID: realm.ID,
+				URL:     playStoreLink + app.PackageName,
```
sethvargo, line 113: I'm reallllllll scared of manually building URLs - been bitten in the past. I would prefer if `playStoreLink` were a `url.URL` and we build the qs param `id` off the URL and then converted it to a string.

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+				OS:      database.OSTypeAndroid,
+				SHA:     app.SHA256CertFingerprints,
+				AppID:   app.PackageName,
+			}
+			if err := c.db.SaveMobileApp(newApp, database.System); err != nil {
+				merr = multierror.Append(merr, fmt.Errorf("failed saving mobile app: %v", err))
```
sethvargo, line 119:
```suggestion
				merr = multierror.Append(merr, fmt.Errorf("failed saving mobile app: %w", err))
```

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,118 @@
...
+				continue
+			}
+		}
+	}
+	return merr
+}
+
+func generateAppName(app clients.App) string {
```
sethvargo, line 127: To avoid the extra check, should we always append a random ID?

pkg/controller/appsync/appsync_test.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		merr := c.syncApps(context.Background(), resp)
+		if merr != nil {
+			t.Fatalf(merr.ErrorOrNil().Error())
+		}
+
+		apps, err := db.ListActiveApps(realm.ID)
+		if err != nil {
+			t.Fatal("failed to list apps", err)
```
sethvargo, line 90: Does Fatal work with two args like this?

pkg/controller/appsync/appsync_test.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		if len(apps) != 2 {
```
sethvargo, line 93: `got, want` syntax would be better here instead of hardcoding 2 twice

pkg/controller/appsync/appsync_test.go
```diff
@@ -0,0 +1,97 @@
...
+						PackageName:            "testAppId2",
+						SHA256CertFingerprints: "BB:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA",
+					},
+				},
+			},
+		}
+
+		merr := c.syncApps(context.Background(), resp)
+		if merr != nil {
+			t.Fatalf(merr.ErrorOrNil().Error())
```
sethvargo, line 85: Calling Error on nil will panic I think?

sethvargo (changes requested):

pkg/controller/appsync/appsync.go
whaught: I started that way. I kind of think the random is ugly (and collisions are unlikely in the real-world) so I figured to optimize for nice names.

pkg/config/appsync_server_config.go
whaught: I was worried if we added more validations later someone would miss that. Done.

pkg/controller/appsync/appsync_test.go
whaught: I believe it does. (just prints each of them)

pkg/controller/appsync/appsync.go
whaught: it should, but for now it was easier to move the existing one this way.

pkg/controller/appsync/appsync.go
flagxor: Play store will use device locale so I think this should be the same regardless of locale.

pkg/controller/appsync/appsync.go
```diff
@@ -12,40 +12,152 @@
...
+
+		// Didn't find an app. make one.
+		if !hasSHA {
+			logger.Infow("app not found during sync, adding", "app", app)
+
+			name := generateAppName(app)
+			if hasGeneratedName { // add a random string to names on collision
+				s, err := project.RandomString()
+				if err != nil {
+					logger.Errorw("error generating app name", "error", err)
```
sethvargo, line 104: Any reason not to add to the multierror instead?

sethvargo (approved):
/lgtm

/hold

one nit

pkg/controller/appsync/appsync.go
whaught: nah let's do that. more consistent

flagxor (approved):
lgtm

PR #1193, opened 2020-11-25
Switch symptomDate to testDate for bulk uploader

whaught:
## Proposed Changes

* Looks like users of this scenario are currently using issueAPI with testDate only. This prompts the user at the app for their symptom date.
* Bulk uploader should use testDate.
* Optional 3rd row of csv for symptomDate
* BUILD BREAK FIX. Add exporter ctx pulled in from key-server 0.17.0 dep


mikehelmick (approved):
Correct - in this use case, testDate is likely known and symptomDate is not

PR #1195, opened 2020-11-25
Add random padding to issue requests

whaught:
## Proposed Changes

* Adds a random encoded string to the "padding" field
* This conforms to the API guidance to vary request sizes which can help obfuscate attempts to sniff traffic.

**Release Note**

```release-note
Add random padding to requests in bulk-uploader
```


sethvargo (approved):
/lgtm

PR #1196, opened 2020-11-26
Use scopes for list mobile apps by OS

whaught:
## Proposed Changes

* Change ListActiveAppsByOS to ListActiveApps that takes a list of scopes
* Pass in a ByOS scope

Note: This is simple, but will lead others to use more scopes if we need other variations of this.

### DIFF
--- a/pkg/controller/associated/android.go
+++ b/pkg/controller/associated/android.go
@@ -39,8 +39,7 @@ func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
 		return nil, fmt.Errorf("unable to lookup realm: %w", err)
 	}
 
-	scopes := []database.Scope{database.WithAppOS(database.OSTypeAndroid)}
-	apps, err := c.db.ListActiveApps(realm.ID, scopes...)
+	apps, err := c.db.ListActiveApps(realm.ID, []database.Scope{database.WithAppOS(database.OSTypeAndroid)}...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get android data: %w", err)
 	}
--- a/pkg/controller/associated/android.go
+++ b/pkg/controller/associated/android.go
@@ -39,7 +39,7 @@ func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
 		return nil, fmt.Errorf("unable to lookup realm: %w", err)
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/database/scopes.go
```diff
@@ -77,6 +77,19 @@ func WithAuditRealmID(id uint) Scope {
 	}
 }
 
+// WithRealmSearch returns a scope that adds querying for realms by name. It's
+// only applicable to functions that query Realm.
+func WithRealmSearch(q string) Scope {
```
whaught, line 82: this is just moved from below to be with the other realm scopes.

pkg/controller/associated/android.go
```diff
@@ -39,7 +39,8 @@ func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
 		return nil, fmt.Errorf("unable to lookup realm: %w", err)
 	}
 
-	apps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
+	scopes := []database.Scope{database.WithAppOS(database.OSTypeAndroid)}
```
sethvargo, line 42: Nit, but since this is the only scope, why not just:

```go
apps, err := c.db.ListActiveApps(realm.ID, database.WithAppOS(database.OSTypeAndroid))
```

throughout?

whaught: sure we can do that

pkg/controller/associated/android.go
```diff
@@ -39,7 +39,7 @@ func (c *Controller) getAndroidData(region string) ([]AndroidData, error) {
 		return nil, fmt.Errorf("unable to lookup realm: %w", err)
 	}
 
-	apps, err := c.db.ListActiveAppsByOS(realm.ID, database.OSTypeAndroid)
+	apps, err := c.db.ListActiveApps(realm.ID, []database.Scope{database.WithAppOS(database.OSTypeAndroid)}...)
```
sethvargo, line 42: Sorry to be a butt, but I don't think we need to allocate a slice here. The signature is variadic, so you should be able to:

```go
apps, err := c.db.ListActiveApps(realm.ID, database.WithAppOS(database.OSTypeAndroid))
```

whaught: oops... I was not paying attention

PR #1198, opened 2020-11-26
Add ExternalIssuers, refactor statistics APIs

### LINKED ISSUE #1076: add audit-ID to issue API
### TL;DR

Add an `auditID` field to the IssueCodeRequest API. This should be ignored via the web UI (the logged in user is the auditID).

This auditID should be used to calculate the current realm statistics alongside actual logged in users.

The use case for this is a 3rd party EHR type integration where they want to (on their side) track which physicians have issued codes by assigning a unique ID for that purpose, and pass that unique ID.

Then if a realm administrator sees that API user `123-abc-456` has issued a bunch of codes, they can know which provider has the API key associated with that audit ID and can go to that EHR provider to find who they physician is.

Needs work on the
* Issue API
* stats collection
* stats display + export

/kind enhancement

sethvargo:
Closes #1159
Fixes #1076 

This commit refactors stats and stats APIs to return standardized responses. Instead of munging JSON on the server and the client, the server defines a custom JSON marshaler and the client uses the JSON API to request the chart data. This greatly simplifies the HTML template and controller logic.

In the same vein, there's a custom MarshalCSV interface that all the stats providers implement, and new renderer functions for creating CSVs.

The biggest part of this refactor is the way stats are calculated. Previously stats were naively calculated with a JOIN, but that left zeros or missing data in the structure. The new stats calculation ensures no gaps and fills in zero values where data is missing.

localhost_8080_realm_stats

**Release Note**

```release-note
**New statistics API** - Statistics via the JSON API are now returned with more structure and any gaps in dates/users are zeroed to ensure data continuity. 
**External issuers** - When issuing a code via the adminapi, callers can specify an `ExternalIssuerID` to associate the issuance with an external identifier. See the API documentation for more information.
```

/assign @whaught


pkg/render/csv.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Rendering worked, flush to the response. Force as a download.
+	w.Header().Set("Content-Type", "text/csv")
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment;filename=%s", filename))
+	w.WriteHeader(code)
+	if _, err := w.Write(b); err != nil {
+		// We couldn't write the buffer. We can't change the response header or
+		// content type if we got this far, so the best option we have is to log the
+		// error.
+		r.logger.Errorw("failed to write csv to response", "error", err)
```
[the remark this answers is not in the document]
sethvargo: Yup, intended.

docs/api.md
```diff
@@ -195,6 +195,17 @@ Request a verification code to be issued. Accepts [optional] symptom date and te
   the padding.
 * `uuid` is optional as request input. The server will generate a uuid on response if omitted.
   * This is a handle which allows the issuer to track status of the issued verification code.
+* `externalIssuerID` is a optional information supplied by the API caller to
```
whaught, line 198:
```suggestion
* `externalIssuerID` is an optional field supplied by the API caller to
```

pkg/controller/realmadmin/show.go
```diff
@@ -16,36 +16,97 @@ package realmadmin
...
+				controller.InternalError(w, r, c.h, err)
+				return
+			}
+
+			c.h.RenderCSV(w, http.StatusOK, filename, stats)
+		case strings.HasSuffix(pth, ".json"):
+			var stats json.Marshaler
+			var err error
+
+			switch r.URL.Query().Get("scope") {
```
whaught, line 78: don't see nested switches much

pkg/database/external_issuer_stats.go
```diff
@@ -0,0 +1,155 @@
...
+	var b bytes.Buffer
+	w := csv.NewWriter(&b)
+
+	if err := w.Write([]string{"date", "realm_id", "issuer_id", "codes_issued"}); err != nil {
+		return nil, fmt.Errorf("failed to write CSV header: %w", err)
+	}
+
+	for i, stat := range s {
+		if err := w.Write([]string{
+			stat.Date.Format("2006-01-02"),
```
whaught, line 58: random, but I've seen "2006-01-02" around  a bunch. I wonder if we should put `const ISO8601Date` in the `project` namespace

pkg/database/external_issuer_stats.go
```diff
@@ -0,0 +1,155 @@
...
+	return b.Bytes(), nil
+}
+
+type jsonExternalIssuerStat struct {
+	RealmID uint                           `json:"realm_id"`
+	Stats   []*jsonExternalIssuerStatStats `json:"statistics"`
+}
+
+type jsonExternalIssuerStatStats struct {
+	Date time.Time                     `json:"date"`
```
whaught, line 81: maybe I'm tired or its my font, but "date" and "data" scan really similarly. I wonder if `issuerData` would be nice.

pkg/database/external_issuer_stats.go
sethvargo: https://github.com/google/exposure-notifications-verification-server/issues/1201

pkg/controller/realmadmin/show.go
sethvargo: I might pull the JSON/CSV method out into its own function in a later PR.

docs/api.md
sethvargo: Done

pkg/database/external_issuer_stats.go
sethvargo: Fixed

sethvargo:
@whaught fixed up suggestions - PTAL

whaught (approved):
/lgtm

PR #1227, opened 2020-12-01
Batch API for code issuing

whaught:
## Proposed Changes

* Refactor the issue code processing for reuse
* Creates a small results struct to share observability
* Create a Batch version of the issue code API
    * Calls our issue helper in our loop

The routing is currently commented out until I finish testing / size limits / etc, but looking for feedback on this partial refactor.


pkg/api/api.go
```diff
@@ -225,6 +225,19 @@ type IssueCodeResponse struct {
...
+// BatchIssueCodeRequest defines the request for issuing many codes at once.
+type BatchIssueCodeRequest struct {
+	Codes []*IssueCodeRequest `json:"codes"`
+}
+
+// BatchIssueCodeResponse defines the response for BatchIssueCodeRequest.
+type BatchIssueCodeResponse struct {
+	Codes []*IssueCodeResponse `json:"codes"`
+
+	Error     string `json:"error"`
```
sethvargo, line 237: `omitempty` here too

pkg/controller/issueapi/batch_issueapi.go
```diff
@@ -0,0 +1,115 @@
...
+		if c.config.IsMaintenanceMode() {
+			c.h.RenderJSON(w, http.StatusTooManyRequests,
+				api.Errorf("server is read-only for maintenance").WithCode(api.ErrMaintenanceMode))
+			return
+		}
+
+		resp := &api.BatchIssueCodeResponse{}
+
+		ctx := r.Context()
+		realm := controller.RealmFromContext(ctx)
```
sethvargo, line 40: Make sure realm exists first

pkg/controller/issueapi/batch_issueapi.go
```diff
@@ -0,0 +1,115 @@
...
+		if !realm.AllowBulkUpload {
+			controller.Unauthorized(w, r, c.h)
+			return
+		}
+
+		hasSMSConfig, err := realm.HasSMSConfig(c.db)
+		if err != nil || !hasSMSConfig {
+			controller.InternalError(w, r, c.h, err)
```
sethvargo, line 48: I don't think we want to throw an internal error (page someone) if there's no SMS config

pkg/controller/issueapi/batch_issueapi.go
```diff
@@ -0,0 +1,115 @@
...
+			httpCode:  http.StatusOK,
+			obsBlame:  observability.BlameNone,
+			obsResult: observability.ResultOK(),
+		}
+		defer observability.RecordLatency(&ctx, time.Now(), mLatencyMs, &result.obsBlame, &result.obsResult)
+
+		var request api.BatchIssueCodeRequest
+		if err := controller.BindJSON(w, r, &request); err != nil {
+			c.h.RenderJSON(w, http.StatusBadRequest, api.Error(err))
+			result.obsBlame, result.obsResult = observability.BlameClient, observability.ResultError("FAILED_TO_PARSE_JSON_REQUEST")
```
sethvargo, line 62: Nit: can you move the o11y above the RenderJSON and return - it'll make the future regex I have to write easier 😄

sethvargo: (throughout)

pkg/controller/issueapi/batch_issueapi.go
```diff
@@ -0,0 +1,115 @@
...
+			return
+		}
+
+		authApp, user, err := c.getAuthorizationFromContext(r)
```
sethvargo, line 66: Hmm - you're loading the realm above, but then also loading it again here?

pkg/controller/issueapi/batch_issueapi.go
```diff
@@ -0,0 +1,115 @@
...
+
+		// TODO(whaught): batch size limits
+		logger := logging.FromContext(ctx).Named("issueapi.HandleBatchIssue")
+
+		resp.Codes = make([]*api.IssueCodeResponse, len(request.Codes))
+		for i, singleIssue := range request.Codes {
+			resp.Codes[i] = c.issue(ctx, singleIssue, realm, user, authApp, result)
+			if result.errorReturn != nil {
+				if result.httpCode == http.StatusInternalServerError {
+					controller.InternalError(w, r, c.h, errors.New(result.errorReturn.Error))
```
sethvargo, line 101: Probably want a multi-error instead?

pkg/controller/issueapi/issue.go
```diff
@@ -143,230 +150,228 @@ func (c *Controller) HandleIssue() http.Handler {
...
-			c.h.RenderJSON(w, http.StatusBadRequest,
-				api.Errorf("unsupported test type: %v", request.TestType))
-			blame = observability.BlameClient
-			result = observability.ResultError("UNSUPPORTED_TEST_TYPE")
-			return
+		c.h.RenderJSON(w, http.StatusOK, resp)
+	})
+}
+
+func (c *Controller) issue(ctx context.Context, request *api.IssueCodeRequest,
```
sethvargo, line 174: This API scares me since it mutates some of the provided parameters...

pkg/api/api.go
```diff
@@ -225,6 +225,19 @@ type IssueCodeResponse struct {
 	ErrorCode string `json:"errorCode,omitempty"`
 }
 
+// BatchIssueCodeRequest defines the request for issuing many codes at once.
+type BatchIssueCodeRequest struct {
```
mikehelmick, line 229: do we intend for this to be a public API or for use by the UI only.
what are the failure semantics of some work / some don't?

pkg/controller/issueapi/batch_issueapi.go
mikehelmick: this is what I asked in the API definition about failure semantics - we need to be able to know which which rows were successfully processed so they don't get retreied

they may have send real people SMS already

pkg/api/api.go
whaught: Ultimately public, and used by the UI.

The BatchIssueCodeResponse has error fields and a list of individual responses (which also have error codes). The response is error with details if the batch request failed overall. If we get to code processing each individual code has its own errorDetails struct in there. We give status-per-code.

Actually now that you bring it up I'm reading Google's official guidelines: https://google.aip.dev/233 which do not allow partial success.... I'm going to ignore that because I don't want to see SMSes sent twice and I don't think we'll be able to make this easily atomic.

mikehelmick: I think it's fine to stop at the first failure - but the client has to know where the failure occurred

we can add error level return conditions specific to this API -i.e. cursor, list of already processed UUID, something like that

since the intent is public API - we need strict limits on the batch sizes. I think even starting very low, 5? 10? would give some performance benefit, but make weird failure scenarios less likely

whaught: right batch size limiting was one of the TODOs I had in there. Put that in and set it to 10 for now. I'm curious what latency will look like with twilio calls and all

pkg/controller/issueapi/batch_issueapi.go
whaught: added in the rest of the error handling

pkg/controller/issueapi/handle_issue.go
```diff
@@ -0,0 +1,104 @@
...
+
+		authApp, user, err := c.getAuthorizationFromContext(r)
+		if err != nil {
+			result.obsBlame = observability.BlameClient
+			result.obsResult = observability.ResultError("MISSING_AUTHORIZED_APP")
+			c.h.RenderJSON(w, http.StatusUnauthorized, api.Error(err))
+			return
+		}
+
+		var realm *database.Realm
```
sethvargo, line 69: I see this pattern in a few places and it confuses me. The [middleware sets the realm from the authApp](https://github.com/google/exposure-notifications-verification-server/blob/eff1dd4c5e9c4cd639cca0fc8bebb0de36d4274f/pkg/controller/middleware/apikey.go#L111) already, and it's not possible to arrive at this page without that middleware running.

pkg/controller/issueapi/handle_issue.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		ctx := observability.WithBuildInfo(r.Context())
+
+		result := &issueResult{
+			httpCode:  http.StatusOK,
+			obsBlame:  observability.BlameNone,
+			obsResult: observability.ResultOK(),
+		}
+		defer recordObservability(&ctx, result)
```
sethvargo, line 51: I dont think this will use the ctx thats later modified

pkg/controller/issueapi/logic.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/google/exposure-notifications-verification-server/pkg/api"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/observability"
+	"github.com/google/exposure-notifications-verification-server/pkg/otp"
+	"github.com/google/exposure-notifications-verification-server/pkg/sms"
+	"go.opencensus.io/stats"
+)
+
+func (c *Controller) issue(ctx context.Context, request *api.IssueCodeRequest,
```
sethvargo, line 36: This function signature is a bit weird because it's unclear what's required and what's option. For example, `db` and `realm` are required - passing `nil` will panic. But `user` and `authApp` are optional and can both be nil.

At the very least, we should document this, but it would be good to add some runtime checks or potentially change the signature to make this more obvious.

pkg/controller/issueapi/logic.go
```diff
@@ -0,0 +1,305 @@
...
+		}, nil
+	}
+
+	// Validate that the request with the provided test type is valid for this realm.
+	if !realm.ValidTestType(request.TestType) {
+		return &issueResult{
+			obsBlame:    observability.BlameClient,
+			obsResult:   observability.ResultError("UNSUPPORTED_TEST_TYPE"),
+			httpCode:    http.StatusBadRequest,
+			errorReturn: api.Errorf("unsupported test type: %v", request.TestType),
```
sethvargo, line 57: .WithCode?

pkg/controller/issueapi/handle_issue.go
whaught: I agree. This is just modelled after what's there, but some of the controllers seem to double-check (with what is unreachable code). Maybe its defensive or was written before middleware? Either way let's simplify

pkg/controller/issueapi/logic.go
whaught: Changed the signature - let's pull what we can off of context instead of passing them all around. Esp if we can assume they already checked/rejected by middleware

pkg/controller/issueapi/logic.go
whaught: done. Also added a `quota_exceeded` code which seems to be missing

pkg/controller/issueapi/logic.go
```diff
@@ -300,6 +300,6 @@ func (c *Controller) getAuthorizationFromContext(r *http.Request) (*database.Aut
 	return authorizedApp, currentUser, nil
 }
 
-func recordObservability(ctx *context.Context, result *issueResult) {
-	observability.RecordLatency(ctx, time.Now(), mLatencyMs, &result.obsBlame, &result.obsResult)
+func recordObservability(ctx context.Context, result *issueResult) {
```
sethvargo, line 303: @yegle I think you said this needed to be a pointer for some reason, but I can't find the PR where you mentioned that.

whaught: I'm going to assume we're trying to capture
`ctx = observability.WithRealmID(ctx, realm.ID)`
I'm not sure if that works - pointers to context is a bit of an odd pattern

whaught: Actually I can't explain why we have `RecordLatency(ctx *context.Context` - it does not seem to really need the ctx pointer and doesn't modify it

sethvargo (approved):
/lgtm

Issue #1202, opened 2020-11-30
Update HostDown playbook


grpandurangipwc:
@yegle 
Could you please update the documentation on HostDown playbook
https://github.com/google/exposure-notifications-verification-server/blob/main/docs/playbooks/alerts/HostDown.md

PR #1228, opened 2020-12-01
Add playbook for HostDown alert.

yegle:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1202


docs/playbooks/alerts/HostDown.md
```diff
@@ -0,0 +1,23 @@
+# HostDown
+
+This is an uptime check on the custom domain. Failing these checks means
```
mariliamelo, line 3: Can you add when this alert would be triggered?

PR #1229, opened 2020-12-01
Fully i18nize issue page

sethvargo:
We're going to submit these for translation, so I removed the spanish one for now.

**Release Note**

```release-note
Fully internationalize strings on code issue page.
```


sethvargo (commented):
/assign @whaught 

I moved codescripts into application.js - there we no Go template dependencies. The only change in the javascript is pulling the text "Expired" from a data attribute so it can be available at runtime.

whaught (approved):

Issue #1242, opened 2020-12-02
Include App Name in MobileApp sync


whaught:
During MobileApp sync we look at a published app list and create new MobileApps that are missing.
@flagxor is adding an app_name field to the published .json file of apps. Include this instead of generating a name.

/assign

PR #1247, opened 2020-12-03
accept app_name for mobile app sync

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1242

## Proposed Changes

* @flagxor added an app_name field today (although it does not yet exist in the export file). Take that as optional and prefer using that field to generating one.


sethvargo (approved):
/lgtm

PR #1249, opened 2020-12-03
point back to the new user

whaught:
## Proposed Changes

* I believe these all will point to the address of `var batchUser` which is not the intent.
    * This data isn't really used on receipt which is why this went uncaught.
* TODO: factor logic out and unit test. Coming soon.

### DIFF
--- a/pkg/controller/user/importbatch.go
+++ b/pkg/controller/user/importbatch.go
@@ -53,7 +53,7 @@ func (c *Controller) HandleImportBatch() http.Handler {
 		newUsers := make([]*api.BatchUser, 0, len(request.Users))
 
 		var batchErr *multierror.Error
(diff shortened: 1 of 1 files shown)


sethvargo (commented):
/hold

I'm not convinced this changes anything. Let's try to repro first.

mikehelmick (approved):

mikehelmick (approved):

whaught:
> /hold
> 
> I'm not convinced this changes anything. Let's try to repro first.

https://medium.com/@betable/3-go-gotchas-590b8c014e0a

I believe I fell into this trap! dereferencing an address to the local var of `range` winds up pointing to the wrong thing.

whaught:
/hold cancel

the new code is nicer looking anyhow

PR #1257, opened 2020-12-03
Put o11y context into middleware

whaught:
## Proposed Changes

* I believe what we're doing is errorprone - each request handler must remember to include build/realm info.
    * Instead the rate-limiter modifies the request context with the o11y BuildInfo
    * The AuthApp or session Realm also stores Realm.ID to the o11y context on the request

Now any handler can call `observability.Record` and be sure that info is there.
No weird pointer to context (not sure if that works as intended)


sethvargo:
A few things

1. I like this approach, but I think it should be a separate middleware (not limitware). I know, it's annoying and small, but it feels like a weird coupling, and we might have things that aren't rate limited in the future. I think it also needs to be higher in the middleware stack than ratelimiting.

2. Are you sure "The AuthApp or session Realm also stores Realm.ID to the o11y context on the request" is happening?

whaught:
> A few things
> 
> 1. I like this approach, but I think it should be a separate middleware (not limitware). I know, it's annoying and small, but it feels like a weird coupling, and we might have things that aren't rate limited in the future. I think it also needs to be higher in the middleware stack than ratelimiting.
> 2. Are you sure "The AuthApp or session Realm also stores Realm.ID to the o11y context on the request" is happening?

1. Shuffled these around to give the observability BuildId its own middleware.
2. Yes... something got lost in a merge resolve or something, but it's in there.

pkg/controller/issueapi/handle_issue_batch.go
```diff
@@ -37,15 +37,13 @@ func (c *Controller) HandleBatchIssue() http.Handler {
...
 		logger := logging.FromContext(ctx).Named("issueapi.HandleBatchIssue")
-		realm := controller.RealmFromContext(ctx)
 
 		resp := &api.BatchIssueCodeResponse{}
 		result := &issueResult{
 			httpCode:  http.StatusOK,
 			obsBlame:  observability.BlameNone,
 			obsResult: observability.ResultOK(),
 		}
-		ctx = observability.WithRealmID(observability.WithBuildInfo(ctx), realm.ID)
```
sethvargo, line 48: Hmm - I guess this would have paniced without a realm before?

pkg/controller/issueapi/handle_issue_batch.go
```diff
@@ -65,7 +63,7 @@ func (c *Controller) HandleBatchIssue() http.Handler {
 		}
 
 		// Add realm so that metrics are groupable on a per-realm basis.
-		if !realm.AllowBulkUpload {
+		if realm := controller.RealmFromContext(ctx); realm == nil || !realm.AllowBulkUpload {
```
sethvargo, line 66: What's the situation in which this can happen?

pkg/controller/middleware/apikey.go
```diff
@@ -35,8 +36,7 @@ const (
 )
 
 // RequireAPIKey reads the X-API-Key header and validates it is a real
-// authorized app. It also ensures currentAuthorizedApp is set in the template
-// map.
+// authorized app. It also ensures currentAuthorizedApp is set in the template map.
```
sethvargo, line 39:
```
and sets the realm id on the observability context
```

pkg/controller/middleware/apikey.go
```diff
@@ -109,6 +109,7 @@ func RequireAPIKey(cacher cache.Cacher, db *database.Database, h *render.Rendere
 			// Save the authorized app on the context.
 			ctx = controller.WithAuthorizedApp(ctx, &authApp)
 			ctx = controller.WithRealm(ctx, &realm)
```
sethvargo, line 111: Instead of having to remember to call `o11y.WithRealmID` after `WithRealm`, what if `WithRealm` automatically did that?

sethvargo (approved):
/lgtm

/hold

in case you want to respond to some of those comments

pkg/controller/issueapi/handle_issue_batch.go
whaught: there should never be nil here, but now that I've said that we've probably jinxed it.

whaught:
/unhold

took suggestions from review comments

sethvargo (approved):
/lgtm

PR #1259, opened 2020-12-03
Recreates Spanish i18n language file

rafaeltello:
## Proposed Changes

* Recreates Spanish i18n file with updated strings.
* Some translation comments 
    * Verified `Exposure Notifications` translation by switching iPhone to Spanish.
    * Translated `signing keys` as `llaves firmantes` this time, after plenty of feedback in a Spanish-speaking technology forum.
    * Spanish uses the same word for `realm` and `kingdom`, which is probably not a good translation. `Ámbito` is frequently used as a translation for realm in Spain, other possibilities could be `dominio` or `universo`, but each one has its own ambiguity problems.
    * `Short backup code` is a bit hard to translate; I focused more on the backup part (`código de respaldo`)

**Release Note**
```release-note
NONE
```

### DIFF
--- a/internal/i18n/locales/es/default.po
+++ b/internal/i18n/locales/es/default.po
@@ -1,0 +1,142 @@
+msgid ""
+msgstr ""
+"Language: es\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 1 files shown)


sethvargo (approved):
This is awesome, thank you so much!

PR #1269, opened 2020-12-04
Only check spelling in Go and HTML files

sethvargo:
This is required because the incoming language files fail spellcheck.

**Release Note**

```release-note
NONE
```


Makefile
```diff
@@ -50,7 +50,7 @@ copyrightcheck:
 
 spellcheck:
 	@command -v misspell > /dev/null 2>&1 || go get github.com/client9/misspell/cmd/misspell
-	@misspell -locale="US" -error -source="text" **/*
+	@misspell -locale="US" -error -source="text" $(GO_FILES) $(HTML_FILES)
```
mikehelmick, line 53: Can we hit markdown too?

sethvargo: Done

PR #1273, opened 2020-12-04
Nit: Show batch limit in error message

whaught:
## Proposed Changes

* Improving an error string to show the violated limit


sethvargo (approved):
/lgtm

PR #1274, opened 2020-12-04
Add portuguese language translations

arturenault:
## Proposed Changes

* Adding Portuguese language translations.
* Small concern - ISO 639-1 doesn't differentiate between Brazilian and European Portuguese. I did Brazilian Portuguese, but I suspect there would be some differences in these phrases.

**Release Note**

```release-note
Added Portuguese translations.
```

### DIFF
--- a/internal/i18n/locales/pt/default.po
+++ b/internal/i18n/locales/pt/default.po
@@ -64,7 +64,7 @@ msgid "codes.issue.header"
 msgstr "Criar código de verificação"
 
 msgid "codes.issue.instructions"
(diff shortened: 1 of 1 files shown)


mikehelmick:
/lgtm

Thanks for the contribution.

arturenault:
Sorry for the noise - realizing I'd like to do some more research as my COVID-related vocabulary is rusty :) will un-draft when I'm more confident.

sethvargo (approved):
/lgtm

PR #1277, opened 2020-12-05
Use the new batch API for the bulk uploader

whaught:
## Proposed Changes

* Uses the new batch-issue API for bulk code issaunce
    * Batch size is 10 which should 10x the throttle on requests
* Start at 1 instead of 0 for line - non CS people probably start there.

**Release Note**

```release-note
Move the Bulk Issue Codes UI to use the batch API
```


sethvargo (approved):
/lgtm

Can we move this into application.js as a follow-up? Looks like the only Go-template dependency is the csrf token, and we already pull that in js from the meta tag.

Issue #1201, opened 2020-11-26
Unify date formatting with project-level functions


sethvargo:
### TL;DR
We have a decent amount of date manipulation. We should centralize it for consistency and testability.

See also: https://github.com/google/exposure-notifications-verification-server/pull/1198#discussion_r530767952

PR #1278, opened 2020-12-05
Unify common date types into project

sethvargo:
Note that this does NOT do anything with dates in HTML, since those
dates tend to be formatted for style, not because of an API dependency.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1201

**Release Note**

```release-note
Unify common date types into project helpers
```


whaught (approved):
I like the  name `RFC3339Squish`

PR #1285, opened 2020-12-07
Install batch-issue in adminAPI and integration

whaught:
## Proposed Changes

* Install the batch-issue API in the adminAPI and integration
* Move the UI "show" bulk issue to package 'codes' with the rest of the  GET requests

**Release Note**

```release-note
Install batch-issue into the adminapi server
```


sethvargo (approved):
/lgtm

PR #1286, opened 2020-12-07
Add CSV and JSON exports for user stats

sethvargo:
This moves to the pattern the realm has where stats are exposed via JSON and the chart just consumes that JSON. I also added help text and clarity to describe what the chart does (and does not) show.

**Release Note**

```release-note
Add CSV and JSON exports for user stats
```

### DIFF
--- a/cmd/server/assets/users/show.html
+++ b/cmd/server/assets/users/show.html
@@ -17,16 +17,17 @@
     {{template "flash" .}}
 
     <h1>{{$user.Name}}</h1>
-    <p class="float-right">
-      <a href="/realm/users/{{$user.ID}}/edit">Edit</a>
-    </p>
-    <p>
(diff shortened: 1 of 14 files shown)


cmd/server/assets/users/show.html
```diff
@@ -37,6 +38,15 @@ <h6 class="card-title">Email</h6>
           {{$user.Email}}
         </div>
 
+        <h6 class="card-title">Password</h6>
+        <div class="mb-3 mt-n2">
+          <a href="/realm/users/{{$user.ID}}/reset-password"
+            data-method="POST"
+            data-confirm="Are you sure you want to reset this user's password?">
```
whaught, line 45: thanks for the `data-confirm`

whaught (approved):

PR #1289, opened 2020-12-07
Update API key stats too (CSV and JSON)

sethvargo:
**Release Note**

```release-note
Provide JSON and CSV output for API key stats
```

### DIFF
--- a/pkg/controller/apikey/stats.go
+++ b/pkg/controller/apikey/stats.go
@@ -16,22 +16,28 @@ package apikey
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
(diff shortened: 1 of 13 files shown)


pkg/controller/apikey/stats.go
```diff
@@ -0,0 +1,111 @@
...
+		logger := logging.FromContext(ctx).Named("apikey.HandleStats")
+
+		session := controller.SessionFromContext(ctx)
+		if session == nil {
+			controller.MissingSession(w, r, c.h)
+			return
+		}
+
+		realm := controller.RealmFromContext(ctx)
+		if realm == nil {
```
whaught, line 49: Do we need this realm and user nil check? Should be middleware enforced

sethvargo: Yea, but we always do it. If the middleware is accidentally excluded, it'll panic

whaught (approved):

PR #1290, opened 2020-12-07
Turkish (tr) translation

ahmetb:
Some strings such as

    msgid "codes.issue.countdown-expires-in"

could use %s formatting to ensure value can be appropriately placed in the
translated string.

If there's a Turkish person working on the project, peer reviews are
appreciated.


internal/i18n/locales/tr/default.po
```diff
@@ -0,0 +1,142 @@
+msgid ""
+msgstr ""
+"Language: en\n"
```
sethvargo, line 3: This needs to be `tr`

ahmetb: Done.

PR #1302, opened 2020-12-09
Initial config for Cloud Armor

yegle:
Enabled predefined rules:
 - XSS
 - Remote Code Execution
 - SQL Injection

All of them are in preview mode so no actual traffic will be blocked.
    
Also added instruction on how to use them, as well as example configs for common use cases.


sethvargo (approved):
/lgtm

PR #1303, opened 2020-12-09
Add Japanese (ja) locale

yuryu:
## Proposed Changes

* Add a Japanese (ja) translation
To make the service accessible to Japanese-speakers.

**Release Note**

```release-note
- Japanese (ja) translation
```

### DIFF
--- a/internal/i18n/locales/ja/default.po
+++ b/internal/i18n/locales/ja/default.po
@@ -1,0 +1,142 @@
+msgid ""
+msgstr ""
+"Language: ja\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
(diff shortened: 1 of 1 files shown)


internal/i18n/locales/ja/default.po
```diff
@@ -0,0 +1,142 @@
...
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=1; plural=0;\n"
+
+#
+# navigation menus
+# ----------------
+
+msgid "nav.issue-code"
+msgstr "コードを発行""
```
mikehelmick, line 14: seems like there might be an extra `"` here

yuryu: thanks, fixed

mikehelmick (approved):

sethvargo:
Thanks Emma!

PR #1306, opened 2020-12-09
Double default rate-limit tokens

whaught:
## Proposed Changes

* Doubles the default number of tokens-per minute to 120

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -189,7 +189,7 @@ export CACHE_REDIS_HOST="${google_redis_instance.cache.host}"
 export CACHE_REDIS_PORT="${google_redis_instance.cache.port}"
 
 export RATE_LIMIT_TYPE="REDIS"
-export RATE_LIMIT_TOKENS="120"
+export RATE_LIMIT_TOKENS="60"
 export RATE_LIMIT_INTERVAL="1m"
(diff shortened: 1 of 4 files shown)


mikehelmick:
this is a blanket increase - we don't want to increase the rate of /api/verify requests

just increase the rate on admin api issue requests

whaught:
> this is a blanket increase - we don't want to increase the rate of /api/verify requests
> 
> just increase the rate on admin api issue requests

Ah I misunderstood, I thought you were advocating a blanket increase.

sethvargo:
120 is fine, but change this for _only_ the `adminapi` in the terraform envvars.

terraform/main.tf
```diff
@@ -189,7 +189,7 @@ export CACHE_REDIS_HOST="${google_redis_instance.cache.host}"
 export CACHE_REDIS_PORT="${google_redis_instance.cache.port}"
 
 export RATE_LIMIT_TYPE="REDIS"
-export RATE_LIMIT_TOKENS="60"
+export RATE_LIMIT_TOKENS="120"
```
mikehelmick: this still applies to all services though

mikehelmick (commented):
/lgtm

whaught:
/close

This change was made directly to the deployed services, we aren't going to check-in a new default

Issue #1299, opened 2020-12-09
Correct universal links for iOS 


whaught:
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

Ensure our links conform the to spec format. We have a report that  the json file format of `/.well-known/apple-app-site-association` is incorrect

/assign

whaught:
Note: our implementation conforms to the docs shown here: https://developer.apple.com/documentation/safariservices/supporting_associated_domains
I'm a little skeptical about this archived one - don't want to break existing things.

The report notes existing support of major sites such as:
https://google.com/apple-app-site-association
which follows the former format.

mikehelmick:
Is this resolved now (in 18.2?)

whaught:
Should be
/close

PR #1307, opened 2020-12-09
Switch to older associated domain format

whaught:
Fixes #https://github.com/google/exposure-notifications-verification-server/issues/1299

## Proposed Changes

* Use the older format for iOS Associated Domains
/hold to see if we can get confirmation from Apple on this (docs are confusing)

**Release Note**

```release-note
Switch to older Associated Domains format for iOS
```

### DIFF
--- a/docs/link_handling.md
+++ b/docs/link_handling.md
@@ -48,15 +48,8 @@ Sample iOS universal link metadata
 	"applinks": {
 		"details": [
 			{
-				"appIDs": [
-					"com.google.test.application"
-				],
-				"components": [
-					{
-						"/": "/*",
-						"comment": "handle all urls"
-					}
-				]
+				"appID": "com.google.test.application",
+				"paths": ["*"]
 			}
 		]
 	}
--- a/pkg/controller/associated/ios.go
+++ b/pkg/controller/associated/ios.go
@@ -37,7 +37,7 @@ type Applinks struct {
 }
 
 type Detail struct {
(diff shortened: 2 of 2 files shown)


mikehelmick (approved):

docs/link_handling.md
```diff
@@ -48,15 +48,8 @@ Sample iOS universal link metadata
 	"applinks": {
 		"details": [
```
wescarr, line 49: According to the docs the following key must exist:
```json
"apps": []
```
From the docs
> The apps key in an apple-app-site-association file must be present and its value must be an empty array
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

docs/link_handling.md
```diff
@@ -48,15 +48,8 @@ Sample iOS universal link metadata
...
-				"appIDs": [
-					"com.google.test.application"
-				],
-				"components": [
-					{
-						"/": "/*",
-						"comment": "handle all urls"
-					}
-				]
+				"appID": "com.google.test.application",
```
wescarr, line 51: The `appID` is missing the Team ID prefix:
> The value of the appID key is the team ID or app ID prefix, followed by the bundle ID

You may want update the mobile app form to include a field for the Team ID, or mention that it should be included in the Application ID

wescarr (commented):
Just noticed 2 potential issues. Thanks for the quick turn around

Issue #1310, opened 2020-12-10
Pass through Twilio 4xx errors to the caller


sethvargo:
Twilio errors are currently server-side 500s. For 4xx errors, we should pass those onto the caller. For 5xx issues, we should still 5xx.

/assign

PR #1313, opened 2020-12-10
Return 400 on Twilio errors

sethvargo:
These are nearly always client errors

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1310

**Release Note**

```release-note
Return 400 (instead of 500) on Twilio errors
```

### DIFF
--- a/pkg/controller/issueapi/logic.go
+++ b/pkg/controller/issueapi/logic.go
@@ -265,13 +265,13 @@ func (c *Controller) issue(ctx context.Context, request *api.IssueCodeRequest) (
 				}
 
 				logger.Errorw("failed to send sms", "error", err)
-				result.obsBlame = observability.BlameServer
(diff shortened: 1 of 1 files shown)


sethvargo:
/assign @whaught 

I thought about parsing the code, but there are over 10k possible codes, and only 14 of them are internal server errors: https://www.twilio.com/docs/api/errors

whaught (approved):
Well this works.

I wonder if we want to go the other way and select a few eg. 503 unknown to keep as Server error

whaught:
> /assign @whaught 
> 
> 
> 
> I thought about parsing the code, but there are over 10k possible codes, and only 14 of them are internal server errors: https://www.twilio.com/docs/api/errors

Just picking one at random: error code 90002: "too many errors" error. This list is a little ridiculous

PR #1319, opened 2020-12-10
Update alerts.tf

mariliamelo:
Fix SlowBurn alert trigger conditions to AND, so the alert will only fire when both 30 minutes and 6 hours alerts are firing.


yuriatgoogle (approved):
/lgtm

sethvargo (approved):
/lgtm

PR #1325, opened 2020-12-11
Add Multi-SMS template columns to realm db

whaught:

## Proposed Changes

* Add fields for alternate SMS templates + labels
    * (original field will be 'default')
* Add validation for all templates
* Add a unit test for Realm validation

```release-note
Enable postgres hstore. Add fields to realm to store multiple SMS templates and add validation for them.
```


pkg/database/migrations.go
```diff
@@ -1801,6 +1801,35 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00075-AddAlternateSMSTemplates",
+			Migrate: func(tx *gorm.DB) error {
+				sqls := []string{
+					"ALTER TABLE realms ADD COLUMN IF NOT EXISTS alternate_sms_templates text[]",
```
sethvargo, line 1808: We should probably use a single hstore column for the alternates instead

whaught: Nice. Postgres has everything.

pkg/database/migrations.go
```diff
@@ -1827,6 +1828,21 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00075-EnableExtension_hstore",
```
sethvargo, line 1832: 76

pkg/database/migrations.go
```diff
@@ -1827,6 +1828,21 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+			Migrate: func(tx *gorm.DB) error {
+				return tx.Exec("CREATE EXTENSION IF NOT EXISTS hstore").Error
+			},
+		},
+		{
+			ID: "00076-AddAlternateSMSTemplates",
```
sethvargo, line 1838: 77

and we should really write a test that migrations are incremental... (https://github.com/google/exposure-notifications-verification-server/issues/1343)

pkg/database/realm.go
```diff
@@ -358,28 +360,17 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
...
-		}
-		if strings.Contains(r.SMSTextTemplate, SMSLongCode) {
-			r.AddError("SMSTextTemplate", fmt.Sprintf("cannot contain %q - the long code is automatically included in %q", SMSLongCode, SMSENExpressLink))
-		}
-
-	} else {
-		// Check that we have exactly one of [code] or [longcode] as template substitutions.
-		if c, lc := strings.Contains(r.SMSTextTemplate, SMSCode), strings.Contains(r.SMSTextTemplate, SMSLongCode); !(c || lc) || (c && lc) {
-			r.AddError("SMSTextTemplate", fmt.Sprintf("must contain exactly one of %q or %q", SMSCode, SMSLongCode))
+	r.validateSMSTemplate(r.SMSTextTemplate)
```
sethvargo, line 363: This API is a bit weird since it doesn't return anything. I don't have a better solution though.

pkg/database/realm.go
```diff
@@ -358,28 +360,17 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
...
+	if r.SMSTextAlternateTemplates != nil {
+		for l, t := range r.SMSTextAlternateTemplates {
+			if t == nil {
```
sethvargo, line 366: We should check `t == nil || *t == ""`

sethvargo:
```
type Hstore map[string]*string
```

pkg/database/realm.go
```diff
@@ -434,6 +425,31 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 	return nil
 }
 
+func (r *Realm) validateSMSTemplate(t string) {
```
sethvargo, line 428: Could we comment this since it's a little weird of an API?

pkg/database/realm.go
```diff
@@ -434,6 +425,31 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
...
+	} else {
+		// Check that we have exactly one of [code] or [longcode] as template substitutions.
+		if c, lc := strings.Contains(t, SMSCode), strings.Contains(t, SMSLongCode); !(c || lc) || (c && lc) {
+			r.AddError("SMSTextTemplate", fmt.Sprintf("must contain exactly one of %q or %q", SMSCode, SMSLongCode))
+		}
+	}
+
+	// Check template length.
+	if l := len(t); l > SMSTemplateMaxLength {
+		r.AddError("SMSTextTemplate", fmt.Sprintf("must be %v characters or less, current message is %v characters long", SMSTemplateMaxLength, l))
```
sethvargo, line 449:
```suggestion
		r.AddError("SMSTextTemplate", fmt.Sprintf("must be %d characters or less, current message is %d characters long", SMSTemplateMaxLength, l))
```

pkg/database/migrations.go
```diff
@@ -1827,6 +1828,21 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
 				return nil
 			},
 		},
+		{
+			ID: "00076-EnableExtension_hstore",
```
mikehelmick, line 1832: can these not be combined?

seems plausible that hstore needs to be committed before it can be used, just curious.

whaught: I want to make sure the transaction doesn't fail (if it can't find hstore type before it is committed)

pkg/database/realm.go
```diff
@@ -434,6 +429,33 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 	return nil
 }
 
+// validateSMSTemplate is a helper method to validate a single SMSTemplate.
+// Errors are returned by appending them to the realm's Errorable fields.
+func (r *Realm) validateSMSTemplate(t string) {
+	if r.EnableENExpress {
+		if !strings.Contains(t, SMSENExpressLink) {
+			r.AddError("SMSTextTemplate", fmt.Sprintf("must contain %q", SMSENExpressLink))
```
mikehelmick, line 437: these errors aren't going to be right, since they are for SMSTextTemplate

validation is going to be tricky...

this is where I think breaking these out into a separate table so you had a stable ID that could be part of a form field ID would be valuable.

pkg/database/migrations.go
sethvargo: Yea, it should be separate as-is here.

pkg/database/realm.go
sethvargo: We should Take the field name as another param to this function. We could also prefix the error with the label name for alt templates.

That being said, we'll mark these as `required` in the HTML, so the only way someone gets to these errors is if they do some significant mucking about. I think it's probably fine to keep this as-is and create a follow-up issue to make it betterer in the future. WDYT?

sethvargo (approved):
/lgtm
/hold

pkg/database/realm.go
mikehelmick: we can also include the label in the error message - I'm worried about a user not being able to tell where the error is
and therefore not be able to fix it.

mikehelmick: I'm fine w/ doing the follow up when the UI is added.

mikehelmick (commented):
/lgtm

whaught:
/unhold

I hear you on errors - I was intending to solve that in the follow up with more complete UX

Issue #1296, opened 2020-12-08
increase test coverage in pkg/database


mikehelmick:
especially around data validation, i.e. `BeforeSave` methods

PR #1326, opened 2020-12-11
allow short code to be in ENX SMS messages / link still required

### LINKED ISSUE #1323: ENX SMS - Allow Short Code in expansion
### TL;DR

Allow ENX users to opt to send short code.

/assign

mikehelmick:
Fixes #1323 
Contributes towards #1296 

## Proposed Changes

* Allow the short code and short expiration to appear in the ENX test message, still requires the long link though. This is a backup in case the link isn't clickable (improper universal link registration)

**Release Note**

```release-note
Allow short code to be in the ENX SMS message.
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -369,20 +369,14 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 		if strings.Contains(r.SMSTextTemplate, SMSRegion) {
 			r.AddError("SMSTextTemplate", fmt.Sprintf("cannot contain %q - this is automatically included in %q", SMSRegion, SMSENExpressLink))
 		}
-		if strings.Contains(r.SMSTextTemplate, SMSCode) {
-			r.AddError("SMSTextTemplate", fmt.Sprintf("cannot contain %q - the long code is automatically included in %q", SMSCode, SMSENExpressLink))
-		}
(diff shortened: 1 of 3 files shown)


whaught:
/lgtm

Love the test coverage!
Also unit-test is mad about gomod

pkg/database/realm.go
```diff
@@ -369,20 +369,14 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 		if strings.Contains(r.SMSTextTemplate, SMSRegion) {
 			r.AddError("SMSTextTemplate", fmt.Sprintf("cannot contain %q - this is automatically included in %q", SMSRegion, SMSENExpressLink))
 		}
-		if strings.Contains(r.SMSTextTemplate, SMSCode) {
```
whaught, line 372: Do we want to keep these validations, but only if [enslink] is included?

mikehelmick: removing this is the intent of the PR

mikehelmick:
fixed go mod issue

pkg/database/realm.go
whaught: oh i see we might include both the link and the code as a backup

whaught (approved):

PR #1327, opened 2020-12-11
Update StackdriverExportFailed alert duration to 15min.

yegle:
The previous 5min of errors is too noisy.


mariliamelo:
/lgtm 

We still need @sethvargo or @mikehelmick approval to be able to merge.

PR #1335, opened 2020-12-12
Introduce RBAC

### LINKED ISSUE #1322: RBAC
### TL;DR
 
Implement finer grained RBAC

Some requested roles

* stats viewer
* user manager (but not admin)
* bulk issuer

/assign @sethvargo

sethvargo:
This introduces Role-Based Access Controls (RBAC) into the system. Common operations in the system are split into Read (view-only) and Write (create, update, delete). Users have 0 or more permissions on a realm through Memberships. Memberships replace the existing `user_realms` table (and obviates the `admin_realms` table). Since realm roles are no longer binary (previously user or admin), many of our existing UI elements no longer made sense. For example, we frequently displayed a "Realm Admin" pill in lists, but those have been removed since permissions are now multi-dimensional. There are two meta-permissions - LegacyRealmUser and LegacyRealmAdmin - which closely correspond to the existing primitive roles. The RBAC system also improves the event log, since individual permissions are now diffed.

The RBAC system has security properties that prevent privilege escalation. The system forbids creating users with permissions greater than your own, and it forbids changing your own permissions entirely. The only exception is system administrators, who are granted full realm permissions when joining a realm and have all permissions revoked when leaving a realm (via the system admin console). The UI and templates can conditionally assert a membership's priviledges and updates accordingly.

One of the biggest changes is conceptual - "Membership" is a first-class entity that must be handled and inspected in the system. The "current membership" defines the currently logged-in user and currently selected realm (if one exists). Additionally, RBAC assertion is now at the controller-level instead of the routing layer. I don't love it, but injecting it into the routing layer proved quite challenging and brittle.

Finally, the system was designed to support more than just "users". If we ever wanted more granular permissions on, say API keys, we could easily add that.

Other miscellaneous changes include:

- Removing some totally dead code
- Changing the response code on controller.NotFound to be 404 instead of 401
- Rendering a real 401 page instead of always forcing a sign-out
- Better redirects after editing some resources

List of future enhancements (I'll file issues after we reach consensus on this PR):

- Display human descriptions of each permission in the UI

Fixes #1322

**Release Note**

```release-note
**Major change** Introduce Role-Based Access Controls (RBAC) to replace legacy user/admin roles. Existing users will retain their existing permissions, but new users will be able to have more granular permissions. This change involves an **irreversible database migration** and should be planned accordingly. We recommend system operators put the servers into maintenance mode before applying these migrations.
```


cmd/server/assets/admin/realms/edit.html
```diff
@@ -23,20 +24,7 @@ <h1>Edit {{$realm.Name}}</h1>
     </p>
 
     <div class="card mb-3 shadow-sm">
-      <div class="card-header">
```
whaught, line 26: no more link to realm-admin settings?

whaught: mm nevermind. realm-admin as a thing is complex now.

pkg/controller/middleware/auth.go
```diff
@@ -29,11 +29,10 @@ import (
...
 )
 
-// RequireAuth requires a user to be logged in. It also ensures that currentUser
-// is set in the template map. It fetches a user from the session and stores the
-// full record in the request context.
+// RequireAuth requires a user to be logged in. It also fetches and stores
+// information about the user on the request context.
 func RequireAuth(cacher cache.Cacher, authProvider auth.Provider, db *database.Database, h *render.Renderer, sessionIdleTTL, expiryCheckTTL time.Duration) mux.MiddlewareFunc {
-	cacheTTL := 15 * time.Minute
+	cacheTTL := 30 * time.Minute
```
whaught, line 35: just wondering any particular reason for the cacheTTL doubling?

whaught:
- standardize on 'currentRealm' over realm
- remove needless log lines from DB
- indentation fixes on html
- html padding / coloring / back buttons
- real error pages
- dead .currentRealms
- remove trailing spaces

This PR has a great set of nits fixed - that said it might have been easier to review if we split into 'functional' and 'nits' PRs. LGTM... but very possible I've missed something.

/hold for when you're ready.
/lgtm

cmd/server/assets/admin/realms/edit.html
sethvargo: Yea, the permissions being non-binary makes some of our "helpful admin things" slightly less helpful.

pkg/controller/middleware/auth.go
sethvargo: I set it to 0 during testing, then changed it back. I apparently changed it back to the wrong value, but we could realistically cache this for 10000 days and it wouldn't matter because the gorm callbacks clear the cache when the key is changed or revoked.

sethvargo:
> This PR has a great set of nits fixed - that said it might have been easier to review if we split into 'functional' and 'nits' PRs. LGTM... but very possible I've missed something.

Yea, sorry 😦. The HTML and trailing spaces is my editor and I didn't see that in the diff, sorry. 

I'm gonna wait for @mikehelmick to give a thumbs up before unholding.

pkg/database/migrations.go
```diff
@@ -236,29 +217,19 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
 				for _, u := range users {
-					logger.Debugw("adding user to default realm", "user", u.ID)
-
-					u.AddRealm(defaultRealm)
-					if u.SystemAdmin {
-						u.AddRealmAdmin(defaultRealm)
-					}
-
-					if err := tx.Save(u).Error; err != nil {
+					if err := u.AddToRealm(db, defaultRealm, rbac.LegacyRealmAdmin, System); err != nil {
```
mikehelmick, line 228: This migration is semantically different than it was before

cmd/server/assets/apikeys/index.html
```diff
@@ -2,6 +2,12 @@
 
 {{$authApp := .authApp}}
 
+{{$currentMembership := .currentMembership}}
+{{$canWrite := false}}
```
mikehelmick, line 6: `{{$canWrite := .currentMembership.Can rbac.APIKeyWrite}}`

simplify?

cmd/server/assets/apikeys/index.html
```diff
@@ -14,13 +20,15 @@
...
         <span class="oi oi-key mr-2 ml-n1" aria-hidden="true"></span>
         API keys
-        <a href="/realm/apikeys/new" class="float-right mr-n1 text-secondary" data-toggle="tooltip" title="New API key">
-          <span class="oi oi-plus small" aria-hidden="true"></span>
-        </a>
+        {{if $canWrite}}
+          <a href="/realm/apikeys/new" class="float-right mr-n1 text-secondary" data-toggle="tooltip" title="New API key">
+            <span class="oi oi-plus small" aria-hidden="true"></span>
+          </a>
+        {{end}}
```
mikehelmick, line 31: do we want to say that the person doesn't have ability to create API keys to avoid confusion - w/a. tooltip maybe?

sethvargo: I actually had something like this originally, but then I looked across a bunch of systems and read a few blog posts that recommend against telling people they lack permissions.

I did update the realm guide to specifically say that "if you don't see this button, you don't have permission".

cmd/server/assets/apikeys/index.html
sethvargo: For some reason, that wasn't working, but I tried again and it's working. /shrug.

pkg/database/migrations.go
```diff
@@ -1801,16 +1723,112 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+					return err
+				}
+
+				// Migrate existing admin_realms permissions into user_realms with the
+				// legacy realm admin permission.
+				if err := tx.Exec(
+					`INSERT INTO user_realms(user_id, realm_id, permissions)
+						SELECT user_id, realm_id, $1 FROM admin_realms
+						ON CONFLICT (user_id, realm_id) DO UPDATE SET permissions = $1`,
+					int64(rbac.LegacyRealmAdmin),
```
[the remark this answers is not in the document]
sethvargo: The new permissions would be `LegacyRealmAdmin | LegacyRealmUser`, but LegacyRealmAdmin includes all the same permissions as User. I'll add User to make it more explicit though.

pkg/database/migrations.go
sethvargo: Done

pkg/database/migrations.go
```diff
@@ -1801,16 +1723,112 @@ func (db *Database) getMigrations(ctx context.Context) *gormigrate.Gormigrate {
...
+
+				// Drop admin_realms.
+				if err := tx.Exec(`DROP TABLE IF EXISTS admin_realms`).Error; err != nil {
+					return err
+				}
+
+				return nil
+			},
+			Rollback: func(tx *gorm.DB) error {
+				// No rollback for this
```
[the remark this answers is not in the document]
sethvargo: Done.

sethvargo:
@mikehelmick @whaught here's a [diff of just the changes since you last reviewed](https://github.com/google/exposure-notifications-verification-server/pull/1335/files/3bfaac15bdde474c6d4856cf2486824bf1262a4d..f7db0fe10cbda8dbb6af148491e5f146c7c12f88).

cmd/server/assets/apikeys/index.html
mikehelmick: sg

mikehelmick (approved):

PR #1337, opened 2020-12-13
Change rate calculating window of StackdriverExportFailed alert.

yegle:
The rate window should be larger than the interval of the errors.
Currently we export to stackdriver every 2min, meaning if the export is
constantly failing, our calculated error rate with 1min window will have
the rate oscillating between 0 and 1, and thus we would never get an
alert with our alerting duration (each time the value reaches 0 the
timer to trigger the alert is reset).

Changing this to 5min window means the condition is "on" as soon as
there's a single export error and last at least 5min. The alert is
firing if the condition is "on" for >15min.


whaught (approved):

mariliamelo:
Thanks for the PR Yuchen. This makes sense. You just forgot to update the comment in code. Consider adding this explanation there when you update it.

Issue #1343, opened 2020-12-14
Write a test that migration numbers are strictly increasing


sethvargo:
/assign

PR #1349, opened 2020-12-14
Add test for ensuring migration numbers are up and to the right

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1343

**Release Note**

```release-note
Add test for ensuring database migration numbers are strictly increasing
```

### DIFF
--- a/pkg/database/database_util.go
+++ b/pkg/database/database_util.go
@@ -343,7 +343,7 @@ func runMigrations(db *Database) error {
 	}()
 
 	// Run migrations.
-	if err := db.RunMigrations(ctx); err != nil {
+	if err := db.MigrateTo(ctx, "", false); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/database/database_util.go
```diff
@@ -343,7 +343,7 @@ func runMigrations(db *Database) error {
 	}()
 
 	// Run migrations.
-	if err := db.RunMigrations(ctx); err != nil {
+	if err := db.MigrateTo(ctx, "", false); err != nil {
```
whaught, line 346: Aren't these equivalent? Why the change

sethvargo: Yea, that's why I dropped `.RunMigrations` 😄

sethvargo: We had two functions that did the same thing.

whaught (approved):

PR #1352, opened 2020-12-15
Allow the issuer to select an SMS template

whaught:

## Proposed Changes

* If more than one SMS template is defined, allow the user to pick one
* Add a field to the issue API and compose with the given template

**Release Note**

```release-note
Allow user to select an SMS template on code-issue. Add template label field to issueAPI.
```

### DIFF
--- a/cmd/server/assets/codes/issue.html
+++ b/cmd/server/assets/codes/issue.html
@@ -249,6 +249,7 @@
     let $form;
       let $inputTestDate;
       let $inputSymptomDate;
+      let $inputSMSTemplate;
       let $inputPhone;
       let $buttonSubmit;
       let $buttonReset;
@@ -271,6 +272,7 @@
       $form = $('form#issue');
         $inputTestDate = $('input#test-date');
         $inputSymptomDate = $('input#symptom-date');
+        $inputSMSTemplate = $('select#sms-template');
         $inputPhone = $('input#phone');
         $buttonSubmit = $('button#submit');
         $buttonReset = $('button#reset');
@@ -322,6 +324,7 @@
         data.tzOffset = new Date().getTimezoneOffset();
 
         {{if $hasSMSConfig}}
+        data['SMSTemplateLabel'] = $inputSMSTemplate.val();
         data['phone'] = iti.getNumber();
         {{end}}
 
--- a/cmd/server/assets/codes/issue.html
+++ b/cmd/server/assets/codes/issue.html
@@ -128,7 +128,7 @@
               <label for="sms-template" class="col-sm-6 col-md-4 col-lg-3">{{t $.locale "codes.issue.sms-template-label"}}</label>
               <div class="col-sm-6 col-md-8 col-lg-9">
                 <div class="input-group">
-                  <select class="form-control" name="sms-template">
+                  <select class="form-control" id="sms-template">
                     <option value="Default SMS template">Default SMS template</option>
                     {{range $k, $v := $currentRealm.SMSTextAlternateTemplates}}
                     <option value="{{$k}}">{{$k}}</option>
--- a/cmd/server/assets/codes/issue.html
+++ b/cmd/server/assets/codes/issue.html
@@ -324,7 +324,7 @@
         data.tzOffset = new Date().getTimezoneOffset();
 
         {{if $hasSMSConfig}}
-        data['SMSTemplateLabel'] = $inputSMSTemplate.val();
+        data['smsTemplateLabel'] = $inputSMSTemplate.val();
         data['phone'] = iti.getNumber();
         {{end}}
 
--- a/cmd/server/assets/codes/issue.html
+++ b/cmd/server/assets/codes/issue.html
@@ -123,8 +123,26 @@
         <div class="card mb-3 shadow-sm">
           <div class="card-header">{{t $.locale "codes.issue.sms-text-message-header"}}</div>
           <div class="card-body">
+            {{if $currentRealm.SMSTextAlternateTemplates}}
             <div class="row form-group">
-              <label for="symptomDate" class="col-sm-6 col-md-4 col-lg-3">{{t $.locale "codes.issue.sms-text-message-label"}}</label>
+              <label for="sms-template" class="col-sm-6 col-md-4 col-lg-3">{{t $.locale "codes.issue.sms-template-label"}}</label>
+              <div class="col-sm-6 col-md-8 col-lg-9">
+                <div class="input-group">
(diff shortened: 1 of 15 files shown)


cmd/server/assets/codes/issue.html
```diff
@@ -123,8 +123,26 @@ <h1>{{t $.locale "codes.issue.header"}}</h1>
...
+            {{if $currentRealm.SMSTextAlternateTemplates}}
             <div class="row form-group">
-              <label for="symptomDate" class="col-sm-6 col-md-4 col-lg-3">{{t $.locale "codes.issue.sms-text-message-label"}}</label>
+              <label for="sms-template" class="col-sm-6 col-md-4 col-lg-3">{{t $.locale "codes.issue.sms-template-label"}}</label>
+              <div class="col-sm-6 col-md-8 col-lg-9">
+                <div class="input-group">
+                  <select class="form-control" name="sms-template">
+                    <option value="Default SMS template">Default SMS template</option>
+                    {{range $k, $v := $currentRealm.SMSTextAlternateTemplates}}
+                    <option value="{{$k}}">{{$k}}</option>
```
sethvargo, line 134: Nit: indent

sethvargo, line 134: Should this be `{{$v}}`? If not, why is `$v` scoped?

pkg/api/api.go
```diff
@@ -186,8 +186,9 @@ type IssueCodeRequest struct {
 	TestType    string `json:"testType"`
 	// Offset in minutes of the user's timezone. Positive, negative, 0, or omitted
 	// (using the default of 0) are all valid. 0 is considered to be UTC.
-	TZOffset float32 `json:"tzOffset"`
-	Phone    string  `json:"phone"`
+	TZOffset         float32 `json:"tzOffset"`
+	Phone            string  `json:"phone"`
+	SMSTemplateLabel string  `json:"SMSTemplateLabel"`
```
sethvargo, line 191: This should be `smsTemplateLabel` (I don't make the rules)

pkg/controller/issueapi/logic.go
```diff
@@ -260,7 +260,11 @@ func (c *Controller) issue(ctx context.Context, authApp *database.AuthorizedApp,
 		if err := func() error {
 			defer observability.RecordLatency(ctx, time.Now(), mSMSLatencyMs, &result.obsBlame, &result.obsResult)
 
-			message := realm.BuildSMSText(code, longCode, c.config.GetENXRedirectDomain())
+			message, err := realm.BuildSMSText(code, longCode, c.config.GetENXRedirectDomain(), request.SMSTemplateLabel)
+			if err != nil {
+				logger.Warnw("failed to compose SMS", "error", err)
```
sethvargo, line 265: Just return err here, it's logged later

pkg/database/realm_test.go
```diff
@@ -34,14 +34,14 @@ func TestSMS(t *testing.T) {
 	realm.SMSTextTemplate = "This is your Exposure Notifications Verification code: [enslink] Expires in [longexpires] hours"
 	realm.RegionCode = "US-WA"
 
-	got := realm.BuildSMSText("12345678", "abcdefgh12345678", "en.express")
+	got, _ := realm.BuildSMSText("12345678", "abcdefgh12345678", "en.express", "")
```
sethvargo, line 37: Check error here and below

internal/i18n/locales/en/default.po
```diff
@@ -102,6 +102,12 @@ msgstr "SMS text message (recommended)"
 msgid "codes.issue.sms-text-message-label"
 msgstr "Patient phone number"
 
+msgid "codes.issue.sms-template-label"
```
sethvargo, line 105: I _think_ you need to translate all the things now or the tests will fail.

sethvargo (changes requested):

cmd/server/assets/codes/issue.html
whaught: templates can't do `$k, _` and single var is just values. I need just the keys (labels)

pkg/api/api.go
whaught: yea that's consistent with tzOffset. Weird though.

cmd/server/assets/codes/issue.html
sethvargo: but then can do `$k := range`

whaught: but `$k := range .someMap` iterates the map values, not the keys. Maybe I'm misunderstanding

internal/i18n/locales/de/default.po
```diff
@@ -102,6 +102,12 @@ msgstr "SMS Textnachricht (empfohlen)"
 msgid "codes.issue.sms-text-message-label"
 msgstr "Patienten Telefonnummer"
 
+msgid "codes.issue.sms-template-label"
+msgstr "SMS Vorlage"
+
+msgid "codes.issue.sms-template-detail"
+msgstr "Der Patient erhält eine SMS mit der ausgewählten Nachrichtenvorlage."
```
whaught, line 109: I don't know enough german to refute that `ausgewählten Nachrichtenvorlage` is `selected template` but it sure looks long

internal/i18n/locales/en/default.po
whaught: Added Google translate translations.

internal/i18n/locales/de/default.po
sethvargo: @hostirosti could you sanity check these two new German translations please?

hostirosti: LGTM

cmd/server/assets/codes/issue.html
sethvargo: My b - apparently this is backwards in Go templates from the Go language /tableflip

docs/api.md
```diff
@@ -207,6 +208,10 @@ Request a verification code to be issued. Accepts [optional] symptom date and te
 * `phone`
   * Phone number to send the SMS to. If a phone number is provided, but the SMS text
     message fails to send, the API will return a 4xx client error.
+* `smsTemplateLabel`
+  * If the realm has more than one SMS template defined, this may be optionally specify
+    the label of the message template which the server should compose. If omitted, the
+    default template will be used.
```
sethvargo, line 214: What happens if they supply a label that does not exist?

sethvargo (approved):
/lgtm

@whaught can do you do a followup to update the docs on the failure modes?

PR #1353, opened 2020-12-15
Properly redirect to login when required

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1341

**Release Note**

```release-note
Properly redirect to login page after session expiration. Previously the user would get an "Unauthorized" page.
```

### DIFF
--- a/pkg/controller/flash/flash.go
+++ b/pkg/controller/flash/flash.go
@@ -85,6 +85,13 @@ func (f *Flash) Clear() {
 	delete(f.values, flashKeyWarning)
 }
 
+// Clone makes a copy of this flash data into the new target.
+func (f *Flash) Clone(values map[interface{}]interface{}) {
+	for k, v := range f.values {
+		values[k] = v
+	}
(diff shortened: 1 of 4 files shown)


pkg/controller/flash/flash.go
```diff
@@ -85,6 +85,13 @@ func (f *Flash) Clear() {
 	delete(f.values, flashKeyWarning)
 }
 
+// Clone makes a copy of this flash data into the new target.
+func (f *Flash) Clone(values map[interface{}]interface{}) {
```
mikehelmick, line 89: it feels weird that the target map is passed in

sethvargo: That's how all flashes are created: https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/controller/flash.go#L29

It's to avoid a circular dependency on session depending on flash depending on session.

mikehelmick (approved):

PR #1354, opened 2020-12-15
fix realm from firewall for apikeys

mikehelmick:
## Proposed Changes

* load realm correctly when using apikeys in the firewall module

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

PR #1355, opened 2020-12-15
Redirect back if bulk upload is not enabled

sethvargo:
This is better than unauthorized since it probably means the user just picked the wrong realm from selection (or bookmarked the page).

**Release Note**

```release-note
NONE
```

/assign @whaught

### DIFF
--- a/pkg/controller/codes/bulk_issue.go
+++ b/pkg/controller/codes/bulk_issue.go
@@ -26,6 +26,13 @@ func (c *Controller) HandleBulkIssue() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
+		session := controller.SessionFromContext(ctx)
+		if session == nil {
+			controller.MissingSession(w, r, c.h)
+			return
+		}
+		flash := controller.Flash(session)
+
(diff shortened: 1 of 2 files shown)


pkg/controller/codes/bulk_issue.go
```diff
@@ -39,7 +46,8 @@ func (c *Controller) HandleBulkIssue() http.Handler {
 		currentRealm := membership.Realm
 
 		if !currentRealm.AllowBulkUpload {
-			controller.Unauthorized(w, r, c.h)
+			flash.Error("That feature is not enabled for your realm!")
```
whaught, line 49: Nit

```suggestion
			flash.Error("Bulk upload is not enabled for your realm!")
```

whaught:
/lgtm
/hold

feel free to add lgtm back if you accept the nit

pkg/controller/codes/bulk_issue.go
sethvargo: I'm intentionally doing this so we don't leak the feature.

sethvargo:
@whaught PTAL

whaught (approved):

mikehelmick (approved):

PR #1367, opened 2020-12-15
Use a single var.project in en-alerting module.     

yegle:
Previously we use two variables (monitoring-host-project, verification-server-project) to follow the best practice. But we didn't really prioritize the work to set up a monitoring host project.

Simplify it by having a single var.project.


sethvargo (approved):
/lgtm

PR #1368, opened 2020-12-15
Document error codes table for IssueResponse

whaught:
## Proposed Changes

* This shows the possible error codes for Issue. Many of them are shared with Verify, but they are slightly different
    * We could dedup with a 'common error codes' table and then specific ones, but I suspect that would be confusing to readers who probably scroll to the specific API they are integrating.


sethvargo (approved):
/lgtm

Issue #1350, opened 2020-12-14
Consider deprecating the bulk upload API


mariliamelo:
### TL;DR
We should consider deprecating the bulk upload API and adjust the rate limiting implementation.

### Design

**Proposal**
Deprecate the bulk upload API and improve the rate limiting system.

**Alternatives considered**
Maintain the bulk upload functionality and better document the possible errors and work around failures from other services. 

**Additional information**
The bulk upload feature although useful can be very confusing for one implementing it, since there can be partial success due to dependence on SMS infrastructure.

sethvargo:
Another option is to make the API internal-only (only available on the UI server).

whaught:
I think I noted when creating  this API that it does not conform to Google's standard https://google.aip.dev/233 which does not allow for partial success/failure and requires the entire batch be atomic. Unfortunately that is not possible given that /issue both creates a resource and sends an SMS.

However batch does allow our own bulk-uploader to perform better without messing with throttling or running from multiple IPs. It also allows for server-side parallelization (not implemented) which could create higher overall throughput.

We should definitely throw up some warnings in its documentation if it remains.

whaught:
/close

We have in-the-wild usage of this and have made changes to improve performance and documentation to try to mitigate the onboarding experience. I don't think deprecating is an option (or easy)

PR #1369, opened 2020-12-15
Clearer outer error message for batch API

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1350

## Proposed Changes

* Switch the outer error message of the batch-issue API to describe success/fail and how to view individual errors
    * Should be clearer and guide developers to not trust the outer message and dive into individual failures. Especially if they don't see the docs.

**Release Note**

```release-note
Clearer outer error message for batch issue API
```

### DIFF
--- a/pkg/controller/issueapi/issue_batch.go
+++ b/pkg/controller/issueapi/issue_batch.go
@@ -16,13 +16,14 @@ package issueapi
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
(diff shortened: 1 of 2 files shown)


pkg/controller/issueapi/issue_batch.go
```diff
@@ -92,21 +93,28 @@ func (c *Controller) HandleBatchIssue() http.Handler {
 				// continue processing if when a single code issuance fails.
```
mariliamelo, line 93: minor, but this commend should read "continue processing *even* when... " ?

mariliamelo (approved):

PR #1371, opened 2020-12-15
add another scrubbing pattern

mikehelmick:
## Proposed Changes

* add `The 'To' number %s is not a valid phone number.` to scrub pattern

**Release Note**

```release-note
Add another twilio error message to scrub pattern
```

### DIFF
--- a/pkg/controller/issueapi/logic.go
+++ b/pkg/controller/issueapi/logic.go
@@ -43,6 +43,10 @@ var (
 			prefix: "phone number: ",
 			suffix: ", ",
 		},
+		{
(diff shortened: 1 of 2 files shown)


whaught:
I'm a little worried we'll never really be sure we scrubbed them all (Twilio has a ton of error types). I wonder if we should eat the errors and replace with our own, scrub all digits, or something else that requires us to allow-list novel errors

mikehelmick:
I didn't want to scrub the "From" number as it is helpful in debugging if
we need to locate who is sending bad messages.

PR #1379, opened 2020-12-16
Remember user's SMS template preference

whaught:

## Proposed Changes

* Remember the issuing user's last used SMS template. 

**Release Note**

```release-note
 Remember the issuing user's last used SMS template.
```


cmd/server/assets/codes/bulk-issue.html
```diff
@@ -98,7 +99,7 @@
               <select class="form-control" id="sms-template">
                 <option value="Default SMS template">Default SMS template</option>
                 {{range $k, $v := $currentRealm.SMSTextAlternateTemplates}}
-                <option value="{{$k}}">{{$k}}</option>
+                <option value="{{$k}}" {{if eq $k $currentUser.DefaultSMSTemplateLabel}}selected{{end}}>{{$k}}</option>
```
sethvargo, line 102: The `selectedIf` helper makes this easier (it's in render.go), but should just be `selectedIf eq ($currentUser.DefaultSMSTemplate $k)`

cmd/server/assets/codes/issue.html
```diff
@@ -131,7 +132,7 @@ <h1>{{t $.locale "codes.issue.header"}}</h1>
                   <select class="form-control" id="sms-template">
                     <option value="Default SMS template">Default SMS template</option>
                     {{range $k, $v := $currentRealm.SMSTextAlternateTemplates}}
-                    <option value="{{$k}}">{{$k}}</option>
+                    <option value="{{$k}}" {{if eq $k $currentUser.DefaultSMSTemplateLabel}}selected{{end}}>{{$k}}</option>
```
sethvargo, line 135: Same

pkg/controller/issueapi/logic.go
```diff
@@ -301,6 +301,14 @@ func (c *Controller) issue(ctx context.Context, authApp *database.AuthorizedApp,
 				return err
 			}
 
+			if request.SMSTemplateLabel != "" && currentUser != nil &&
```
sethvargo, line 304: I feel like "remember my last template" should be entirely in the UI. I don't think we should do this in the core logic. So maybe we move it to the controller method that's _only_ for the UI?

pkg/database/migrations.go
```diff
@@ -1842,6 +1842,38 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 				return tx.Exec(`ALTER TABLE realms DROP COLUMN IF EXISTS alternate_sms_templates`).Error
 			},
 		},
+		{
+			ID: "00078-AddUserDefaultSMSTemplate",
+			Migrate: func(tx *gorm.DB) error {
+				sqls := []string{
+					`ALTER TABLE users ADD COLUMN IF NOT EXISTS remember_last_used_sms_template BOOL`,
+					`UPDATE users SET remember_last_used_sms_template = TRUE WHERE remember_last_used_sms_template IS NULL`,
+					`ALTER TABLE users ALTER COLUMN remember_last_used_sms_template SET DEFAULT TRUE`,
```
sethvargo, line 1851: Should we even make this configurable? I feel like the system should just remember (no bool field to check)

pkg/database/migrations.go
```diff
@@ -1842,6 +1842,38 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+					`ALTER TABLE users ALTER COLUMN remember_last_used_sms_template SET NOT NULL`,
+					`ALTER TABLE users ADD COLUMN IF NOT EXISTS default_sms_template_label VARCHAR(255)`,
```
sethvargo, line 1853: We need to add some validation or a database constraint here. Either:

- Changing an SMS label should fail if a user is depending on it
- Changing an SMS label clears the default for any users

pkg/database/user.go
```diff
@@ -43,6 +43,9 @@ type User struct {
 	Name        string `gorm:"type:varchar(100)"`
 	SystemAdmin bool   `gorm:"column:system_admin; default:false;"`
 
+	RememberLastUsedSMSTemplate bool   `gorm:"type:boolean; not null; default:true;"`
```
sethvargo, line 46: I'm not convinced we need this column or check

pkg/database/user.go
```diff
@@ -43,6 +43,9 @@ type User struct {
...
+	DefaultSMSTemplateLabel     string `gorm:"type:varchar(255);"`
```
sethvargo, line 47: A few things here:

- A user can be a member of multiple realms (with different templates), so this might be better as a property on the `Membership` instead of the user
- There needs to be some validation that the template is one of the keys in the hstore for the realm
- There needs to be some validation/changes when the hstore column is updated to cascade here

sethvargo (changes requested):

whaught:
/hold

this needs some tests too

pkg/database/user.go
whaught: I don't think we need to keep the membership preferred template in sync with the existing ones. When the page loads and the template doesn't exist, that selectedIf doesn't match anything and the user goes back to the default template in the selector.

mikehelmick: how does this work for users that in in more than 1 realm?

whaught: This uses Seth's suggestion now - it's attached to Membership

mikehelmick: so I have to join a club?

whaught: Club Realm

pkg/database/membership.go
```diff
@@ -31,9 +33,59 @@ type Membership struct {
...
+
+	if actor == nil {
+		return fmt.Errorf("auditing actor is nil")
+	}
+
+	return db.db.Transaction(func(tx *gorm.DB) error {
+		var audits []*AuditEntry
+
+		var existing Membership
+		if err := tx.
```
mikehelmick, line 55: should we be reading this for update?

whaught: We do that to diff existing and add audits. This is the pattern we're using in SaveRealm, SaveUser, etc

pkg/database/user.go
sethvargo: ™️

pkg/database/membership.go
```diff
@@ -59,3 +111,146 @@ func (m *Membership) Can(p rbac.Permission) bool {
 	}
 	return rbac.Can(m.Permissions, p)
 }
+
+// ListMemberships lists the memberships for this user. Use
+// ListMembershipsCached where possible.
+func (u *User) ListMemberships(db *Database) ([]*Membership, error) {
```
sethvargo, line 117: Hmm... I find it weird that the user functions are in membership.go. I feel like the functions should be in the file for their pointer receiver.

sethvargo (approved):
/lgtm

PR #1382, opened 2020-12-16
Do not return 500s from redirect service for missing realms

### LINKED ISSUE #1380: ENX Redirect service returns 500 for unknown host name
### TL;DR

This should be a 404 instead of a 500.

sethvargo:
Fixes #1380

**Release Note**

```release-note
Do not return 500s from redirect service for missing realms
```

/assign @mikehelmick 
/assign @flagxor

### DIFF
--- a/cmd/enx-redirect/main.go
+++ b/cmd/enx-redirect/main.go
@@ -18,24 +18,18 @@ import (
 	"context"
 	"crypto/sha1"
 	"fmt"
-	"net/http"
 	"os"
 	"strconv"
 
+	"github.com/google/exposure-notifications-verification-server/internal/routes"
 	"github.com/google/exposure-notifications-verification-server/pkg/buildinfo"
 	"github.com/google/exposure-notifications-verification-server/pkg/cache"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
-	"github.com/google/exposure-notifications-verification-server/pkg/controller"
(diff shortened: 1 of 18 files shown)


cmd/enx-redirect/main.go
```diff
@@ -18,24 +18,18 @@ import (
 	"context"
```
sethvargo, line 18: This moved to internal (like server routes) so we can use it in tests

cmd/server/assets/400.html
```diff
@@ -7,9 +7,9 @@
 
 <body>
   <main role="main" class="container mt-5">
-    <h1>Not found</h1>
```
sethvargo, line 10: I'm not sure why our 400 page said "not found", but fixed.

pkg/controller/redirect/helpers.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: These are moved into their own file so we can test them. The integration tests need an `_test` package name to prevent the import cycle.

pkg/controller/redirect/helpers_test.go
```diff
@@ -0,0 +1,361 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: idk why `git mv` never works

mikehelmick (approved):

PR #1391, opened 2020-12-16
Ignore .terraform.lock.hcl

chaodaiG:
This file was added in terraform 0.14, it causes terraform deployment failure due to uncommitted local changes

https://www.terraform.io/docs/configuration/dependency-lock.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,9 @@
 *.terraform/
 *.tfvars
 terraform/state.tf
+terraform/.terraform.lock.hcl
 terraform-e2e-ci/state.tf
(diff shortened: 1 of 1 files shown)


yegle:
This was supposed to be checked in. I think each of us has created a PR like this...

sethvargo:
Yea, it's technically supposed to be checked in:

> Terraform automatically creates or updates the dependency lock file each time you run the terraform init command. You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.

/close

PR #1400, opened 2020-12-16
Add RBAC check to issue_batch api

whaught:
## Proposed Changes

* We currently perform this check on the bulk-issue page rendering, but we also need to check the API

**Release Note**

```release-note
RBAC check for BulkIssue on the API
```

### DIFF
--- a/pkg/controller/issueapi/issue_batch.go
+++ b/pkg/controller/issueapi/issue_batch.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/api"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/observability"
+	"github.com/google/exposure-notifications-verification-server/pkg/rbac"
 )
 
 const maxBatchSize = 10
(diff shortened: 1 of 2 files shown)


pkg/controller/issueapi/issue_batch.go
```diff
@@ -71,6 +72,13 @@ func (c *Controller) HandleBatchIssue() http.Handler {
 			return
 		}
 
+		if !membership.Can(rbac.CodeBulkIssue) {
```
sethvargo, line 75: I think we can only do this check if it was called via the UI. There is no "membership" on the API.

I'm also pretty sure I did this - isn't there a specific handler for the UI server to server this?

sethvargo (changes requested):

pkg/controller/issueapi/issue_batch.go
whaught: No the UI server remounts HandleIssue and HandleBulkIssue at another route (for ajax), but it's the same code.

What if we only check this if User is present (not authorized_app)

sethvargo: I guess that's fine? I just really dislike when a code path behaves dramatically different based on some other data that's a few layers of indirection away.

pkg/controller/issueapi/issue_batch.go
```diff
@@ -72,7 +72,7 @@ func (c *Controller) HandleBatchIssue() http.Handler {
 			return
 		}
 
-		if !membership.Can(rbac.CodeBulkIssue) {
+		if membership.User != nil && !membership.Can(rbac.CodeBulkIssue) {
```
sethvargo, line 75: I think you need to check if _membership_ is nil. Membership.User will never be nil.

whaught: Yep sorry. Users have RBAC, not AuthApps.

sethvargo (approved):
/lgtm

PR #1405, opened 2020-12-17
Add descriptions for rbac permissions

### LINKED ISSUE #1387: Add descriptions for RBAC roles
Right now, the UI only displays names. We should add descriptions to the enum fields and then extract them into a map to display in the UI.

**Alternatives**

- Just hard code them, but that seems bad.

/assign

sethvargo:
Fixes #1387

Some sacrifices are made, including dropping stringer because it doesn't support multiple `iotas` in a single const.

**Release Note**

```release-note
Add descriptions for RBAC permissions
```


pkg/rbac/rbac_test.go
```diff
@@ -28,18 +28,18 @@ func TestRBAC_Permissions(t *testing.T) {
 	}{
 		{AuditRead, 2},
 		{APIKeyRead, 4},
-		{APIKeyWrite, 8},
+		{APIKeyWrite, 12},
```
sethvargo, line 31: This will technically break anyone using old permissions, but this code has never been in a release.

mikehelmick (approved):

mikehelmick (approved):

PR #1406, opened 2020-12-17
Refactoring issue codes

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1440
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1439

## Proposed Changes

* Separate concerns of code-issuance into parts
    1) the outer API shell that handles receiving the request and rendering
    2) issue one or many in a batch
        3) validate and populate a VerificationCode
        4) generate a random code and commit to the DB
* Use go routines to parallelize SMS send for batch
* Adds lots of test coverage

**Release Note**

```release-note
Increased test coverage for /issue and /batch-issue APIs
Parallel execution for batch
```


pkg/controller/issueapi/handle_issue.go
```diff
@@ -0,0 +1,76 @@
...
+// HandleIssue responds to the /issue API for issuing verification codes
+func (c *Controller) HandleIssue() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if result := c.handleIssueFn(w, r); result != nil {
+			recordObservability(r.Context(), result)
+		}
+	})
+}
+
+func (c *Controller) handleIssueFn(w http.ResponseWriter, r *http.Request) *issueResult {
```
sethvargo, line 35: I'm also not sure I see the value in pulling this out into its own function as-written. It just makes it harder to test because it's package private, but it's the same signature as the public function.

I was hoping we could do something more like:

```go
func (c *Controller) HandleIssueAPI() http.Handler {
    ctx := r.Context()

    realm := controller.RealmFromContext(ctx)
    if realm == nil {
      controller.MissingRealm(w, r, c.h)
      return
    }

    authApp := controller.AuthorizedAppFromContext(ctx)
    if authApp == nil {
      controller.MissingAuthorizedApp(w, r, c.h)
      return
    }

    if authApp.Type != database.APIKeyTypeAdmin {
      controller.Unauthorized(w, r, c.h)
      return
    }

    if err := c.handleIssue(ctx, realm, authApp, nil); err != nil {
      // build and return response
    }
	})
}

func (c *Controller) HandleIssueUI() http.Handler {
    ctx := r.Context()

    membership := controller.MembershipFromContext(ctx)
    if membership == nil {
      controller.MissingMembership(w, r, c.h)
      return
    }
    realm := membership.Realm

    if !membership.Can(rbac.CodeIssue) {
      controller.Unauthorized(w, r, c.h)
      return
    }

    if err := c.handleIssue(ctx, realm, nil, membership); err != nil {
      // build and return response
    }
	})
}

func (c *Controller) handleIssue(ctx context.Context, realm *database.Realm, authApp *database.AuthorizedApp, membership *database.Membership) *issueResult {
  // ...
}
```

Then you've separated rbac handling and http stuff out of the core logic.

pkg/controller/issueapi/handle_issue.go
```diff
@@ -0,0 +1,76 @@
...
+		return result
+	}
+
+	authApp, membership, realm, err := c.getAuthorizationFromContext(ctx)
+	if err != nil {
+		result.obsResult = observability.ResultError("MISSING_AUTHORIZED_APP")
+		c.h.RenderJSON(w, http.StatusUnauthorized, api.Error(err))
+		return result
+	}
+
```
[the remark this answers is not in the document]
sethvargo: Or see my code sample above.

pkg/controller/issueapi/handle_issue_batch.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Ensure bulk upload is enabled on this realm.
+	if !realm.AllowBulkUpload {
+		result.obsResult = observability.ResultError("BULK_ISSUE_NOT_ENABLED")
+		c.h.RenderJSON(w, http.StatusBadRequest, api.Errorf("bulk issuing is not enabled on this realm"))
+		return result
+	}
+
+	if membership != nil && !membership.Can(rbac.CodeBulkIssue) {
+		result.obsResult = observability.ResultError("BULK_ISSUE_NOT_ENABLED")
```
sethvargo, line 74: I'm not sure this is the right result

pkg/controller/issueapi/parse_date.go
```diff
@@ -50,6 +56,41 @@ func init() {
 	}
 }
 
+func (c *Controller) parseDate(d string, tzOffset int, parseSettings *dateParseSettings) (*time.Time, *issueResult) {
```
sethvargo, line 59: I don't think this should be coupled with the issueResult. Just return `*time.Time, error` and handle the rest upstream.

pkg/controller/issueapi/validate_code.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/google/exposure-notifications-verification-server/internal/project"
+	"github.com/google/exposure-notifications-verification-server/pkg/api"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/observability"
+	"github.com/google/exposure-notifications-verification-server/pkg/sms"
+)
+
+// populateCode populates and validates a code from an issue request.
```
sethvargo, line 32: Maybe `buildVerificationCode` could be a better name?

pkg/controller/issueapi/validate_code.go
```diff
@@ -0,0 +1,148 @@
...
+		} else if code != nil {
+			return nil, &issueResult{
+				obsResult:   observability.ResultError("UUID_CONFLICT"),
+				httpCode:    http.StatusConflict,
+				errorReturn: api.Errorf("code for %s already exists", request.UUID).WithCode(api.ErrUUIDAlreadyExists),
+			}
+		}
+	}
+
+	vCode.Code = "placeholder"
```
sethvargo, line 121: Is this because you're about to call `Validate` and it'll fail otherwise?

pkg/controller/issueapi/validate_code.go
```diff
@@ -0,0 +1,148 @@
...
+	vCode.LongCode = "placeholder"
+	if err := vCode.Validate(realm); err != nil {
```
sethvargo, line 123: I think we should drop this and move everything into `BeforeSave` instead.

pkg/controller/issueapi/validate_code.go
```diff
@@ -0,0 +1,148 @@
...
+	if request.Phone == "" || smsProvider == nil {
+		// If this isn't going to be send via SMS, make the long code expiration time same as short.
+		// This is because the long code will never be shown or sent.
+		vCode.LongExpiresAt = vCode.ExpiresAt
+	}
+
+	// If there is a client-provided UUID, check if a code has already been issued.
+	// this prevents us from consuming quota on conflict.
+	if vCode.UUID != "" {
+		if code, err := realm.FindVerificationCodeByUUID(c.db, request.UUID); err != nil {
```
sethvargo, line 104: This is currently using the untrimmed UUID. It's also a weird mismatch between the model value and request. See my other comment.

pkg/controller/issueapi/validate_code.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if request.Phone == "" || smsProvider == nil {
+		// If this isn't going to be send via SMS, make the long code expiration time same as short.
+		// This is because the long code will never be shown or sent.
+		vCode.LongExpiresAt = vCode.ExpiresAt
+	}
+
+	// If there is a client-provided UUID, check if a code has already been issued.
+	// this prevents us from consuming quota on conflict.
+	if vCode.UUID != "" {
```
sethvargo, line 103: Style nit: I'd prefer if we _didn't_ set UUID on the vcode above and instead set it here. This is a little unclear because you're calling `vCode.UUID`, but you're actually asserting the `request.UUID`. I'd prefer:

```go
if uuid := project.TrimSpaceNotPrintable(request.UUID); uuid != "" {
  if _, err := realm.FindVerificationCodeByUUID(c.db, uuid); err != nil {
    if database.IsNotFound(err) {
      vcode.UUID = uuid
    }
    return err
  }

  // no error, meaning record with UUID exists, conflict
  return "UUID_CONFLICT"
}
```

pkg/controller/json_bind.go
```diff
@@ -35,6 +35,9 @@ func BindJSON(w http.ResponseWriter, r *http.Request, data interface{}) error {
 	if !IsJSONContentType(r) {
 		return fmt.Errorf("content-type is not application/json")
 	}
+	if r.Body == nil {
+		return fmt.Errorf("no body set")
```
sethvargo, line 39: I think there might be cases where it's valid to have an empty json body

pkg/database/vercode.go
```diff
@@ -323,8 +316,8 @@ func (db *Database) ExpireCode(uuid string) (*VerificationCode, error) {
 
 // SaveVerificationCode created or updates a verification code in the database.
 // Max age represents the maximum age of the test date [optional] in the record.
-func (db *Database) SaveVerificationCode(vc *VerificationCode, maxAge time.Duration) error {
-	if err := vc.Validate(maxAge); err != nil {
+func (db *Database) SaveVerificationCode(vc *VerificationCode, realm *Realm) error {
```
sethvargo, line 319: I'm wondering if `SaveVerificationCode` should dangle off the realm instead? A lot of its validation is tied to the realm, and it does have a FK relationship to the realm.

pkg/database/vercode_test.go
```diff
@@ -73,12 +73,12 @@ func TestVerificationCode_FindVerificationCode(t *testing.T) {
 	t.Parallel()
 
 	db, _ := testDatabaseInstance.NewDatabase(t, nil)
+	realm := NewRealmWithDefaults("Test Realm")
```
sethvargo, line 76: Avoid creating a realm - the migrations already create one. 

```go
realm, err := db.FindRealm(1)
if err != nil {
  t.Fatal(err)
}
```

sethvargo: (Throughout all tests)

pkg/database/vercode_test.go
```diff
@@ -288,26 +287,6 @@ func TestVerCodeValidate(t *testing.T) {
 			},
 			Err: ErrInvalidTestType,
 		},
-		{
```
sethvargo, line 291: I'm a little hesitant about this validation moving into the controller.

sethvargo (changes requested):

pkg/database/vercode_test.go
whaught: At the least it isn't duplicated. The date validation relies on config values for maxAge. It's a little odd to have those kinds of dependencies in the model.

We could go all-in and try to get all of the validation, deps and all, in there but this gets us to a good spot for now?

pkg/controller/json_bind.go
whaught: moving a line around. The defer r.Body.Close() panics. `defer` is tricky

pkg/controller/issueapi/validate_code.go
whaught: Yea otherwise it's 'code too short'. It looks kinda ugly though.
If we go further with pushing validation to the model we might be able to get rid of some of these weirdnesses.

pkg/controller/issueapi/gen_code.go
```diff
@@ -0,0 +1,187 @@
...
+		if err != nil {
+			logger.Errorw("failed to take from limiter", "error", err)
+			return &issueResult{
+				obsResult:   observability.ResultError("FAILED_TO_TAKE_FROM_LIMITER"),
+				httpCode:    http.StatusInternalServerError,
+				errorReturn: api.Errorf("failed to verify realm stats, please try again").WithCode(api.ErrInternal),
+			}
+		}
+
+		stats.Record(ctx, mRealmTokenUsed.M(1))
```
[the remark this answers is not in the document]
whaught: I think we should move this to after the !ok, but not after commitCode. This metric is trying to log limiter tokens used - if commitCode fails we have still consumed quota

pkg/controller/issueapi/gen_code.go
```diff
@@ -0,0 +1,187 @@
...
+				return &issueResult{
+					obsResult:   observability.ResultError("QUOTA_EXCEEDED"),
+					httpCode:    http.StatusTooManyRequests,
+					errorReturn: api.Errorf("exceeded realm quota, please contact a realm administrator").WithCode(api.ErrQuotaExceeded),
+				}
+			}
+		}
+	}
+
+	if err := c.commitCode(ctx, vCode, realm, c.config.GetCollisionRetryCount()); err != nil {
```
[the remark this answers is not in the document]
whaught: I think the retrys were implemented to prevent shortcode collision, not uuid.

pkg/controller/issueapi/handle_issue.go
whaught: it pulls out http.HandlerFunc so I can test without real mux routing.
The real reason I did this was to `recordObservability` here because the variable capture of `defer` was confusing me

whaught: But I like your idea too, makes it less weird to be serving this at two routes

pkg/controller/issueapi/gen_code.go
```diff
@@ -0,0 +1,185 @@
...
+	b, err := retry.NewConstant(50 * time.Millisecond)
+	if err != nil {
+		return fmt.Errorf("failed to create backoff: %w", err)
+	}
+
+	retry.Do(ctx, retry.WithMaxRetries(uint64(retryCount), b), func(ctx context.Context) error {
+		var code string
+		code, err = GenerateCode(realm.CodeLength)
+		if err != nil {
+			logger.Errorf("code generation error: %v", err)
```
sethvargo, line 111: Can we remove these log messages or let the error propagate up? Else, we should use `logger.Errorw`

pkg/controller/issueapi/gen_code.go
```diff
@@ -0,0 +1,185 @@
...
+			return err
+		}
+		longCode := code
+		if realm.LongCodeLength > 0 {
+			longCode, err = GenerateAlphanumericCode(realm.LongCodeLength)
+			if err != nil {
+				logger.Errorf("long code generation error: %v", err)
```
sethvargo, line 118: Same

pkg/controller/issueapi/handle_issue.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (c *Controller) IssueWithUIAuth(w http.ResponseWriter, r *http.Request) *IssueResult {
+	ctx := r.Context()
+	result := &IssueResult{
+		HTTPCode:  http.StatusOK,
+		obsResult: observability.ResultOK(),
+	}
+
+	if membership := controller.MembershipFromContext(ctx); membership == nil || !membership.Can(rbac.CodeIssue) {
```
sethvargo, line 85: If the UI Auth was called, the middleware should have set the membreship, so a missing membership should be an internal error. But a lacking permission should still be unauthorized.

pkg/controller/issueapi/validate_code_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	if err := db.SaveVerificationCode(existingCode, realm); err != nil {
+		t.Fatal(err)
+	}
+
+	authApp := &database.AuthorizedApp{
+		Model: gorm.Model{ID: 123},
+	}
+	ctx = controller.WithAuthorizedApp(ctx, authApp)
+	membership := &database.Membership{UserID: 456}
```
sethvargo, line 63: Nit: I prefer to embed this in the WithMembership cal

pkg/controller/json_bind.go
```diff
@@ -36,8 +36,8 @@ func BindJSON(w http.ResponseWriter, r *http.Request, data interface{}) error {
 		return fmt.Errorf("content-type is not application/json")
 	}
 
-	defer r.Body.Close()
```
sethvargo, line 39: 🛑 I think this is introducing a bug. We have to close the main body, not the limited reader.

pkg/controller/middleware/membership.go
```diff
@@ -83,6 +83,7 @@ func LoadCurrentMembership(cacher cache.Cacher, db *database.Database, h render.
 
 			// Save the membership and realm on the context.
 			ctx = controller.WithMembership(ctx, membership)
+			ctx = controller.WithRealm(ctx, membership.Realm)
```
sethvargo, line 86: I intentionally didn't set this because I don't want to pull the realm from the context. It should always be pulled from the membershpi.

pkg/database/migrations.go
```diff
@@ -28,7 +28,8 @@ import (
 )
 
 const initState = "00000-Init"
-const VercodeUUIDUniqueIndex = "idx_vercode_uuid_unique"
+const VerCodesLongCodeUniqueIndex = "uix_verification_codes_code"
+const VerCodesCodeUniqueIndex = "uix_verification_codes_long_code"
```
sethvargo, line 32: The current index name is `uix_verification_codes_realm_code`

pkg/database/migrations.go
```diff
@@ -28,7 +28,8 @@ import (
 )
 
 const initState = "00000-Init"
-const VercodeUUIDUniqueIndex = "idx_vercode_uuid_unique"
+const VerCodesLongCodeUniqueIndex = "uix_verification_codes_code"
```
sethvargo, line 31: The current index name is `uix_verification_codes_realm_long_code`

pkg/database/migrations.go
sethvargo: Also, I think this is swapped with the one above it

pkg/database/migrations.go
```diff
@@ -1505,8 +1506,8 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					"ALTER TABLE verification_codes ALTER COLUMN long_code DROP NOT NULL",
 					"CREATE UNIQUE INDEX IF NOT EXISTS uix_verification_codes_realm_code ON verification_codes (realm_id,code) WHERE code != ''",
 					"CREATE UNIQUE INDEX IF NOT EXISTS uix_verification_codes_realm_long_code ON verification_codes (realm_id,long_code) WHERE long_code != ''",
-					"DROP INDEX IF EXISTS uix_verification_codes_code",
-					"DROP INDEX IF EXISTS uix_verification_codes_long_code",
+					fmt.Sprintf("DROP INDEX IF EXISTS %s", VerCodesLongCodeUniqueIndex),
+					fmt.Sprintf("DROP INDEX IF EXISTS %s", VerCodesLongCodeUniqueIndex),
```
sethvargo, line 1510: One of these should be `VerCodesCodeUniqueIndex`

sethvargo (changes requested):

pkg/controller/middleware/membership.go
whaught: I don't think there is a membership for API based calls.
https://github.com/google/exposure-notifications-verification-server/blob/8a910cfa82ee86fa54762a8228d339bbb44876e4/pkg/controller/middleware/apikey.go#L110

If we want to remove WithRealm and always take either membership or authapp and re-look it up, we could. If so I'd rather also remove controller.WithRealm entirely

pkg/controller/issueapi/gen_code.go
whaught: yea looks like there's an Errorw one up from this. Let's leave it at that.

sethvargo (approved):
/lgtm

PR #1410, opened 2020-12-17
Mount Apache logger in binary

sethvargo:
We don't want these appearing in tests.

**Release Note**

```release-note
NONE
```


sethvargo (commented):
/assign @mikehelmick

mikehelmick (approved):

PR #1411, opened 2020-12-17
Add implied permissions - write permissions need read permissions to be useful

### LINKED ISSUE #1408: RBAC Write permissions should include Read
RBAC Write permissions should include Read. For example, granting APIKeyWrite should also grant APIKeyRead (or fail and require the user to select both).

mikehelmick:
Fixes #1408 

## Proposed Changes

* All write permissions automatically add required read permissions if they are not set.

**Release Note**

```release-note
Setting write permissions automatically add required read permissions.
```

### DIFF
--- a/pkg/controller/user/update_test.go
+++ b/pkg/controller/user/update_test.go
@@ -63,9 +63,14 @@ func TestUpdate(t *testing.T) {
 
 	for _, permission := range rbac.NamePermissionMap {
 		permission := permission
-		target := fmt.Sprintf(`input#permission-%d`, permission)
+		targets := []string{fmt.Sprintf(`input#permission-%d`, permission)}
+		// We also ned to remove permissions that imply this permission, or it will be added back in.
+		for _, superPerm := range rbac.ImpliedBy(permission) {
+			targets = append(targets, fmt.Sprintf(`input#permission-%d`, superPerm))
+		}
 
-		if err := chromedp.Run(taskCtx,
+		// Build the actions as an array prior to run since super-permissions actions aren't known until runtime.
+		actions := []chromedp.Action{
 			// Pre-authenticate the user.
 			browser.SetCookie(cookie),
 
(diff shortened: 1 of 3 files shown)


pkg/controller/user/update_test.go
```diff
@@ -63,9 +63,14 @@ func TestUpdate(t *testing.T) {
 
 	for _, permission := range rbac.NamePermissionMap {
 		permission := permission
-		target := fmt.Sprintf(`input#permission-%d`, permission)
+		targets := []string{fmt.Sprintf(`input#permission-%d`, permission)}
+		// We also ned to remove permissions that imply this permission, or it will be added back in.
```
whaught, line 67:
```suggestion
		// We also need to remove permissions that imply this permission, or it will be added back in.
```

pkg/rbac/rbac.go
```diff
@@ -145,6 +160,39 @@ const (
...
+		MobileAppWrite: {MobileAppRead},
+		UserWrite:      {UserRead},
+	}
+
+	// This is the inverse of the above map, set by the init() func.
+	// Done in code to ensure it always stays in sync with requiredPermission.
+	impliedBy = make(map[Permission][]Permission)
+)
+
+// Note: there are multiple init functions in this file. They are organized to be
```
whaught, line 182: This is very clever and require/implied will always be in-sync.

but would it be easier to read as:
`requiredPermission = map[Permission][]Permission{
		APIKeyRead:    {APIKeyWrite},
		SettingsRead:  {SettingsWrite},
		MobileAppRead: {MobileAppWrite},
		UserRead:      {UserWrite},
	}`

and perhaps keep init() in but instead just compare them and `Panic`  if they don't match up. That would also allow for them to be asymmetrical eg. 
`APIKeyWrite:    {APIKeyRead, SuperPowers},`
and
`APIKeyRead:    {APIKeyWrite},`

whaught: I realize I'm advocating both duplication and panic, but multiple-init is also weird

sethvargo: Multi-init is pretty common tbh

sethvargo (approved):
/lgtm

I'll cleanup the spelling in another PR (gonna add some javascript that uses this)

pkg/rbac/rbac.go
mikehelmick: This actually allows for the asymmetry - we just don't have it yet.

PR #1412, opened 2020-12-17
Factor out test server config from the test server

whaught:
## Proposed Changes

* This breaks the  envstest into a second package.
    1 ) The existing package  starts the full server and all of  its routes for testing
    2) A sub-package that just gets test configuration for a server
* This breaks a circular dependency where envstest depends on all routes. Later I will have unit tests that take the server config and just test an individual controller


sethvargo:
> This breaks a circular dependency where envstest depends on all routes. Later I will have unit tests that take the server config and just test an individual controller

The easier way around this is to create tests in package `thing_test` instead of `thing`.

internal/envstest/testconfig/random.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package envstest
```
sethvargo, line 15: I think random should still be in the `envstest` package, no?

whaught:
> > This breaks a circular dependency where envstest depends on all routes. Later I will have unit tests that take the server config and just test an individual controller
> 
> The easier way around this is to create tests in package `thing_test` instead of `thing`.

That's what the chrome tests do, I'm trying to create unit tests that keep functions private-scope

sethvargo (changes requested):
I'm -1 on this change. For controllers that need these things, it's pretty easy to do this yourself. You shouldn't need the routes if you're testing a single controller method (e.g. https://github.com/google/exposure-notifications-verification-server/blob/505f73248b74e168d92a9763ac04ad514b4d5ed1/pkg/controller/associated/index_test.go#L32)

internal/envstest/testconfig/random.go
whaught: random is only called to seed random HMACs in the serverconfig

sethvargo: I think it's called in integration.go too, but that's a different build tag, so you might not see the compilation error?

Either way, I think envstest.Random() is better than testconfig.Random()

whaught: envstest.NewServer is the only external reference. I've moved random() to package scope for good measure

PR #1413, opened 2020-12-17
Also includes 3xx responses as part of the good SLI signal.

yegle:
Services like enx-redirect responds a lot 3xx.


yegle:
cc @yuriatgoogle

mikehelmick (approved):

PR #1414, opened 2020-12-17
Create a no-op failure SMS provider

whaught:
## Proposed Changes

* Create a no-op SMS provider that only throws an error
* This will be used later in unit test cases to assert what happens when an SMS fails
    * (Pulling out easily reviewable pieces from a larger refactor)


sethvargo (approved):
/lgtm

PR #1426, opened 2020-12-17
Adding an test for bulk-issue error handling

whaught:
## Proposed Changes

* Integration test looks at partial success / failure responses
* These code paths are shared with single-issue, so should be adding coverage

```release-note
correct response for invalid_test_type previously returned unsupported_test_type even if unparsable.

added integration test coverage for code issue
```

### DIFF
--- a/pkg/controller/issueapi/issue_batch_test.go
+++ b/pkg/controller/issueapi/issue_batch_test.go
@@ -34,14 +34,14 @@ func TestIssueBatch(t *testing.T) {
 	testSuite := testsuite.NewIntegrationSuite(t, ctx)
 	adminClient, err := testSuite.NewAdminAPIClient(ctx, t)
 	if err != nil {
-		t.Fatalf("failed to create admin API client, err: %v", err)
+		t.Fatal(err)
 	}
 	db := testSuite.DB
 	realm := testSuite.Realm
 
 	realm.AllowedTestTypes = database.TestTypeConfirmed
 	if err := db.SaveRealm(realm, database.SystemTest); err != nil {
-		t.Fatalf("error saving realm: %v", err)
+		t.Fatal(err)
 	}
 
 	smsConfig := &database.SMSConfig{
--- a/pkg/controller/issueapi/issue_batch_test.go
+++ b/pkg/controller/issueapi/issue_batch_test.go
@@ -1,0 +1,208 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package issueapi_test
+
+import (
+	"context"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/google/exposure-notifications-verification-server/internal/project"
+	"github.com/google/exposure-notifications-verification-server/pkg/api"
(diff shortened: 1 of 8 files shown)


pkg/controller/issueapi/issue_batch_test.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+func TestIssueBatch(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	testSuite := testsuite.NewIntegrationSuite(t, ctx)
+	adminClient, err := testSuite.NewAdminAPIClient(ctx, t)
+	if err != nil {
+		t.Fatalf("failed to create admin API client, err: %v", err)
```
sethvargo, line 37: Just a note that I tend to find added context to error messages in tests unhelpful. The test output will include the exact error line, so `t.Fatal(err)` is easier for errors. The opposite, however, for assertions.

pkg/controller/issueapi/logic.go
```diff
@@ -82,24 +82,24 @@ func (c *Controller) issue(ctx context.Context, authApp *database.AuthorizedApp,
...
+	// Verify the test type
+	request.TestType = strings.ToLower(request.TestType)
+	if _, ok := c.validTestType[request.TestType]; !ok {
 		return &issueResult{
 			obsBlame:    observability.BlameClient,
-			obsResult:   observability.ResultError("UNSUPPORTED_TEST_TYPE"),
+			obsResult:   observability.ResultError("INVALID_TEST_TYPE"),
 			httpCode:    http.StatusBadRequest,
-			errorReturn: api.Errorf("unsupported test type: %v", request.TestType).WithCode(api.ErrUnsupportedTestType),
+			errorReturn: api.Errorf("invalid test type").WithCode(api.ErrInvalidTestType),
```
sethvargo, line 92: Hmmm - this feels like an API change. Is that intended?

pkg/testsuite/clients.go
```diff
@@ -36,38 +36,39 @@ type AdminClient struct {
 }
 
 // BatchIssueCode calls the issue-batch API call.
-func (c *AdminClient) BatchIssueCode(req api.BatchIssueCodeRequest) (*api.BatchIssueCodeResponse, error) {
+func (c *AdminClient) BatchIssueCode(req api.BatchIssueCodeRequest) (int, *api.BatchIssueCodeResponse, error) {
```
sethvargo, line 39: Document returned int value

sethvargo, line 39: Thinking out loud... should we return a different type here?

```go
type ImBadAtNaming struct {
  Successes int
  Failures int
  Responses []*api.IssueCodeResponse
```

pkg/testsuite/integration.go
```diff
@@ -222,8 +222,8 @@ func NewIntegrationSuite(tb testing.TB, ctx context.Context) *IntegrationSuite {
 
 	return &IntegrationSuite{
 		cfg:       cfg,
-		db:        db,
-		realm:     realm,
+		DB:        db,
```
sethvargo, line 225: Why make these public?

pkg/controller/issueapi/issue_batch_test.go
whaught: removed

pkg/controller/issueapi/logic.go
whaught: Yea I think is what was intended / documented

sethvargo: Ack - let's call this out in the changelog. I have paranoia about people depending on broken behaviors given this project's history.

pkg/testsuite/integration.go
whaught: My test uses the DB and Realm to set up some preconditions for the test.

Specifically it creates an SMSConfig and preloads a VerificationCode to test UUID conflict.

pkg/testsuite/clients.go
whaught: we could, but I want to return relatively raw data here because I want to validate and traverse it as an external caller might

PR #1428, opened 2020-12-17
Use envvar to increase stackdriver timeout to 1min.

yegle:


sethvargo (approved):
/lgtm

I like the idea of keeping the o11y_config block, even if it's empty though.

PR #1431, opened 2020-12-18
Part 1: refactor issue-code

whaught:
## Proposed Changes

* This encapsulates the issue-code logic into a shareable package
    * This will allow for better scoping and testing 
* The issueapi logic is now only responsible for parsing the request and rendering the response

This is broken into chunks for reviews, but the sum of these parts is here: https://github.com/google/exposure-notifications-verification-server/pull/1406


pkg/controller/issueapi/issuelogic/issuelogic.go
```diff
@@ -0,0 +1,78 @@
...
+	config     config.IssueAPIConfig
+	db         *database.Database
+	limiter    limiter.Store
+	authApp    *database.AuthorizedApp
+	membership *database.Membership
+	realm      *database.Realm
+}
+
+// New creates a new IssueLogic struct.
+func New(config config.IssueAPIConfig, db *database.Database, limiter limiter.Store,
```
sethvargo, line 68: Design wise, I think it makes sense for authApp, membership, and realm to be params to .Issue, not part of the constructor.

whaught: sure... allright too complicated to sync all these PRs

Issue #1399, opened 2020-12-16
Tracking unit test coverage 


mariliamelo:
### TL;DR
Tracking issue to increase unit tests for validation failures, permissions, failed sms messages w/ rollback, etc

**Expected behavior**
Most critical use cases of our system should be covered by unit tests.

**Observed behavior**
e2e tests cover some, but not all use cases. 

**Additional information**
Filing this issue for tracking this work that may be deprioritized in name of new features.

sethvargo:
I actually think this is done as of https://github.com/google/exposure-notifications-verification-server/pull/1372 (#1296). Let me know if you disagree.

whaught:
The db coverage is great, but I think we can go further to cover the controller layer better. Assert things like HTTP error codes / error enums are correct. Eg. several IssueCode scenarios or the recent 500s on redirector (which I think you also added coverage for)

sethvargo:
Controller tests aren't unit tests though - most are gonna be clicky-clicky. I'm also generally against issues that have no clear success criteria. If we want to open specific issues for specific packages, that will ensure we have the coverage and people can parallelize working on them.

mikehelmick:
We could set a target for test coverage for the whole project and attempt to meet it. Not sure if clickyclicky gets go coverage stats.

setting an initial benchmark of 70%?

sethvargo:
```text
internal/i18n	0.033s	coverage: 0.0% of statements
internal/project	0.029s	coverage: 30.8% of statements
internal/routes	0.187s	coverage: 20.4% of statements
pkg/api	0.034s	coverage: 41.2% of statements
pkg/cache	2.555s	coverage: 62.6% of statements
pkg/controller/admin	24.209s	coverage: 51.4% of statements
pkg/controller/apikey	23.223s	coverage: 23.4% of statements
pkg/controller/appsync	16.547s	coverage: 60.6% of statements
pkg/controller/associated	29.159s	coverage: 85.2% of statements
pkg/controller/codes	22.016s	coverage: 12.6% of statements
pkg/controller/issueapi	0.135s	coverage: 8.4% of statements
pkg/controller/mobileapps	22.923s	coverage: 24.1% of statements
pkg/controller/modeler	15.662s	coverage: 72.0% of statements
pkg/controller/realmadmin	23.122s	coverage: 29.5% of statements
pkg/controller/redirect	17.269s	coverage: 87.4% of statements
[...]

PR #1432, opened 2020-12-18
Test for single issue code

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1399

## Proposed Changes

* Integration test for a few failures of issue-code
* Adds an error code for symptom/test dates that are parsable but rejected - too old

**Release Note**

```release-note
added more test coverage of issue error cases
added error code 'invalid_date' for dates older the allowed for the realm
```

### DIFF
--- a/pkg/testsuite/clients.go
+++ b/pkg/testsuite/clients.go
@@ -57,7 +57,7 @@ func (c *AdminClient) BatchIssueCode(req api.BatchIssueCodeRequest) (int, *api.B
 
 	httpResp, err := c.client.Do(httpReq)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/testsuite/clients.go
```diff
@@ -151,12 +155,12 @@ func (c *APIClient) GetToken(req api.VerifyCodeRequest) (*api.VerifyCodeResponse
 
 	httpResp, err := c.client.Do(httpReq)
 	if err != nil {
-		return nil, fmt.Errorf("failed to POST /api/issue: %w", err)
```
whaught, line 154: this is actually /api/verify

sethvargo (approved):
/lgtm

PR #1433, opened 2020-12-18
Paginate recent codes

whaught:
## Proposed Changes

* This allows a user to find further back into their issued-codes history.
    * Currently there is not search filter by date, but that might make this more useful

recentpaginated

**Release Note**

```release-note
Paginate recently issued codes on code/status
```

### DIFF
--- a/cmd/server/assets/codes/index.html
+++ b/cmd/server/assets/codes/index.html
@@ -60,6 +60,7 @@
         {{end}}
       </div>
     </div>
+    {{template "shared/pagination" .}}
   </main>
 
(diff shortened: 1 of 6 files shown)


mikehelmick:
/hold

We need to think through usefulness of this and make sure there are no privacy issues.

Let's regroup internally

whaught:
> /hold
> 
> We need to think through usefulness of this and make sure there are no privacy issues.
> 
> Let's regroup internally

Yes was going to discuss this with you. It's totally do-able, just a question of if we want to show them.

PR #1436, opened 2020-12-18
separate batch-issue from other e2e flows

mikehelmick:
## Proposed Changes

* e2e runner is currently failing due to wrong path name in e2e client
* separate out batch-issue to separate handler w/ independent schedule so we can monitor separate
* ensure realm is set up correctly for bulk issue when e2e runner starts

**Release Note**

```release-note
end to end test runner now tests bulk issue in a separate handler and scheduler
```

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -116,6 +116,13 @@ func realMain(ctx context.Context) error {
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


sethvargo (approved):
/lgtm

whaught (approved):

PR #1437, opened 2020-12-19
Add full integration test coverage for API keys

sethvargo:
Found  and fixed two bugs too!

**Release Note**

```release-note
Add full integration test coverage for API keys
```

/assign @mikehelmick 
/cc @whaught


pkg/controller/apikey/index.go
```diff
@@ -42,16 +42,11 @@ func (c *Controller) HandleIndex() http.Handler {
...
-		if !membership.Can(rbac.APIKeyRead) {
-			controller.Unauthorized(w, r, c.h)
-			return
-		}
-
 		currentRealm := membership.Realm
 
 		pageParams, err := pagination.FromRequest(r)
 		if err != nil {
-			controller.InternalError(w, r, c.h, err)
```
sethvargo, line 54: Twas bug. Would have paged if someone edited query params. Changed to user error.

pkg/controller/apikey/index.go
```diff
@@ -42,16 +42,11 @@ func (c *Controller) HandleIndex() http.Handler {
 			controller.Unauthorized(w, r, c.h)
 			return
 		}
-		if !membership.Can(rbac.APIKeyRead) {
```
sethvargo, line 45: This is duplicate of the above line.

pkg/controller/apikey/update.go
```diff
@@ -75,13 +75,6 @@ func (c *Controller) HandleUpdate() http.Handler {
 		var form FormData
 		if err := controller.BindForm(w, r, &form); err != nil {
 			authApp.Name = form.Name
-
```
sethvargo, line 78: This does nothing, removing.

pkg/controller/apikey/update.go
```diff
@@ -97,7 +90,7 @@ func (c *Controller) HandleUpdate() http.Handler {
 		}
 
 		flash.Alert("Successfully updated API key!")
-		http.Redirect(w, r, "/apikeys", http.StatusSeeOther)
+		http.Redirect(w, r, fmt.Sprintf("/realm/apikeys/%d", authApp.ID), http.StatusSeeOther)
```
sethvargo, line 93: Twas bug. /apikeys doesn't exist

internal/envstest/server.go
```diff
@@ -336,3 +354,23 @@ func ServerAssetsPath() string {
 func LocalesPath() string {
 	return filepath.Join(project.Root(), "internal", "i18n", "locales")
 }
+
+func randomBytes(tb testing.TB, len int) []byte {
```
whaught, line 358: Nice - I was going for some of this in: https://github.com/google/exposure-notifications-verification-server/pull/1412

pkg/controller/apikey/index.go
whaught: good call. Did we look for other instances of pagination.FromRequest doing this? might be common

whaught (approved):

pkg/controller/apikey/create_test.go
```diff
@@ -43,51 +45,90 @@ func TestHandleCreate(t *testing.T) {
...
-		t.Errorf("expected %v to be %v", got, want)
-	}
-	if got, want := record.Name, "Example API key"; got != want {
-		t.Errorf("expected %v to be %v", got, want)
-	}
-	if got, want := record.APIKeyType, database.APIKeyTypeDevice; got != want {
-		t.Errorf("expected %v to be %v", got, want)
-	}
+	t.Run("middleware", func(t *testing.T) {
+		t.Parallel()
```
whaught, line 49: Seeing test races. chromedp might not be able to handle as much parallel

pkg/controller/apikey/index.go
sethvargo: Pulled this out into https://github.com/google/exposure-notifications-verification-server/pull/1438. I can rebase if that makes it in first.

pkg/controller/apikey/create_test.go
sethvargo: Nah, it's the renderer. There's a legit data race that occurs in debug mode when templates are reloaded. We'd never hit it in prod because templates aren't reloaded. Fixed.

sethvargo:
And we're happy again. There were two data races.

whaught (approved):

PR #1445, opened 2020-12-22
Use realm from current membership for emails

whaught:
## Proposed Changes

* Use the current membership's realm for email templates
* Require UserWrite rbac for reset password

**Release Note**

```release-note
require UserWrite for admin reset of another user's password
```

### DIFF
--- a/pkg/controller/login/reset_password.go
+++ b/pkg/controller/login/reset_password.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/rbac"
 )
 
 func (c *Controller) HandleShowResetPassword() http.Handler {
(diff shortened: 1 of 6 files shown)


mikehelmick (approved):

mikehelmick:
compilation error

pkg/controller/login/reset_password.go
```diff
@@ -59,6 +60,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			return
 		}
 
+		membership := controller.MembershipFromContext(ctx)
+		if !membership.Can(rbac.UserWrite) {
+			controller.Unauthorized(w, r, c.h)
+			return
+		}
 		// Does the user exist?
```
sethvargo, line 68: Nit newline

pkg/controller/login/reset_password.go
```diff
@@ -59,6 +60,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			return
 		}
 
+		membership := controller.MembershipFromContext(ctx)
+		if !membership.Can(rbac.UserWrite) {
```
sethvargo, line 64: Doesn't this method also get called if the user resets their own password? That's why I didn't enforce rbac here.

pkg/controller/login/reset_password.go
```diff
@@ -59,6 +60,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			return
 		}
 
+		membership := controller.MembershipFromContext(ctx)
```
sethvargo, line 63: I don't think we require that a current realm is select for the user to reset their own password.

sethvargo (approved):
/lgtm

PR #1448, opened 2020-12-22
Use first available membership for password reset

whaught:
## Proposed Changes

* Password reset occurs from a context that has no auth set and we allow realms to localize the email templates if using their own SMTP server
* Select the first available membership realm for password reset so the user can get a localized email.
    * Not idea for members of many realms, but that is not expected for most users 

**Release Note**

```release-note
Select a realm-localized template for reset password
```

### DIFF
--- a/pkg/controller/login/reset_password.go
+++ b/pkg/controller/login/reset_password.go
@@ -87,8 +87,13 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			// Use the first membership available. This will help get the user a realm-localized template.
 			// For most users the expected number of memberships is 1.
 			membership, err = user.SelectFirstMembership(c.db)
-			if err != nil && !database.IsNotFound(err) {
-				logger.Warnw("failed retrieving membership", "error", err)
+			if err != nil {
+				if !database.IsNotFound(err) {
+					logger.Infof("No membership found for %s", user.Email)
+				} else {
+					controller.InternalError(w, r, c.h, err)
+					return
+				}
 			}
 		}
 
--- a/pkg/controller/login/reset_password.go
+++ b/pkg/controller/login/reset_password.go
@@ -84,14 +84,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 
 		// This is likely - most users reset password from un-authed context.
 		if membership == nil {
-			memberships, err := user.ListMemberships(c.db)
-			if err != nil {
-				logger.Warnw("failed retrieving memberships", "error", err)
-			}
 			// Use the first membership available. This will help get the user a realm-localized template.
 			// For most users the expected number of memberships is 1.
-			if len(memberships) > 0 {
-				membership = memberships[0]
+			membership, err = user.SelectFirstMembership(c.db)
+			if err != nil {
+				logger.Warnw("failed retrieving membership", "error", err)
 			}
 		}
 
--- a/pkg/controller/login/reset_password.go
+++ b/pkg/controller/login/reset_password.go
@@ -88,7 +88,7 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 			// For most users the expected number of memberships is 1.
 			membership, err = user.SelectFirstMembership(c.db)
 			if err != nil {
-				if !database.IsNotFound(err) {
+				if database.IsNotFound(err) {
 					logger.Infof("No membership found for %s", user.Email)
 				} else {
 					controller.InternalError(w, r, c.h, err)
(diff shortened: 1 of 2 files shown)


pkg/controller/login/reset_password.go
```diff
@@ -79,6 +81,20 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 		var resetComposer auth.ResetPasswordEmailFunc
 
 		membership := controller.MembershipFromContext(ctx)
+
+		// This is likely - most users reset password from un-authed context.
+		if membership == nil {
```
sethvargo, line 86: The middleware sets CurrentMemberships (plural) too - you can use that instead of doing a lookup.

sethvargo (changes requested):

pkg/controller/login/reset_password.go
whaught: only for RequireAuth pages, which this is not. I don't believe the user will have a session (with currentmembership and realm set)

sethvargo: It gets populated by LoadCurrentMembership, not requireauth. And I'm pretty sure this route calls LoadCurrentMembership

whaught: These do:
https://github.com/google/exposure-notifications-verification-server/blob/008e6956679230a7c51085ef8724fbe9f00dae30/internal/routes/server.go#L182

this one is here: https://github.com/google/exposure-notifications-verification-server/blob/008e6956679230a7c51085ef8724fbe9f00dae30/internal/routes/server.go#L168

sethvargo: Ah blarg. I messed up that ordering then. Do you have bandwidth to refactor the middleware to load memberships and require that middleware here?

Loading memberships is really expensive (it's an O(n+1) query), so I'd prefer to do it in middleware and cache it on the context once.

whaught: This is unauthenticated (the user types in their email address) so there is not user or membership yet. I've change this to be less expensive by db querying without loading all memberships.

pkg/database/user.go
```diff
@@ -144,6 +144,27 @@ func (u *User) ListMemberships(db *Database) ([]*Membership, error) {
 	return memberships, nil
 }
 
+// SelectFirstMembership selects the first memberships for this user.
+func (u *User) SelectFirstMembership(db *Database) (*Membership, error) {
```
sethvargo, line 148: Since all our other finder functions raise an error on NotFound, maybe we should name this `FirstMembershipOrNil`?

pkg/database/user.go
```diff
@@ -144,6 +144,27 @@ func (u *User) ListMemberships(db *Database) ([]*Membership, error) {
...
+		Preload("Realm").
+		Preload("User").
+		Model(&Membership{}).
+		Where("user_id = ?", u.ID).
+		Joins("JOIN realms ON realms.id = memberships.realm_id").
+		Order("realms.name").
+		First(&membership).
+		Error; err != nil {
+		if IsNotFound(err) {
+			return membership, nil
```
sethvargo, line 161: Return `nil` instead of the empty membership

pkg/controller/login/reset_password.go
```diff
@@ -84,14 +84,11 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
...
-			if err != nil {
-				logger.Warnw("failed retrieving memberships", "error", err)
-			}
 			// Use the first membership available. This will help get the user a realm-localized template.
 			// For most users the expected number of memberships is 1.
-			if len(memberships) > 0 {
-				membership = memberships[0]
+			membership, err = user.SelectFirstMembership(c.db)
+			if err != nil {
+				logger.Warnw("failed retrieving membership", "error", err)
```
sethvargo, line 91: controller.InternalError & return

pkg/database/user.go
whaught: Let's just return the notfound and not have this one be weird

pkg/controller/login/reset_password.go
```diff
@@ -79,6 +81,22 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
...
 		membership := controller.MembershipFromContext(ctx)
+
+		// This is likely - most users reset password from un-authed context.
+		if membership == nil {
+			// Use the first membership available. This will help get the user a realm-localized template.
+			// For most users the expected number of memberships is 1.
+			membership, err = user.SelectFirstMembership(c.db)
+			if err != nil {
+				if database.IsNotFound(err) {
+					logger.Infof("No membership found for %s", user.Email)
```
sethvargo, line 92: Can we make this `Infow` in another PR?

sethvargo (approved):
/lgtm

PR #1450, opened 2020-12-22
Add more tests for realmadmin

sethvargo:
Also refactor some of the earlier ones to be reusable.

**Release Note**

```release-note
Add more tests for realmadmin
```

### DIFF
--- a/pkg/controller/apikey/create.go
+++ b/pkg/controller/apikey/create.go
@@ -85,12 +85,12 @@ func (c *Controller) HandleCreate() http.Handler {
 }
 
 func bindCreateForm(r *http.Request, app *database.AuthorizedApp) error {
-	type CreateFormData struct {
+	type FormData struct {
 		Name string              `form:"name"`
 		Type database.APIKeyType `form:"type"`
 	}
 
-	var form CreateFormData
+	var form FormData
 	err := controller.BindForm(nil, r, &form)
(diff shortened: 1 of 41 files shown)


pkg/controller/apikey/create.go
```diff
@@ -66,7 +66,7 @@ func (c *Controller) HandleCreate() http.Handler {
 		apiKey, err := currentRealm.CreateAuthorizedApp(c.db, &authApp, currentUser)
 		if err != nil {
 			if database.IsValidationError(err) {
-				w.WriteHeader(http.StatusBadRequest)
+				w.WriteHeader(http.StatusUnprocessableEntity)
```
whaught, line 69: Just wondering - when should we be using 422 over 400? We have several places that validate input and mostly go with the 400

sethvargo: 422 generally means "you gave us data, we got the data, we parsed the data, but the data was invalid for some reason (usually validation)"

400 generall means "we didn't understand the request"

https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm

sethvargo: That being said, we have to keep any _APIs_ at 400 because that's what they're documented as.

whaught (approved):

PR #1451, opened 2020-12-22
Bump core server to pull in logger changes

sethvargo:
/assign @yegle

**Release Note**

```release-note
NONE
```


sethvargo:
Doing this now so we don't forget

whaught (approved):

PR #1454, opened 2020-12-22
Prefer net split over custom indexof

whaught:
## Proposed Changes

* This should be less errorprone as net.SplitHostPort understands more complex combinations which include zone, less maintenance for us


sethvargo (approved):
/lgtm

Issue #1442, opened 2020-12-21
Provide tracking UUIDs from bulk upload


whaught:
Users may still be interested in tracking issued codes from the bulk-upload interface. We expect CSVs to be quite large, so showing all of them on the page might be a non-starter. Instead allow the user to download a copy of the input CSV with the associated UUIDs. Potentially we could also show the errors in such a file too.

/assign

PR #1458, opened 2020-12-23
Show success UUIDs and provide file receipt for bulk-issue

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1442

## Proposed Changes

* Show first 50 success / error cases with uuid for bulk-issue
* After 50 we stop showing them in the card for (potentially) very large lists
    * 50 was arbitrarily chosen, but we don't want the browser to be overwhelmed in large cases
* Update a file-download link with a full CSV formatted receipt of the issuance.

savereceipt

**Release Note**

```release-note
Show first 50 success / error cases for bulk-issue with UUIDs
Allow download of log file for bulk-issue
```

### DIFF
--- a/cmd/server/assets/codes/bulk-issue.html
+++ b/cmd/server/assets/codes/bulk-issue.html
@@ -129,9 +129,25 @@
         </table>
       </div>
     </div>
+
+    <div class="card mb-3 shadow-sm d-none" id="receipt-div">
+      <div class="card-header">Receipt</div>
+      <div class="card-body">
+        <a id="save" class="card-link" download="bulk-issue-log.csv" href="data:text/plain,">Save code results log</a>
+        <small class="form-text text-muted">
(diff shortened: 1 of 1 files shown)


sethvargo:
I _feel_ (read: no evidence) that the UI should only display failures. I like the idea of providing a downloadable report of all data though. I'm curious what @mikehelmick thinks.

whaught:
> I _feel_ (read: no evidence) that the UI should only display failures. I like the idea of providing a downloadable report of all data though. I'm curious what @mikehelmick thinks.

Showing these may not be useful in full production use of them, but some HI testers wanted to see this information and so it may alleviate frustration when testing smaller samples.

I think it only makes sense if the CSV file is sufficiently small (thus the 50 limit). We might also limit the number of errors shown so the page doesn't grow ridiculously but we need the file alternative first (coming soon)

sethvargo (approved):

PR #1461, opened 2020-12-27
Add zapw to detect misuse of logger

sethvargo:
This leverages a static analysis tool I wrote to find and alert when we misuse the logger.

**Release Note**

```release-note
Add zapw to detect misuse of logger
```

/hold


sethvargo:
/unhold

/assign @whaught @mikehelmick 

I wrote a static analyzer to catch these. If you like this, I'll submit a similar PR to the key server.

whaught (approved):

PR #1462, opened 2020-12-28
Staticify bulk issue js

whaught:
## Proposed Changes

* Moves much of the javascript of bulk-issue to application.js
    * Why: caches better in the client browser, separate js controller from html view, IDEs can figure it out too


sethvargo (approved):
/lgtm

I'm gonna trust this is just a copy-paste 😄

PR #1465, opened 2020-12-28
An odd test flake in integration_e2e

whaught:
## Proposed Changes

* example flake: https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/google_exposure-notifications-verification-server/1464/pull-en-server-release-unit/1343646210786856960
* The standard ParseWithClaims fails if !claims.Valid() but we continue on and check for that ourselves too. This resulted in the test failing with the wrong error message.
    * Why does this only fail sometimes? I am not sure.

### DIFF
--- a/pkg/controller/certapi/certapi.go
+++ b/pkg/controller/certapi/certapi.go
@@ -73,7 +73,7 @@ func (c *Controller) validateToken(ctx context.Context, verToken string, publicK
 	logger := logging.FromContext(ctx).Named("certapi.validateToken")
 
 	parser := &jwt.Parser{
-		SkipClaimsValidation: true, // Manually check claims.Valid() below
+		SkipClaimsValidation: true, // we manually check claims.Valid() below
 	}
 	// Parse and validate the verification token.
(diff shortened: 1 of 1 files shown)


pkg/controller/certapi/certapi.go
```diff
@@ -72,8 +72,11 @@ func New(ctx context.Context, config *config.APIServerConfig, db *database.Datab
 func (c *Controller) validateToken(ctx context.Context, verToken string, publicKeys map[string]crypto.PublicKey) (string, *database.Subject, error) {
 	logger := logging.FromContext(ctx).Named("certapi.validateToken")
 
+	parser := &jwt.Parser{
```
whaught, line 75: This triggers the parser to skip this validation: https://github.com/dgrijalva/jwt-go/blob/dc14462fd58732591c7fa58cc8496d6824316a82/parser.go#L62

Note that on line 104 we are checking claims.Valid() ourselves

sethvargo (approved):
/lgtm

Issue #1463, opened 2020-12-28
iOS / ENX redirect


mikehelmick:
### TL;DR

We need a new setting for iOS where a custom app and ENX co-exist.

That setting is to not redirect to the app store if a redirect request reaches the redirect server AND there is iOS metadata.

If EN Express is enabled && there is an iOS app, we should still build the ens:// URL because if the custom app was installed, it would have grabbed the link.

sethvargo:
Added in #1466. FWIW, I think we have a bug in the current implementation in that there's no checking to see if ENX is enabled.

If the user provides a valid user agent for iOS or Android and there's no app registered, the current behavior is to redirect to the `ens://` link, even if the realm doesn't have enx enabled. That's fixed in the PR too.

PR #1466, opened 2020-12-28
Give admins the ability to disable AppStore redirects

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1463

**Release Note**

```release-note
Give admins the ability to disable AppStore redirects for apps
```

/assign @whaught 

/cc @mikehelmick

### DIFF
--- a/cmd/server/assets/mobileapps/_app.html
+++ b/cmd/server/assets/mobileapps/_app.html
@@ -2,11 +2,20 @@
 
 {{$app := .app}}
 
+<style>
+  .d-android-only, .d-app-none, .d-ios-only {
+    display: none;
+  }
+</style>
+
 <div class="form-label-group">
   <input type="text" name="name" id="name" class="form-control{{if $app.ErrorsFor "name"}} is-invalid{{end}}" value="{{$app.Name}}"
     placeholder="Application name" autofocus>
   <label for="name">Application name</label>
   {{template "errorable" $app.ErrorsFor "name"}}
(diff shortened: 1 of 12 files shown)


pkg/database/migrations.go
```diff
@@ -1900,3 +1911,13 @@ func (db *Database) MigrateTo(ctx context.Context, target string, rollback bool)
 	}
 	return m.Migrate()
 }
+
+// multiExec is a helper that executes the given sql clauses against the tx.
+func multiExec(tx *gorm.DB, sqls ...string) error {
```
sethvargo, line 1916: @whaught unrelated, but I got annoyed writing migrations...

cmd/server/assets/mobileapps/_app.html
```diff
@@ -2,41 +2,75 @@
 
 {{$app := .app}}
 
+<style>
```
whaught, line 5: should this be in application.css?

pkg/database/migrations.go
whaught: nice - we do this a lot

whaught (approved):
/lgtm
/hold one question about css. feel free to unhold

cmd/server/assets/mobileapps/_app.html
sethvargo: Meh, if we end up using it across multiple files we can move it over.

Issue #1321, opened 2020-12-11
SMS template - custom expansions


mikehelmick:
### TL;DR

Allow for "caseID" expansion in SMS template

A realm should enable this in their realm settings and it should only be shown in issue UI and accepted on the API if enabled.

Use case: a state wants to give a user a caseID# so that if they call back in, they can get in touch w/ case workers easily.

### Design

Add new setting in the UI
add new expansion in the template [caseID]
add to UI and API (and docs) as appropriate.

whaught:
Note, the issue API already gives out a UUID for tracking. We also take in UUID from the client to prevent multiple SMSes on retries. Although this is uuid format, not a generic string, it might be a good stand-in for something like caseID from the users perspective.

whaught:
We had some concerns about identifying pII and caseID - after circling back with the state that asked, this is no longer a requirement.

/close

PR #1467, opened 2020-12-28
externalCaseID field and SMS expansion

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1321

## Proposed Changes

* Adds a new field to issue request that allows for a transparent string to be included with the issunace
*  Allow SMS expansion of this caseID field
*  **I'm worried about PII**
    * Alternative is to only show our UUID: https://github.com/google/exposure-notifications-verification-server/pull/1464 

**Release Note**

```release-note
Add externalCaseID field to issue request
Allow externalCaseID in SMS expansion
```


sethvargo:
> - **I'm worried about PII**

Hey, me too. I'd like to get 30min with the state that requested this feature before building it.

PR #1469, opened 2020-12-28
Add in the current date to the bulk upload receipt csv filename

whaught:
## Proposed Changes

* It's annoying that this filename is always the same, append the current date.


sethvargo (approved):
/lgtm

PR #1476, opened 2020-12-29
Adding some realm test coverage

whaught:
## Proposed Changes

* Adding some unit-test coverage for some common Realm operations (and removing a dead one)

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -874,21 +874,6 @@ func (r *Realm) ListMemberships(db *Database, p *pagination.PageParams, scopes .
 	return memberships, paginator, nil
 }
 
-// FindMembership finds the membership with the given id, if it exists.
-func (r *Realm) FindMembership(db *Database, userID interface{}) (*Membership, error) {
-	var membership Membership
-	if err := db.db.
-		Preload("User").
-		Preload("Realm").
-		Model(&Membership{}).
-		Where("realm_id = ? AND user_id = ?", r.ID, userID).
-		First(&membership).
-		Error; err != nil {
-		return nil, err
-	}
-	return &membership, nil
-}
-
 // MembershipPermissionMap returns a map where the key is the ID of a user and
 // the value is the permissions for that user.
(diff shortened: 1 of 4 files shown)


pkg/database/realm.go
```diff
@@ -874,21 +874,6 @@ func (r *Realm) ListMemberships(db *Database, p *pagination.PageParams, scopes .
 	return memberships, paginator, nil
 }
 
-// FindMembership finds the membership with the given id, if it exists.
-func (r *Realm) FindMembership(db *Database, userID interface{}) (*Membership, error) {
```
whaught, line 878: `user.FindMembership(db, realmId)` is in use, probably doesn't make sense to have two

pkg/database/realm_test.go
```diff
@@ -501,6 +501,59 @@ func TestRealm_UserStats(t *testing.T) {
...
+
+	db, _ := testDatabaseInstance.NewDatabase(t, nil)
+
+	realm1 := NewRealmWithDefaults("realm1")
+	realm1.RegionCode = "US-MOO"
+	if err := db.SaveRealm(realm1, SystemTest); err != nil {
+		t.Fatal(err)
+	}
+
+	realm2 := NewRealmWithDefaults("realm2")
```
sethvargo, line 515: I don't see realm2 used in the test?

pkg/database/realm_test.go
```diff
@@ -501,6 +501,59 @@ func TestRealm_UserStats(t *testing.T) {
...
+	if err := db.SaveRealm(realm2, SystemTest); err != nil {
+		t.Fatal(err)
+	}
+
+	// FindRealm by ID
+	{
```
sethvargo, line 521: Instead of a scope, can you try using a sub-test?

```go
t.Run("by_id", func(t *testing.T) {
  t.Parallel()

  // ...
})
```

That will allow them to run in parallel (faster) and keep a clean scope.

whaught: made it a table test

sethvargo (approved):

PR #1481, opened 2020-12-29
Add tests for header middleware

sethvargo:
**Release Note**

```release-note
Add tests for header middleware
```

### DIFF
--- a/pkg/controller/middleware/header.go
+++ b/pkg/controller/middleware/header.go
@@ -15,10 +15,8 @@
 package middleware
 
 import (
-	"context"
 	"net/http"
 
-	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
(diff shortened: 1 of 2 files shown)


pkg/controller/middleware/header.go
```diff
@@ -47,34 +40,29 @@ func RequireHeader(h render.Renderer, header string) mux.MiddlewareFunc {
...
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			vals := r.Header.Values(header)
 			if len(vals) == 0 {
-				logger.Debugw("missing required header", "header", header)
 				controller.Unauthorized(w, r, h)
 				return
 			}
 
 			found := false
-		LOOP:
```
whaught, line 63: why no more `break LOOP` once its found?

whaught: sorry inner loop is gone

whaught (approved):

PR #1484, opened 2020-12-29
Display OS on the mobile apps index table

whaught:
## Proposed Changes

*  Add the OS to the mobile_apps index table
    * In real life folks seem to have the same name for their iOS app and Android. Eg. Covidwise.  It's impossible to tell which is which from the index page.

**Release Note**

```release-note
Display OS on the mobile apps index table
```


sethvargo (approved):
/lgtm

PR #1485, opened 2020-12-29
Calculate grace periods from membership date

sethvargo:
This changes the MFA grace period calculation to calculate against the date on which the user was added to the realm vs the date on which the user was created. All membership creation dates are backfilled to NOW().

**Release Note**

```release-note
Calculate grace periods from membership date
```

/assign @whaught

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -469,14 +469,14 @@ func (r *Realm) GetLongCodeDurationHours() int {
 	return int(r.LongCodeDuration.Duration.Hours())
 }
 
-// EffectiveMFAMode returns the realm's default MFAMode but first
-// checks if the user is in the grace-period (if so, required becomes prompt).
-func (r *Realm) EffectiveMFAMode(user *User) AuthRequirement {
+// EffectiveMFAMode returns the realm's default MFAMode but first checks if the
+// time is in the grace-period (if so, required becomes prompt).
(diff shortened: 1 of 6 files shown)


pkg/database/realm.go
```diff
@@ -469,14 +469,14 @@ func (r *Realm) GetLongCodeDurationHours() int {
 	return int(r.LongCodeDuration.Duration.Hours())
 }
 
-// EffectiveMFAMode returns the realm's default MFAMode but first
-// checks if the user is in the grace-period (if so, required becomes prompt).
-func (r *Realm) EffectiveMFAMode(user *User) AuthRequirement {
+// EffectiveMFAMode returns the realm's default MFAMode but first checks if the
+// time is in the grace-period (if so, required becomes prompt).
+func (r *Realm) EffectiveMFAMode(t time.Time) AuthRequirement {
```
whaught, line 474: alternative: pass in the membership, use membership.CreatedAt if not null or fall back to user.CreatedAt?

whaught: actually looks like we make that not-null and set to now(). I wonder will that be a problem for existing users (gives them an extra N days)?

whaught: meh this is just for grace periods. I've convinced myself this is fine

whaught (approved):

PR #1488, opened 2020-12-29
Add tests for sessions middleware

sethvargo:
**Release Note**

```release-note
Add tests for sessions middleware
```

### DIFF
--- a/pkg/controller/middleware/sessions_test.go
+++ b/pkg/controller/middleware/sessions_test.go
@@ -1,0 +1,56 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


pkg/controller/middleware/sessions_test.go
```diff
@@ -0,0 +1,56 @@
...
+	r.Header.Set("Accept", "application/json")
+
+	w := httptest.NewRecorder()
+
+	// Verify the proper fields are added to the template map.
+	requireSession(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		session := controller.SessionFromContext(ctx)
+		if session == nil {
```
whaught, line 52: Can we assert that some piece of data is in session? That might be better than just not nil

sethvargo: The middleware doesn't insert any data into the session.

whaught (approved):

PR #1491, opened 2020-12-29
Cover some audits unit tests

whaught:
## Proposed Changes

* Covers some audits for MobileApp, AuthorizedApp, Realm, and User
    * I'll keep going with some others, but this is very boring

### DIFF
--- a/pkg/database/authorized_app_test.go
+++ b/pkg/database/authorized_app_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/timeutils"
+	"github.com/google/exposure-notifications-verification-server/pkg/pagination"
 	"github.com/jinzhu/gorm"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/database/authorized_app_test.go
```diff
@@ -322,3 +323,33 @@ func TestDatabase_PurgeAuthorizedApps(t *testing.T) {
...
+	}
+
+	authorizedApp := &AuthorizedApp{
+		Name: "Appy",
+	}
+	if _, err := realm.CreateAuthorizedApp(db, authorizedApp, SystemTest); err != nil {
+		t.Fatal(err)
+	}
+
+	authorizedApp.Name = "something else"
```
whaught, line 343: We currently don't have an audit for API key fields. Do users tend to delete and recreate there? If there's an update flow we should cover that

sethvargo, line 343: Hmm - we should probably generate an audit entry

sethvargo (approved):
/lgtm

PR #1492, opened 2020-12-29
audit key changes for authorized apps

whaught:
## Proposed Changes

* These fields of authorized_app are missing for audits - trying to ensure everything is covered


pkg/database/authorized_app.go
```diff
@@ -264,6 +264,18 @@ func (db *Database) SaveAuthorizedApp(a *AuthorizedApp, actor Auditable) error {
 				audits = append(audits, audit)
 			}
 
+			if existing.APIKey != a.APIKey {
```
sethvargo, line 267: The API key would never change, its immutable and also one-way hashed. I also don't want the API key to appear in a logl.

sethvargo (changes requested):

pkg/database/authorized_app.go
whaught: /close

PR #1493, opened 2020-12-30
Add tests for users, fix template NPE, fix user HTTP return codes, show validations

sethvargo:
Still needs tests for stats and import, but I'll do that in another PR since this is already large.

**Release Note**

```release-note
Add tests for users
```


cmd/server/assets/header.html
```diff
@@ -55,39 +55,41 @@
     </div>
   {{end}}
 
-  <nav class="nav nav-tabs navbar-expand-md navbar-light bg-light">
```
sethvargo, line 58: This is just indented (?w=1). It can really only fail in tests, but it's good to guard against it anyway in the template.

pkg/controller/user/stats.go
```diff
@@ -84,7 +83,7 @@ func (c *Controller) HandleUserStats() http.Handler {
 			c.h.RenderJSON(w, http.StatusOK, stats)
 			return
 		default:
-			controller.InternalError(w, r, c.h, fmt.Errorf("unknown path %q", pth))
+			controller.BadRequest(w, r, c.h)
```
sethvargo, line 86: Twas a bug

pkg/controller/user/export.go
```diff
@@ -51,17 +59,42 @@ func (c *Controller) HandleExport() http.Handler {
 			return
 		}
 
-		w.Header().Add("Content-Disposition", "")
-		w.Header().Add("Content-Type", "text/CSV")
+		filename := fmt.Sprintf("%s-users.csv", time.Now().Format(project.RFC3339Squish))
```
whaught, line 62: Yay nice touch.

I did this to the bulk-upload files in javascript: https://github.com/google/exposure-notifications-verification-server/pull/1469/files

Now I'm wondering if I should flip it with the date first (for string sortability)

pkg/controller/user/export.go
```diff
@@ -51,17 +59,42 @@ func (c *Controller) HandleExport() http.Handler {
...
+
+	if err := w.Write([]string{"name", "email", "joined"}); err != nil {
+		return nil, fmt.Errorf("failed to write CSV header: %w", err)
+	}
+
+	for i, m := range s {
+		if err := w.Write([]string{
+			m.User.Name,
+			m.User.Email,
+			m.CreatedAt.Format(project.RFC3339Date),
```
whaught, line 88: Wondering: why add CreatedAt - presumably when we import these we want a fresh CreatedAt?

pkg/database/user.go
```diff
@@ -52,8 +52,7 @@ func (u *User) BeforeSave(tx *gorm.DB) error {
 	u.Email = project.TrimSpace(u.Email)
 	if u.Email == "" {
 		u.AddError("email", "cannot be blank")
-	}
-	if !strings.Contains(u.Email, "@") {
+	} else if !strings.Contains(u.Email, "@") {
 		u.AddError("email", "appears to be invalid")
```
whaught, line 56: nit: `appears to be invalid` is weirdly passive. `invalid email address` or `email address must include @`?

pkg/controller/user/create.go
```diff
@@ -41,87 +35,113 @@ func (c *Controller) HandleCreate() http.Handler {
...
+		}
+		if existing != nil && existing.ID != 0 {
+			user.ID = existing.ID
+		}
+
+		// Create or update user.
+		if err := c.db.SaveUser(user, currentUser); err != nil {
+			if database.IsValidationError(err) {
+				w.WriteHeader(http.StatusUnprocessableEntity)
+				c.renderNew(ctx, w, user, userMembership)
```
whaught, line 80: Can we add `show User validation errors` to the description too

pkg/controller/user/export.go
sethvargo: The date is first

pkg/controller/user/export.go
sethvargo: I think we would discard that value on import (we don't consume it). This is for if someone wants to know _when_ someone was added to a realm.

pkg/controller/user/export.go
whaught: I agree! I'm saying can you change where I did it wrong for consistency (or I'll slip that in to my next PR)

pkg/database/user.go
sethvargo: Done

pkg/controller/user/create.go
sethvargo: Done

whaught (approved):

Issue #1305, opened 2020-12-09
Captcha fails to load on failed login


sethvargo:
There may be other ways to reproduce but:

1. Attempt to login with a valid email, but invalid password
2. Get login failed error
3. Try to login with the same account with the valid password
4. Get invalid captcha popup:

    <img width="499" alt="screenshot-20201209-150638@2x" src="https://user-images.githubusercontent.com/408570/101681529-411a0e80-3a30-11eb-8331-e76e22b5e559.png">

- Probably should be an alert, maybe a popover
- Probably shouldn't happen 😄 

/assign @whaught

whaught:
Docs for the firebase reset-on-failure is here: https://firebase.google.com/docs/auth/web/phone-auth#use-invisible-recaptcha

PR #1495, opened 2020-12-30
Late load recaptcha - only require for 2nd factor SMS

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1305

## Proposed Changes

* Previously we protected the entire login flow with recaptcha, but it is only required for SMS send. Unfortunately the recaptcha reset-on-failure does not appear to work (on not-localhost).
Instead we load the recaptcha an only render it after successful sign-in with username/password for the SMS send.  

**Release Note**

```release-note
Don't load captcha unless SMS factor needed
```

### DIFF
--- a/cmd/server/assets/login/login.html
+++ b/cmd/server/assets/login/login.html
@@ -62,6 +62,8 @@
         </div>
       </div>
     </div>
+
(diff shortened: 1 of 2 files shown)


sethvargo (approved):
/lgtm

PR #1498, opened 2020-12-30
Pass in zaptest.Logger to tests

sethvargo:
Now that we have many more tests, the log output is largely impossible when a single test fails.

[zaptest](https://pkg.go.dev/go.uber.org/zap/zaptest) fixes this as a special logger than only logs when `t` fails. Unfortunately that required a bit more plumbing than I had expected.

**Release Note**

```release-note
Pass in zaptest.Logger to tests
```


whaught (approved):
this change was less scary that the file-count would have you believe

whaught (approved):

whaught (approved):

PR #1503, opened 2021-01-04
Ping db on health instead of opening new connection

sethvargo:
The previous logic relied on creating a new database connection. This would be a different database connection from the one the rest of the app is using, and could lead to false successes. This changes the health endpoint to use the same db that the other handlers use.

Additionally, since it's just a ping now (instead of a full config parsing), the limit is increased from 1/min to 1/sec. The response code when the limit is exceeded is now a 429 instead of silently falling through to 200.

Finally, the monitoring uptime checks are updated to use this endpoint instead of the generic /health. It's safe to say that the service is "down" if it cannot connect to the database.

**Release Note**

```release-note
Ping db on /health instead of opening new connection
```

/assign @yegle 
/assign @whaught

### DIFF
--- a/internal/routes/adminapi.go
+++ b/internal/routes/adminapi.go
@@ -93,7 +93,7 @@ func AdminAPI(
 	processFirewall := middleware.ProcessFirewall(h, "adminapi")
 
 	// Health route
-	r.Handle("/health", controller.HandleHealthz(ctx, &cfg.Database, h)).Methods("GET")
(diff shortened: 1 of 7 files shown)


whaught:
/lgtm

Maybe I'm lacking context of history, but I wonder why we were opening a connection here in the first place: your code looks a lot more like the rest of the service.

sethvargo:
I couldn't think of any reason nor could I find anything in the commits on that file. I _suspect_ we wrote that before config parsing and config loading were separate functions?

PR #1504, opened 2021-01-05
Unify in-memory server bootstrap

sethvargo:
This brings the API server and Admin API server bootstrapping into envstest like the UI server, making it possible to spin up in-memory instances of both. This removed a TON of duplication between envstest and testsuite, eliminating the need for some intermediate packages.

This also unifies the 3 different "client library" packages floating around into a single internal/clients package that supports the superset of all the APIs, dangled off an appropriate API client.

**Release Note**

```release-note
Unify in-memory server bootstrap
```

/assign @whaught


internal/clients/appsync.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
```
whaught, line 1: oh yea... it is 2021 now isn't it

whaught (approved):
Yay deleted code - I always thought the existence of two clients was unusual (I actually screwed up the batch e2e test the first time, thinking I had already made the client and not realizing it didn't execute on Prow - Mike fixed it when it started paging on as we deployed). I was not, however, ambitious enough to make refactors like this, but this should make our testing story better and leaves less room for differences from e2e and integration tests.

Love it.

PR #1507, opened 2021-01-05
Temporarily don't check database connection

sethvargo:
This is required because Terraform configs are run from HEAD, but the previous tag (0.18.5) is not compatible with this change. We can re-add this after 0.19 is tagged and ready.

**Release Note**

```release-note
Temporarily disable database service check due to incompatibility between releases
```


yuriatgoogle (approved):
/lgtm

mikehelmick (approved):

PR #1518, opened 2021-01-06
Update firebase to the latest version

whaught:
## Proposed Changes

* Update Firebase dep from 7.22.1 to 8.2.1
    * from the release notes:
        *  "Fixed an issue with IndexedDB retry logic causing uncaught errors."
        * "Updated auth token logic to rely on device clock time instead of server time. This fixes an issue seen when a device's clock is skewed by a lot"


sethvargo (approved):
/lgtm

Issue #1388, opened 2020-12-16
Allow bulk management of user permissions


sethvargo:
We introduced RBAC, but we should add realm admin pages to display inverted lists (e.g. list all users with permission X) and provide a management console to bulk add/remove permissions for a subset of users.

/assign

PR #1525, opened 2021-01-07
Introduce advanced search and bulk permissions

sethvargo:
This change has two major parts: advanced searching and bulk permission management.

**Advanced searching**

Advanced searching is a new feature for user search that allows for more granular searching. It's backwards-compatible with the former search syntax, but now includes the ability to filter searches by fields. Specifically, you can now query like "name:foo" or "can:APIKeyWrite" to perform advanced queries. Advanced searching is useful for bulk permission management. For example, I can search for all users that have a particular permission, then click the checkboxes next to the names that should have said permission, then remove it in a single form.

**Bulk permission management**

Bulk permission management is a new feature that allows people with UserWrite permissions to modify other users' permissions in bulk from the users index page or a user search.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1388

**Release Note**

```release-note
Advanced searching is a new feature for user search that allows for more granular searching. It's backwards-compatible with the former search syntax, but now includes the ability to filter searches by fields. Specifically, you can now query like "name:foo" or "can:APIKeyWrite" to perform advanced queries.

Bulk permission management is a new feature that allows people with `UserWrite` permissions to modify other users' permissions in bulk in their realm.
```

/assign @whaught 
/assign @mikehelmick

### DIFF
--- a/cmd/server/assets/users/index.html
+++ b/cmd/server/assets/users/index.html
@@ -2,6 +2,7 @@
 
 {{$csrfField := .csrfField}}
 {{$memberships := .memberships}}
+{{$permissions := .permissions}}
 
 {{$currentMembership := .currentMembership}}
 {{$canWrite := $currentMembership.Can rbac.UserWrite}}
@@ -11,6 +12,17 @@
 
 <head>
   {{template "head" .}}
+
+  <style>
+    .dropdown-hover:hover {
+      background-color: #f8f9fa;
+    }
+
+    .custom-control-label.py-2::before,
+    .custom-control-label.py-2::after {
+      top: 0.72rem;
+    }
+  </style>
 </head>
 
 <body id="users-index" class="tab-content">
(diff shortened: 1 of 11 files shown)


cmd/server/assets/users/index.html
```diff
@@ -49,45 +61,105 @@
       </div>
 
       {{if $memberships}}
-        <table class="table table-bordered table-striped table-fixed table-inner-border-only border-top mb-0">
-          <thead>
-            <tr>
-              <th scope="col">Name</th>
-              <th scope="col" width="300">Email</th>
-              {{if $canWrite}}
+        <form method="POST" id="users-form">
```
sethvargo, line 64: https://github.com/google/exposure-notifications-verification-server/pull/1525/files?w=1 is a better diff

whaught, line 64: I think I'll need to try this out and see it in action to really verify this.

How are we going to document this? Should `#search` above have a `?` / `advanced` link with a pop up dialog explaining the advanced query options?

/lgtm
/hold

sethvargo:
I realized last night that this implementation missing auditing. I'll need to refactor this a bit, so hold off on reviews until re-ping.

cmd/server/assets/users/index.html
sethvargo: I added an advanced query box

sethvargo:
@whaught @mikehelmick can you re-review this. I had to make some changes for auditing.

whaught (approved):

PR #1526, opened 2021-01-07
Support for Twilio messaging service

whaught:
## Proposed Changes

* Twilio provides a service to manage one or a pool many phone numbers and rules about how to use them. This requires passing an identifier instead of a from-number. See: https://www.twilio.com/docs/messaging/services
* Adds support for a messagingServiceSid instead of a from-number
* Note: I'd like to test this against Twilio, but trial accounts don't get these - I hear they cost more.

**Release Note**

```release-note
Added support for Twilio messaging services
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1929,6 +1929,17 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`ALTER TABLE authorized_app_stats DROP COLUMN IF EXISTS tokens_invalid`)
 			},
 		},
+		{
+			ID: "00082-AddTwilioMessageSid",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ADD COLUMN IF NOT EXISTS twilio_messaging_service_sid varchar(34)`)
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs DROP COLUMN IF EXISTS twilio_messaging_service_sid`)
+			},
+		},
 	}
 }
 
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1933,11 +1933,11 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 			ID: "00083-ExpandTwilioFrom",
 			Migrate: func(tx *gorm.DB) error {
 				return multiExec(tx,
-					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number  TYPE varchar(255)`)
+					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number TYPE varchar(255)`)
 			},
 			Rollback: func(tx *gorm.DB) error {
 				return multiExec(tx,
-					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number  TYPE varchar(16)`)
+					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number TYPE varchar(16)`)
 			},
 		},
 	}
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1930,14 +1930,14 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 			},
 		},
 		{
-			ID: "00083-AddTwilioMessageSid",
+			ID: "00083-ExpandTwilioFrom",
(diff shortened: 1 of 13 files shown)


whaught:
/hold

This may not be a good idea if we wish to steer users away from this feature of Twilio. Hold for discussion.

pkg/database/migrations.go
```diff
@@ -1929,6 +1929,17 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`ALTER TABLE authorized_app_stats DROP COLUMN IF EXISTS tokens_invalid`)
 			},
 		},
+		{
+			ID: "00083-AddTwilioMessageSid",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ADD COLUMN IF NOT EXISTS twilio_messaging_service_sid varchar(34)`)
```
sethvargo, line 1936: We should let this be TEXT or varchar(255)

sethvargo (approved):
We have a paid Twilio account for testing. Ping me on chat and I can walk you through how to access it.

sethvargo:
My high-level feedback is that a MessagingServiceSid is basically a "from" number. The only functional change is the query parameter key that gets sent. From a UX perspective, I think it would be better to use the existing "from" number field (updated some text in the UI). Then, it's trivial for us to detect if the value is a phone number or from number (MSIDs always start with MG).

pkg/database/migrations.go
whaught: done

whaught:
> My high-level feedback is that a MessagingServiceSid is basically a "from" number. The only functional change is the query parameter key that gets sent. From a UX perspective, I think it would be better to use the existing "from" number field (updated some text in the UI). Then, it's trivial for us to detect if the value is a phone number or from number (MSIDs always start with MG).

I have merged them into one field in the UI - I've left the placeholder as "from number" but updated the small text detail

pkg/database/migrations.go
```diff
@@ -1929,6 +1929,17 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`ALTER TABLE authorized_app_stats DROP COLUMN IF EXISTS tokens_invalid`)
 			},
 		},
+		{
+			ID: "00083-AddTwilioMessageSid",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ADD COLUMN IF NOT EXISTS twilio_messaging_service_sid varchar(255)`)
```
sethvargo, line 1936: Sorry, I was trying to say: I don't think we want a new field/column at all.

- Make sure `from` can handle 255 characters
- Add something in sms.go:
    ```go
    if strings.HasPrefix(from, "MG") {
      q.Add("MessagingServiceSid", from)
    } else {
      q.Add("From", from)
    }
    ```

That should dramatically reduce complexity and avoid boolean logic for `if msid else from`.

whaught: sure - expanded the existing db column instead.

pkg/database/migrations.go
```diff
@@ -1929,6 +1929,17 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
 		},
+		{
+			ID: "00083-ExpandTwilioFrom",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number TYPE varchar(255)`)
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ALTER COLUMN twilio_from_number TYPE varchar(16)`)
```
sethvargo, line 1940: Let's not rollback the size change, that _will_ lose data unnecessarily.

pkg/sms/twilio.go
```diff
@@ -25,6 +25,8 @@ import (
 	"time"
 )
 
+const TwilioMessagingServiceSidPrefix = "MG"
```
sethvargo, line 28: Nit doc

sethvargo (approved):
/lgtm

There's two nits: one on migration and one about docs. Feel free to do those here and add your own LGTM or a followup.

pkg/database/migrations.go
whaught: sure - we do have similar rollbacks when we've done this in the past:
eg: `"ALTER TABLE verification_codes ALTER COLUMN long_code TYPE varchar(20)",`

sethvargo: hmm yea we shouldn't roll back to a smaller column

PR #1527, opened 2021-01-07
Add a CloudSchedulerJobFailed alert based on log metric.

yegle:
```release-note
Monitoring: Add ClouldSchedulerJobFailed alert.
```


sethvargo (approved):
/lgtm

I think we need to do this on the key server too

PR #1529, opened 2021-01-07
Nits: never rollback code sql type, check content-type header

whaught:
## Proposed Changes

* Never rollback the code and long_code fields to be shorter - this would lose data
* Check the content-type header when decoding JSON and give a more detailed response if unexpected type

### DIFF
--- a/internal/clients/clients.go
+++ b/internal/clients/clients.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


internal/clients/clients.go
```diff
@@ -126,10 +126,20 @@ func (c *client) do(req *http.Request, out interface{}) (*http.Response, error)
 	defer resp.Body.Close()
 
 	r := io.LimitReader(resp.Body, c.maxBodySize)
+
+	ct := resp.Header.Get("Content-Type")
+	if ct != "application/json" {
```
sethvargo, line 131: We need to check the prefix here, not exact value. `Content-Type: application/json; charset=uft8` is a valid header, for example.

sethvargo (approved):

internal/clients/clients.go
whaught: done

sethvargo (approved):
/lgtm

PR #1531, opened 2021-01-07
Support testType field as input for bulk-upload

whaught:
## Proposed Changes

* Currently `TestType` is hard-coded to "confirmed"
* Allows an optional field to provide another value, remains "confirmed" if omitted

**Release Note**

```release-note
Support testType field as input for bulk-upload
```


sethvargo (approved):
/lgtm

PR #1539, opened 2021-01-08
Pad batch requests and eliminate inner padding

whaught:
## Proposed Changes

* Adds padding to the batch API request / repsonse
* Initializes padding of the inner IssueRequest
    * Change the custom marshaller to ignore already-initialized padding
* We add `omitempty` to the standard IssueResponse (so the batch doesn't get it), but in practice all single-issue calls will have the padding

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -126,24 +126,26 @@ type Padding []byte
 // MarshalJSON is a custom JSON marshaler for padding. It generates and returns
 // 1-2kb (random) of base64-encoded bytes.
 func (p Padding) MarshalJSON() ([]byte, error) {
-	bi, err := rand.Int(rand.Reader, big.NewInt(1024))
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -213,7 +215,7 @@ type IssueCodeRequest struct {
 
 // IssueCodeResponse defines the response type for IssueCodeRequest.
 type IssueCodeResponse struct {
-	Padding Padding `json:"padding"`
+	Padding Padding `json:"padding,omitempty"`
```
mikehelmick, line 218: this is a potentially breaking change - shouldn't be since we never leave blank on single issue, but perhaps worth noting.

whaught: added to the pr description, but yes the single-issue should be unchanged

PR #1540, opened 2021-01-08
fix lints. Add presubmit overall test coverage output

mikehelmick:
## Proposed Changes

* fix lint / static check errors on test mains
* add test coverage output to presubmit script

**Release Note**

```release-note
DEV: Overall test coverage displayed on presubmit.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,8 +75,13 @@ test-acc:
 		-race \
 		-timeout=10m \
 		-vet="${VETTERS}" \
-		./...
+		./... \
+		-coverprofile=coverage.out
 .PHONY: test-acc
+
+test-coverage:
+	@go tool cover -func ./coverage.out | grep total
+.PHONY: test-coverage
 
 e2e-test:
 	@go test \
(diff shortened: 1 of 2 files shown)


pkg/controller/admin/admin_test.go
```diff
@@ -25,5 +26,6 @@ var testDatabaseInstance *database.TestInstance
 func TestMain(m *testing.M) {
 	testDatabaseInstance = database.MustTestInstance()
 	defer testDatabaseInstance.MustClose()
-	m.Run()
```
sethvargo, line 28: Don't do this. Update staticcheck and it'll go away.

sethvargo (changes requested):

pkg/controller/admin/admin_test.go
sethvargo: https://github.com/dominikh/go-tools/issues/708

sethvargo (approved):
/lgtm

PR #1541, opened 2021-01-08
Fix padding

whaught:
## Proposed Changes

* @mikehelmick was right and I should have tested this. **https://github.com/google/exposure-notifications-verification-server/pull/1539** added `omitempty` to the json response which causes MarshalJSON to never be called to begin with
* Corrected and added test
    * The downside is we do get inner `padding:""` for batch, but that's not the worst thing

**Release Note**

```release-note
correct padding
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -126,24 +126,26 @@ type Padding []byte
 // MarshalJSON is a custom JSON marshaler for padding. It generates and returns
 // 1-2kb (random) of base64-encoded bytes.
 func (p Padding) MarshalJSON() ([]byte, error) {
-	bi, err := rand.Int(rand.Reader, big.NewInt(1024))
-	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/api/api.go
```diff
@@ -149,6 +149,16 @@ func (p Padding) MarshalJSON() ([]byte, error) {
 	return []byte(s), nil
 }
 
+// UnmarshalJSON is a custom JSON unmarshaler for padding.
+// The field is meaningless bytes, so this is just a passthrough.
+func (p *Padding) UnmarshalJSON(b []byte) error {
```
mikehelmick, line 154: is this necessary?

whaught: Functionally no, but we need this to be able to assert that the padding is there in the tests

mikehelmick (approved):

PR #1546, opened 2021-01-09
Split base url and path for appsync

whaught:
## Proposed Changes

* Split base url and path for appsync

**Release Note**

```release-note
Fix for appsync which currently fails to reach its intended url
```


whaught:
/close

https://github.com/google/exposure-notifications-verification-server/pull/1547
should allow us to hit a generic absolute path without this problem

eg. code before this fix would attempt `"https://www.gstatic.com/exposurenotifications/apps.json/`

PR #1549, opened 2021-01-10
Check session expiration on login

sethvargo:
Prior to this commit, the login page would re-authenticate, even if the time since last activity exceeded the configuration. This forces the login controller to check the session TTL before rendering the HTML (which would refresh the TTL).

**Release Note**

```release-note
Check session expiration on login
```

/assign @whaught


pkg/controller/middleware/auth.go
```diff
@@ -53,8 +53,6 @@ func RequireAuth(cacher cache.Cacher, authProvider auth.Provider, db *database.D
 				// If it's been more than the TTL since we've seen this session,
 				// expire it by creating a new empty session.
 				if time.Since(t) > sessionIdleTTL {
-					authProvider.ClearSession(ctx, session)
```
sethvargo, line 56: Clearing the session here actually makes it so the login controller "forgets" which session to revoke.

whaught (approved):

Issue #1389, opened 2020-12-16
Alert when a human accesses a secret


sethvargo:
### TL;DR
Fire an alert when a user account accesses a secret in Secret Manager

### Design

- [x] Enable Data Access Logs `DATA_READ` for Google Secret Manager ([more info](https://cloud.google.com/secret-manager/docs/audit-logging)), ideally via a configurable Terraform variable. The default value should be "off" due to the increased costs of enabling DAL.
- [x] Fire a monitoring alert when that audit log appears and the actor type is "user". We **must** exclude service accounts since it's expected that service accounts access the secrets

sethvargo:
After some research, there's no way to enable DAL via Terraform. There's an API, but it also doesn't seem to work. As such, I've enabled DAL on the _organization_ which will apply to all current and future projects.

sethvargo:
(To be clearer, it [exists](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_iam#google_project_iam_audit_config), but it's authoritative. There's no way to say "enable this" - it's the entire audit policy on the project/folder. Since we don't know if there's other configuration on the project, it's not safe to apply since we might be deleting an existing audit config).

sethvargo:
Most of this is done, but there's still an open issue with Terraform accessing secrets as part of its run...

sethvargo:
These alerts fire!

PR #1552, opened 2021-01-11
Create metrics for when humans access secrets or decrypt values

sethvargo:
Part of https://github.com/google/exposure-notifications-verification-server/issues/1389

**Release Note**

```release-note
Create metrics for when humans access secrets or decrypt values
```

/assign @yegle
/assign @yuriatgoogle

### DIFF
--- a/terraform/alerting/monitoring.tf
+++ b/terraform/alerting/monitoring.tf
@@ -68,3 +68,72 @@ resource "google_logging_metric" "stackdriver_export_error_count" {
     value_type  = "INT64"
   }
 }
+
+resource "google_logging_metric" "human_accessed_secret" {
+  name    = "human_accessed_secret"
+  project = var.project
+
+  filter = <<EOT
+resource.type="audited_resource"
+resource.labels.service="secretmanager.googleapis.com"
+resource.labels.method:"AccessSecretVersion"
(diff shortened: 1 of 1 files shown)


yuriatgoogle (approved):
/lgtm

terraform/alerting/monitoring.tf
```diff
@@ -68,3 +68,72 @@ resource "google_logging_metric" "stackdriver_export_error_count" {
...
 }
+
+resource "google_logging_metric" "human_accessed_secret" {
+  name    = "human_accessed_secret"
+  project = var.project
+
+  filter = <<EOT
+resource.type="audited_resource"
+resource.labels.service="secretmanager.googleapis.com"
+resource.labels.method:"AccessSecretVersion"
```
yegle, line 79: You can merge these two metrics together with a filter on `resource.label.method`. If the goal is to set up alert regardless which metric is triggered, there's no need to have two metrics.

You can extract method as a label if needed.

sethvargo: I want to keep them separate because some stakeholders only care about secrets. Once I figure out the alerts, "alert_on_kms" will be configurable, so I think it'll be easier to keep them separate

PR #1555, opened 2021-01-11
Add other email templates to the realm settings UI

whaught:
## Proposed Changes

* Adds password-reset and email-verification email templates into the email realm settings
    * We've got all the db fields wired up
    * 
templates

**Release Note**

```release-note
Add password-reset and email-verification email templates to realm settings
```


sethvargo (approved):
/lgtm

PR #1556, opened 2021-01-11
Fire alerts when humans do secret things

sethvargo:
Part of https://github.com/google/exposure-notifications-verification-server/issues/1389

We still need to figure out how to handle the situation where humans are running Terraform. 

**Release Note**

```release-note
Fire alerts when humans do secret things
```

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -253,3 +253,85 @@ resource "google_monitoring_alert_policy" "CloudSchedulerJobFailed" {
     null_resource.manual-step-to-enable-workspace,
   ]
 }
+
+resource "google_monitoring_alert_policy" "HumanAccessedSecret" {
+  count = var.alert_on_human_accessed_secret ? 1 : 0
+
+  project      = var.project
+  display_name = "HumanAccessedSecret"
+  combiner     = "OR"
+
+  conditions {
+    display_name = "A non-service account accessed a secret."
+
+    condition_monitoring_query_language {
+      duration = "60s"
+
+      query = <<-EOT
+      fetch audited_resource
+      | metric 'logging.googleapis.com/user/human_accessed_secret'
+      | align rate(5m)
+      | every 1m
+      | group_by [resource.project_id],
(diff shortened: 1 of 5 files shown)


terraform/alerting/alerts.tf
```diff
@@ -253,3 +253,85 @@ resource "google_monitoring_alert_policy" "CloudSchedulerJobFailed" {
...
+
+  documentation {
+    content   = "${local.playbook_prefix}/HumanAccessedSecret.md"
+    mime_type = "text/markdown"
+  }
+
+  notification_channels = [for x in values(google_monitoring_notification_channel.channels) : x.id]
+
+  depends_on = [
+    null_resource.manual-step-to-enable-workspace,
```
yegle, line 294: Add a dependency on the custom log based metric.

terraform/alerting/alerts.tf
```diff
@@ -253,3 +253,85 @@ resource "google_monitoring_alert_policy" "CloudSchedulerJobFailed" {
...
+    condition_monitoring_query_language {
+      duration = "60s"
+
+      query = <<-EOT
+      fetch audited_resource
+      | metric 'logging.googleapis.com/user/human_accessed_secret'
+      | align rate(5m)
+      | every 1m
+      | group_by [resource.project_id],
+          [value_human_accessed_secret_aggregate: aggregate(value.human_accessed_secret)]
```
yegle, line 276: Nit: I usually name it `val`, and then you can write `condition val > 0`.

```
| group by [resource.project_id], [val: aggregate(value.human_accessed_secret)]
```

terraform/alerting/alerts.tf
sethvargo: Done

terraform/alerting/alerts.tf
sethvargo: Looks nicer your way 😄 . Done

sethvargo:
@yegle done, PTAL

whaught (approved):

Issue #1554, opened 2021-01-11
"bad realm" error should return 4XX error


bschlaman:
### TL;DR
If a realm doesn't have a jwks config, https://encv.org/jwks/{realm} returns a 500 error.

**Expected behavior**
Should return a 4XX error

sethvargo:
Thanks for finding this in the noise Brendan! And thanks for fixing it Weston!

PR #1557, opened 2021-01-11
Bad request for jwks invalid region

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1554

## Proposed Changes

* Give not-found or invalid-request for malformed or missing realm_id

**Release Note**

```release-note
not-found and invalid for malformed realm_id in jwks request
```

### DIFF
--- a/pkg/controller/jwks/jwks.go
+++ b/pkg/controller/jwks/jwks.go
@@ -19,10 +19,8 @@ package jwks
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
-	"strconv"
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/cache"
@@ -33,9 +31,6 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/rakutentech/jwk-go/jwk"
 )
-
-// Error codes
-var errBadRealm = errors.New("bad realm")
 
 // Controller holds all the pieces necessary to show the jwks encoded keys.
 type Controller struct {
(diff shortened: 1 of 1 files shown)


pkg/controller/jwks/jwks.go
```diff
@@ -67,18 +68,27 @@ func (c *Controller) HandleIndex() http.Handler {
...
-			if err != nil {
-				return nil, err
+		// Find the realm, and the key abstractions from the DB.
+		realm, err := c.getRealm(realmID)
+		if err != nil {
+			if database.IsNotFound(err) {
+				c.h.RenderJSON(w, http.StatusNotFound, fmt.Errorf("no realm exists for region %q", realmID))
+				return
+			}
+			if err == errBadRealm {
```
sethvargo, line 81: I think we can drop this - see my comment below.

pkg/controller/jwks/jwks.go
```diff
@@ -130,12 +140,7 @@ func (c *Controller) getRealm(realmStr string) (*database.Realm, error) {
 	if err != nil {
 		return nil, errBadRealm
 	}
-	var realm *database.Realm
```
sethvargo, line 133: The only way this returns errBadRealm now is if the string fails to parse as an int. But `FindRealm` shouldn't require an int. If it does, we should make that an interface and remove this function entirely and call c.db.FindRealm(..) in the main function.

sethvargo (changes requested):

pkg/controller/jwks/jwks.go
whaught: even easier!

sethvargo (approved):
/lgtm

I love me some negative diffs 😄

PR #1558, opened 2021-01-11
fix bulk_issue_test

whaught:
## Proposed Changes

* Assert the status 200 ok for the test

### DIFF
--- a/pkg/render/renderer_impl.go
+++ b/pkg/render/renderer_impl.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/internal/project"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/rbac"
 	"github.com/leonelquinteros/gotext"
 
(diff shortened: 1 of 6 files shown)


pkg/render/renderer_impl.go
```diff
@@ -323,13 +324,23 @@ func templateFuncs() htmltemplate.FuncMap {
...
+		"hasOne":           hasOne,
+		"hasMany":          hasMany,
 
 		"rbac": func() map[string]rbac.Permission {
 			return rbac.NamePermissionMap
 		},
 	}
 }
 
+func hasOne(a []*interface{}) bool {
```
sethvargo, line 336: Neither of these function signatures are generic enough unfortunately. You'll need to accept `interface{}` (`[]interface{}` is NOT the same as `[]T`, unfortunately), then use reflection to determine if it's an array. I already do this in another template helper, so you should be able to copy that or pull it into a helper.

sethvargo (changes requested):

pkg/render/renderer_impl.go
```diff
@@ -333,12 +332,14 @@ func templateFuncs() htmltemplate.FuncMap {
 	}
 }
 
-func hasOne(a []*interface{}) bool {
-	return len(a) == 1
+func hasOne(a interface{}) bool {
+	s := reflect.ValueOf(a)
+	return s.Len() == 1
```
sethvargo, line 337: Len will panic on non-array/slice/string types. Isn't reflection like the most funest?

sethvargo (approved):
/lgtm

PR #1559, opened 2021-01-11
Allow JWKs to work with region code as well

sethvargo:
**Release Note**

```release-note
Allow JWK lookup URLs to work with region codes in addition to IDs
```

### DIFF
--- a/pkg/controller/jwks/jwks.go
+++ b/pkg/controller/jwks/jwks.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/pkg/cache"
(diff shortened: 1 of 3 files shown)


pkg/controller/jwks/jwks.go
```diff
@@ -48,7 +49,7 @@ func (c *Controller) HandleIndex() http.Handler {
 		// key is the key in the cacher where the values for this JWK are cached.
 		key := &cache.Key{
 			Namespace: "jwks",
-			Key:       r.URL.String(),
+			Key:       strings.ToLower(r.URL.String()),
```
sethvargo, line 52: Optimizes caching

whaught (approved):

mikehelmick (approved):

sethvargo:
Well, I had to make it less elegant unfortunately.

sethvargo:
@whaught PTAL

whaught (approved):

PR #1561, opened 2021-01-12
allow for redirect cert changes without outages

mikehelmick:
## Proposed Changes

* Allow for adding new enx-redirect domains through additive SSL certs rather than causing an outage

**Release Note**

```release-note
ENX redirect, add new sub-domains without outage.
```

/hold


sethvargo (approved):
/lgtm

PR #1566, opened 2021-01-12
Delete users who are deleted

sethvargo:
**Release Note**

```release-note
Delete users who are deleted
```

/assign @mikehelmick

### DIFF
--- a/pkg/database/user.go
+++ b/pkg/database/user.go
@@ -418,7 +418,7 @@ func (db *Database) DeleteUser(u *User, actor Auditable) error {
 		}
 
 		// Delete the user
-		if err := tx.Delete(u).Error; err != nil {
+		if err := tx.Unscoped().Delete(u).Error; err != nil {
 			return fmt.Errorf("failed to save user: %w", err)
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/database/user.go
```diff
@@ -418,7 +418,7 @@ func (db *Database) DeleteUser(u *User, actor Auditable) error {
 		}
 
 		// Delete the user
-		if err := tx.Delete(u).Error; err != nil {
+		if err := tx.Unscoped().Delete(u).Error; err != nil {
```
mikehelmick, line 421: question - does this impact auditing?

I want to make sure that someone's trail of codes issued or audit logs things doesn't immediately disappear.

originally this was "disable user"

mikehelmick (changes requested):

pkg/database/user.go
sethvargo: No, audit logs are intentionally not tied back to the user via a FK or database-level link

sethvargo: It was disable user, but then the cleanup job _used_ to delete them. I think Brendan got caught in a weird place where he had performed the old action of "disable", but then the cleanup job was updated to focus on memberships and never deleted him.

mikehelmick (approved):

PR #1582, opened 2021-01-13
Save Realm into context

yegle:
This allows the realm id to be picked up by various metrics.

### DIFF
--- a/pkg/controller/middleware/auth.go
+++ b/pkg/controller/middleware/auth.go
@@ -122,6 +122,18 @@ func RequireAuth(cacher cache.Cacher, authProvider auth.Provider, db *database.D
 				return
 			}
 
+			// Insert realm id to context so metrics can pick it up.
+			realmID := controller.RealmIDFromSession(session)
+			// This is inserted by realm selector, so it could be 0 if no realm
+			// has been selected.
+			if realmID != 0 {
+				realm, err := db.FindRealm(realmID)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/middleware/auth.go
```diff
@@ -122,6 +122,18 @@ func RequireAuth(cacher cache.Cacher, authProvider auth.Provider, db *database.D
 				return
 			}
 
+			// Insert realm id to context so metrics can pick it up.
+			realmID := controller.RealmIDFromSession(session)
+			// This is inserted by realm selector, so it could be 0 if no realm
+			// has been selected.
+			if realmID != 0 {
+				realm, err := db.FindRealm(realmID)
```
mikehelmick, line 130: this should attempt to read the realm from the cache before going to DB

mikehelmick (changes requested):

sethvargo (commented):
This is already set by apikey.go. I think this belongs in membership.go around L85. We shouldn't need to lookup the realm if we do it there:

```
ctx = controller.WithRealm(ctx, membership.Realm)
```

pkg/controller/middleware/auth.go
sethvargo: You're right, but I think this is the wrong place to do this entirely. We should set this in membership.go instead, where the realm already exists with no lookup required.

sethvargo (approved):
/lgtm

Issue #1581, opened 2021-01-13
api/issue incomplete IssueCodeResponse


kantmakm:
### TL;DR
When programmatically sending to the api/issue endpoint our scripts have been getting back an errorCode per the api documentation.  That stopped with today's batch.  We have changed our code to look for 'error' instead and that seems to be working.

**Expected behavior**
IssueCodeResponse:

```json
{
  "uuid": "string UUID",
  "code": "short verification code",
  "expiresAt": "RFC1123 formatted string timestamp",
  "expiresAtTimestamp": 0,
  "expiresAt": "RFC1123 UTC timestamp",
  "expiresAtTimestamp": 0,
  "longExpiresAt": "RFC1123 UTC timestamp",
  "longExpiresAtTimestamp": 0,
  "error": "descriptive error message",
  "errorCode": "well defined error code from api.go",
}
```

**Observed behavior**
IssueCodeResponse

```json
{
  "uuid": "string UUID",
  "code": "short verification code",
  "expiresAt": "RFC1123 formatted string timestamp",
  "expiresAtTimestamp": 0,
  "expiresAt": "RFC1123 UTC timestamp",
  "expiresAtTimestamp": 0,
  "longExpiresAt": "RFC1123 UTC timestamp",
  "longExpiresAtTimestamp": 0,
  "error": "descriptive error message",
}
```

Python script used to work:

```
        elif response.status_code == 400:
            response_text = response.json()
            log_line_for_retry('400_error', row)
            logging.warning(f"400 error for CEDRS '{row[CEDRS_ID]}' phone number '{row[PHONE_NUMBER]}' status  '{response.status_code}' error '{response_text['    error']}' errorcode '{response_text['**errorCode**']}'. \U0001F44E ")
            continue
```

failure log:

```
Traceback (most recent call last):
  File "sendcodes_v2.py", line 190, in <module>
    logging.warning(f"400 error for CEDRS '{row[CEDRS_ID]}' phone number '{row[PHONE_NUMBER]}' status  '{response.status_code}' error '{response_text['error']}' errorcode '{response_text['errorCode']}'. \U0001F44E ")
KeyError: 'errorCode'
```

Code change to workaround missing errorCode

```
        elif response.status_code == 400:
            response_text = response.json()
            log_line_for_retry('400_error', row)
[...]

PR #1584, opened 2021-01-13
Marshal errors with *ErrorReturn type

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1581

## Proposed Changes

* Marshal json with ErrorReturn struct rather than the IssueResponse struct, although identical they differ in that ErrorReturn does not include `OmitEmpty`
* Add an error code `sms_failure` for generic Twilio errors

### DIFF
--- a/pkg/controller/issueapi/handle_issue.go
+++ b/pkg/controller/issueapi/handle_issue.go
@@ -113,7 +113,7 @@ func (c *Controller) decodeAndIssue(ctx context.Context, w http.ResponseWriter, 
 		c.h.RenderJSON(w, http.StatusOK, res.IssueCodeResponse())
 		return
 	default:
-		c.h.RenderJSON(w, result.HTTPCode, res.ErrorReturn)
+		c.h.RenderJSON(w, res.HTTPCode, res.ErrorReturn)
 		return
 	}
 }
--- a/pkg/controller/issueapi/handle_issue.go
+++ b/pkg/controller/issueapi/handle_issue.go
@@ -104,16 +104,16 @@ func (c *Controller) decodeAndIssue(ctx context.Context, w http.ResponseWriter, 
 
 	res := c.IssueOne(ctx, &request)
 	result.HTTPCode = res.HTTPCode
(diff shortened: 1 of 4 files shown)


pkg/controller/issueapi/handle_issue.go
```diff
@@ -104,16 +104,16 @@ func (c *Controller) decodeAndIssue(ctx context.Context, w http.ResponseWriter,
...
-	resp := res.IssueCodeResponse()
-	if resp.Error != "" {
-		if result.HTTPCode == http.StatusInternalServerError {
-			controller.InternalError(w, r, c.h, errors.New(resp.Error))
-			return
-		}
-		c.h.RenderJSON(w, result.HTTPCode, resp)
+
+	switch res.HTTPCode {
+	case http.StatusInternalServerError:
```
sethvargo, line 109: I'm not sure I follow why these cases behave differently?

mikehelmick:
can we add tests?

whaught:
> can we add tests?

That's actually difficult to do - we'd need a known case where there is no error code to assert that the key is there with empty-string after json marshal. In this case I've added a code for twilio failure so that's covered now.

pkg/controller/issueapi/handle_issue.go
whaught: before I was rendering `struct IssueCodeResponse` which has:
`ErrorCode string 'json:"errorCode,omitempty"'`

We put error and errorCode on the responses which perhaps help document the possible responses. In practice we respond with:
`type ErrorReturn struct`
which has
`ErrorCode string 'json:"errorCode"'` with no `omitempty`

whaught: the key difference here (other than the switch which I just thought looked better:
`c.h.RenderJSON(w, result.HTTPCode, resp)`
is now
`c.h.RenderJSON(w, result.HTTPCode, res.ErrorReturn)`

mikehelmick:
this causes the issue_tests to timeout.

mikehelmick (approved):

PR #1591, opened 2021-01-14
Set maxScale to 2.

yegle:
The limit was set to low so it doesn't cause too much spending when
running as a dev env. Set it to 1 will cause brief downtime in some edge
cases.

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -260,7 +260,7 @@ variable "enable_lb_logging" {
 
 locals {
   default_revision_annotations = {
-    "autoscaling.knative.dev/maxScale" : "1",
+    "autoscaling.knative.dev/maxScale" : "2",
     "run.googleapis.com/vpc-access-egress" : "private-ranges-only"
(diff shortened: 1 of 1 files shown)


mikehelmick (approved):

sethvargo:
@yegle @mikehelmick 

1. Reading the internal bug, it looks like this needs to be "greater than 3".
2. Some of these were set to `max-scale` because it's not safe to run multiple instances concurrently.

I'm not sure this is a good change.

mikehelmick:
There shouldn't be any services where concurrency > 1 isn't safe. If there is, then we need to add distributed locking like we do in the key server.

sethvargo:
I believe the following services are _not_ safe to run concurrently:

- `appsync`
- `e2e-runner`
- `migrate`
- `modeler`

That's about half our services...

mikehelmick:
modeler has correct locking
migrate isn't deployed a a service
e2e-runner is save for concurrent execution (race conditions have been fixed)

I'm looking at appsync

PR #1592, opened 2021-01-14
Revert "Set maxScale to 2. (#1591)"

yegle:
This reverts commit d9a46845e9cbc37549380f7147776c5bd28d45c1.

There are services that's not safe to run with more than 1 instances
concurrently. Will revert first and only set maxScale on services that's
safe to do.

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -260,7 +260,7 @@ variable "enable_lb_logging" {
 
 locals {
   default_revision_annotations = {
-    "autoscaling.knative.dev/maxScale" : "2",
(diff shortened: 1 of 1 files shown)


mikehelmick:
I don't think we need this

yegle:
There are concerns that some of the services doesn't work well with more than 1 instances. Is it still worth reverting and only set maxScale on `server` service?

sethvargo:
Mike and I audited the services. The only one that could have had an issue was appsync, and it's now fixed at HEAD. I think it's safe to

/close

this now.

PR #1593, opened 2021-01-14
add locking to appsync handler

mikehelmick:
## Proposed Changes

* ensure app sync doesn't run too frequently

**Release Note**

```release-note
App sync will run no more than once every 5 minutes, controlled via database.
```

### DIFF
--- a/pkg/config/appsync_server_config.go
+++ b/pkg/config/appsync_server_config.go
@@ -42,10 +42,10 @@ type AppSyncConfig struct {
 	RateLimit uint64 `env:"RATE_LIMIT,default=60"`
 
 	// AppSync config
-	AppSyncURL         string        `env:"APP_SYNC_URL"`
-	FileSizeLimitBytes int64         `env:"APP_SYNC_SIZE_LIMIT, default=64000"`
-	Timeout            time.Duration `env:"APP_SYNC_TIMEOUT, default=1m"`
-	AppSyncPeriod      time.Duration `env:"APP_SYNC_PERIOD, default=5m"`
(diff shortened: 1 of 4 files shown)


pkg/controller/appsync/appsync.go
```diff
@@ -42,6 +47,20 @@ func (c *Controller) HandleSync() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
+		if ok, err := c.shouldSync(ctx); err != nil {
```
sethvargo, line 50: Nit: this could be easier to read without the `else`

ok, err := c.shoudlSync()
...

if !ok {
  429
}

sethvargo (commented):
We should consider renaming CleanupStatus if we're gonna repurpose it this way.

pkg/config/appsync_server_config.go
```diff
@@ -45,6 +45,7 @@ type AppSyncConfig struct {
 	AppSyncURL         string        `env:"APP_SYNC_URL"`
 	FileSizeLimitBytes int64         `env:"APP_SYNC_SIZE_LIMIT, default=64000"`
 	Timeout            time.Duration `env:"APP_SYNC_TIMEOUT, default=1m"`
+	AppSyncPeriod      time.Duration `env:"APP_SYNC_PERIOD, default=5m"`
```
whaught, line 48: Nit: this is really `AppSyncMinimumPeriod` since it's scheduled externally. I wouldn't want to confuse future readers. Maybe a comment or rename?

whaught (approved):

Issue #1567, opened 2021-01-12
Proposal: Automatically rotate verification certificate signing keys


mikehelmick:
### TL;DR

Automatically rotate verification certificate signing keys

### Design

**Proposal**

1. Realm opt in to auto rotate keys ever 30 days (they can still do it manually)

   * Big warning that your key server should be using JWKS import

2. New periodic background job to rotate keys

   * 30 days after last key was created - create a new key
   * 12 hours later - make that new key active
   * 1 hour later - revoke the old key

mikehelmick:
- [ ] add db columns
- [ ] add UI to enable disable / flag controlled
- [ ] add background rotation
- [ ] add terraform for scheduling
- [ ] enable auto rotation by default (v0.21.0 or later)

PR #1594, opened 2021-01-14
add database columns and config for feature flag

mikehelmick:
forst towards: #1567

## Proposed Changes

* Add DB column for realm auto verification certificate key rotation
* Add config feature flag that will control this - default to OFF at a system level

**Release Note**

```release-note
Adds database column for realm level auto key rotation settings.
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -1971,11 +1971,13 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 		{
 			ID: "00086-AddRealmAutoRotateSetting",
 			Migrate: func(tx *gorm.DB) error {
-				return tx.AutoMigrate(&Realm{}).Error
+				return multiExec(tx,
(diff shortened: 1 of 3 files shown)


pkg/database/migrations.go
```diff
@@ -1968,6 +1968,16 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`DELETE FROM users WHERE deleted_at IS NOT NULL`)
 			},
 		},
+		{
+			ID: "00086-AddRealmAutoRotateSetting",
+			Migrate: func(tx *gorm.DB) error {
+				return tx.AutoMigrate(&Realm{}).Error
```
whaught, line 1974: if it's just the one bool column, maybe we should just do the standard `ALTER TABLE` dance. I feel like automigrate has screwed us up before

whaught: Also do we want the column to be NOT NULL?

whaught (approved):

pkg/database/migrations.go
mikehelmick: removed auto migrate and added not null.

whaught (approved):

Issue #1569, opened 2021-01-12
Proposal: automatically rotate verification token keys


mikehelmick:
### TL;DR

Currently, specific token signing key versions are managed through terraform.
While the ENV VAR accepts a list, allowing for human scale rotation, it would be better to automate this.

### Design

**Proposal**

Change TokenSigningConfig to accept a key and not a key version. For backwards compatibility, accepting a key version and normalizing it back to the key would be good.

Introduce DB table to track token signing key rotation.

Periodically rotate the signing key.

1. Introduce and start using key version n+1
2. Wait > 24 hours (how long verification tokens are valid for)
3. Destroy key version n

**Alternatives considered**

Manual rotation via manual version creation and terraform config update.

The downside being this is a human process and prone to forget-fullness and errors.

PR #1597, opened 2021-01-14
Add rotation harness and rotate token signing keys

sethvargo:
This adds a new service, rotation, which handles key and rotation events. It's designed to be invoked via Cloud Scheduler and has its own locking and time intervals. It's conceptually similar to the cleanup job.

This also adds the database schema and example rotation for token signing keys. Note that this code is NOT in use in main code paths yet.

There's some TODOs in the code assigned to me. I intend to do those in follow-up PRs where all of this will be wired together.

Part of https://github.com/google/exposure-notifications-verification-server/issues/1569

**Release Note**

```release-note
Add rotation service and schema for rotating token signing keys
```

/assign @mikehelmick

### DIFF
--- a/pkg/config/cleanup_server_config.go
+++ b/pkg/config/cleanup_server_config.go
@@ -31,20 +31,28 @@ type CleanupConfig struct {
 	Database      database.Config
 	Observability observability.Config
 
+	// TokenSigning is the token signing configuration to purge old keys in the
+	// key manager when they are cleaned.
+	TokenSigning TokenSigningConfig
+
 	// DevMode produces additional debugging information. Do not enable in
 	// production environments.
 	DevMode bool `env:"DEV_MODE"`
 
+	// Port is the port on which to bind.
 	Port string `env:"PORT,default=8080"`
-
-	RateLimit uint64 `env:"RATE_LIMIT,default=60"`
 
 	// Cleanup config
 	AuditEntryMaxAge    time.Duration `env:"AUDIT_ENTRY_MAX_AGE, default=720h"`
 	AuthorizedAppMaxAge time.Duration `env:"AUTHORIZED_APP_MAX_AGE, default=336h"`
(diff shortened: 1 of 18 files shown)


sethvargo:
/assign @whaught 

It'd be good to get 2 LGTMs on this

pkg/config/cleanup_server_config.go
```diff
@@ -31,20 +31,28 @@ type CleanupConfig struct {
...
+	TokenSigning TokenSigningConfig
+
 	// DevMode produces additional debugging information. Do not enable in
 	// production environments.
 	DevMode bool `env:"DEV_MODE"`
 
+	// Port is the port on which to bind.
 	Port string `env:"PORT,default=8080"`
 
-	RateLimit uint64 `env:"RATE_LIMIT,default=60"`
```
sethvargo, line 40: This was unused

pkg/controller/cleanup/cleanup.go
```diff
@@ -93,6 +96,7 @@ func (c *Controller) HandleCleanup() http.Handler {
 				OK:     false,
 				Errors: []error{fmt.Errorf("too early")},
 			})
+			return
```
sethvargo, line 99: This is an existing nasty bug in the cleanup server.

whaught (commented):
Don't we need an entry in builders/build.yaml to be able to create /sevice_rotation.tf?

sethvargo:
> Don't we need an entry in builders/build.yaml to be able to create /sevice_rotation.tf?

Done

pkg/controller/cleanup/cleanup.go
mikehelmick: good catch

whaught (approved):

mikehelmick (approved):

Issue #1512, opened 2021-01-05
PHA stats: retrieve, store, and display stats from the key sever


mikehelmick:
### TL;DR

The key sever now collects metrics on publish requests and has an API for retrieval of that information
  - https://github.com/google/exposure-notifications-server/pull/1259

### Design

**Proposal**

* This must be opt-in (default off). There must be a disclaimer that this should be covered in the privacy policy for the PHAs app in terms of operational metrics that may be collected sever side and displayed in aggregate only, not tied to any individual.

* New setting for the key server audience value and key sever to retrieve from. These should be system level setting w/ realm level override available (only needed in some scenarios)

* New periodic job to update stats for all realms that have this enabled. Once an hour, all enabled realms should mint the appropriate JWT (using the active verification certificate key) and call the `/v1/stats` API on the key server. Any data retrieved should replace what is in local storage. The verification server is however free to retain data longer (up to 30d).

* Display of the new information to PHA authorized users
  * total number of publish request should be graphed w/ codes issued and claimed
  * histogram of oldest tek
  * histogram of onset to upload

/assign @sethvargo

whaught:
I'm going to call this done. We may have specific updates/tweaks to make, new stats to add, etc, but it will be easier to track those under more specific issues.

/close

PR #1599, opened 2021-01-15
DB model for storing key server stats

whaught:

## Proposed Changes

*  Key Server Stats database model
* Realm settings UI to set or delete the value

**Release Note**

```release-note
New tables for storing key-server statistics
```

### DIFF
--- a/cmd/server/assets/realmadmin/_form_general.html
+++ b/cmd/server/assets/realmadmin/_form_general.html
@@ -2,6 +2,7 @@
 
 {{$realm := .realm}}
 {{$testTypes := .testTypes}}
+{{$statsConfig := .statsConfig}}
 
 <p class="mb-4">
   These are common settings that apply to all of {{$realm.Name}}.
@@ -52,6 +53,41 @@
       This field supports the common <a
       href="https://daringfireball.net/projects/markdown/syntax">markdown</a>
       standard.
+    </small>
+  </div>
+
+  <hr>
+  <div class="form-group">
+    <div class="form-check">
+      <input type="checkbox" name="allow_key_server_stats" id="allow-key-server-stats" class="form-check-input"
+        value="true" {{checkedIf (ne $statsConfig.RealmID 0)}} />
+      <label for="allow-bulk-true" class="form-check-label">
+        Enable key-server statistics
+        <small class="form-text text-muted mb-3">
+          Checking this box will allow this realm to collect statistics from the key-server about codes
+          that this realm has issued.
+        </small>
+      </label>
+    </div>
+  </div>
+
+  <div class="form-label-group">
+    <input type="text" name="key_server_url" id="key-server-url" class="form-control{{if $statsConfig.ErrorsFor "keyServerURLOverride"}} is-invalid{{end}}"
+      value="{{$statsConfig.KeyServerURLOverride}}" placeholder="Key server"/>
+    <label for="name">Key-server URL override</label>
+    {{template "errorable" $statsConfig.ErrorsFor "keyServerURLOverride"}}
+    <small class="form-text text-muted">
+      A key-server url to gather statistics from. This overrides the default value of the server and should be left empty to use the default.
+    </small>
+  </div>
+
+  <div class="form-label-group">
+    <input type="text" name="key_server_audience" id="key-server-audience" class="form-control{{if $statsConfig.ErrorsFor "keyServerAudienceOverride"}} is-invalid{{end}}"
+      value="{{$statsConfig.KeyServerAudienceOverride}}" placeholder="Audience"/>
+    <label for="name">Audience override</label>
+    {{template "errorable" $statsConfig.ErrorsFor "keyServerAudienceOverride"}}
+    <small class="form-text text-muted">
+      The key-server audience. This overrides the default value of the server and should be left empty to use the default.
     </small>
   </div>
 
--- a/pkg/database/key_server_stats.go
+++ b/pkg/database/key_server_stats.go
@@ -1,0 +1,106 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"time"
+
+	"github.com/jinzhu/gorm/dialects/postgres"
+	"github.com/lib/pq"
+)
+
+// KeyServerStats represents statistics for a key-server for this realm
+type KeyServerStats struct {
+	Errorable
+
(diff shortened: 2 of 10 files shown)


pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+	// IsSystem determines if this is a system-level stats configuration. There can
+	// only be one system-level stats configuration.
+	IsSystem bool `gorm:"column:is_system; type:bool; not null; default:false;"`
+
+	// KeyServerURL allows a realm to override the system's URL with its own
+	KeyServerURL string `gorm:"column:key_server_url; type:varchar(150); default: '';"`
+	// KeyServerAudience allows a realm to override the system's audience
+	KeyServerAudience string `gorm:"column:key_server_audience; type:varchar(150); default: '';"`
+
+	Stats []*KeyServerStatsDay `gorm:"PRELOAD:false; SAVE_ASSOCIATIONS:false; ASSOCIATION_AUTOUPDATE:false, ASSOCIATION_SAVE_REFERENCE:false;"`
```
sethvargo, line 40: Don't do this, gorm will make you have a bad day. The way we've done this in other models is to have a public method on the parent struct that accepts a database.Database as the first parameter and runs the query.

```go
func (s *KeyServerStats) DailyStats(db *Database) ... {
}
```

pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// RealmId that these stats belong to.
+	RealmID uint `gorm:"column:realm_id; primary_key; type:integer; not null;"`
+
+	// Day will be set to midnight UTC of the day represented. An individual day
+	// isn't released until there is a minimum threshold for updates has been met.
+	Day time.Time `gorm:"column:day; primary_key;"`
+
+	// PublishRequests is a count of requests per OS
+	PublishRequests postgres.Hstore `gorm:"column:publish_requests; type:hstore;"`
```
sethvargo, line 55: I think using hstore is going to make some future queries wrong. It's possible to query hstore fields, but I don't know how difficult that is in gorm:

```sql
SELECT publish_requests -> 'ios' AS apple
```

pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+	// The count at index 0-29 represents the number of uploads with that symptom onset age.
+	// Index 30 represents > 29 days.
+	OnsetToUploadDistribution pq.Int64Array `gorm:"column:onset_to_upload_distribution; type:bigint[];"`
+
+	// RequestsMissingOnsetDate is the number of publish requests where no onset date
+	// was provided. These request are not included in the onset to upload distribution.
+	RequestsMissingOnsetDate int64 `gorm:"column:request_missing_onset_date; type:bigint; not null; default: 0;"`
+}
+
+func (db *Database) GetKeyServerStats(realmID uint) (*KeyServerStats, error) {
```
sethvargo, line 77: Doc

sethvargo: (everywhere)

sethvargo, line 77: This should probably hang off the realm instead:

```go
func (r *Realm) KeyServerStats(db *Database) ... {
}
```

pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+		Where("realm_id = ?", realmID).
+		First(&stats).
+		Error; err != nil {
+		return nil, err
+	}
+	return &stats, nil
+}
+
+func (db *Database) SaveKeyServerStats(stats *KeyServerStats) error {
+	return db.db.Save(stats).Error
```
sethvargo, line 89: Probably want some BeforeSave validation to make sure at least RealmID is not empty.

pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (db *Database) GetKeyServerStatsDay(realmID uint, day time.Time) (*KeyServerStatsDay, error) {
```
sethvargo, line 92: I'm not sure you'll ever want the stats for just a single day. Even if you did, it's probably better for this to return all stats for the past N days as an ordered slice.

sethvargo (changes requested):

pkg/database/key_server_stats.go
mikehelmick: +1 to that - it should also limit it to the past 30d (time period agreed upon for retention)

whaught: Well this was just testing the schema - its likely we'll want a query in its real use. The 30d limit could be on such a query but we'll also want it to purge on the cleanup job so we aren't retaining in the first place.

pkg/database/key_server_stats.go
whaught: yea but then you need to have actually loaded the realm which may not be the case

pkg/database/key_server_stats.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// RealmId that these stats belong to.
+	RealmID uint `gorm:"column:realm_id; primary_key; type:integer; not null;"`
+
+	// IsSystem determines if this is a system-level stats configuration. There can
+	// only be one system-level stats configuration.
+	IsSystem bool `gorm:"column:is_system; type:bool; not null; default:false;"`
+
+	// KeyServerURL allows a realm to override the system's URL with its own
+	KeyServerURL string `gorm:"column:key_server_url; type:varchar(150); default: '';"`
```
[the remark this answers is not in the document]
whaught: the reason not to is because that's where if IsSystem == true

pkg/database/migrations.go
```diff
@@ -1999,6 +1999,19 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`DROP TABLE IF EXISTS token_signing_keys`)
 			},
 		},
+		{
+			ID: "00088-KeyServerStats",
+			Migrate: func(tx *gorm.DB) error {
+				err := tx.AutoMigrate(&KeyServerStats{}, &KeyServerStatsDay{}).Error
```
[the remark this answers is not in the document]
whaught: *sigh* I'll do it

cmd/server/assets/realmadmin/_form_general.html
```diff
@@ -55,6 +56,41 @@
...
 
+  <hr>
+  <div class="form-group">
+    <div class="form-check">
+      <input type="checkbox" name="allow_key_server_stats" id="allow-key-server-stats" class="form-check-input"
+        value="true" {{checkedIf (ne $statsConfig.RealmID 0)}} />
+      <label for="allow-bulk-true" class="form-check-label">
+        Enable key-server statistics
+        <small class="form-text text-muted mb-3">
+          Checking this box will allow this realm to collect statistics from the key-server about codes
```
mikehelmick, line 67: there needs to be a note here about how this may need to be disclosed your applications privacy policy.

we should also describe what's here -onset to upload, uploads by operating system.

cmd/server/assets/realmadmin/_form_general.html
```diff
@@ -55,6 +56,41 @@
     </small>
   </div>
 
+  <hr>
```
mikehelmick, line 59: let's flag disable this until the visualizations are ready

mikehelmick:
/lgtm
/hold
feel free to address my 2 comments in a follow up

whaught:
/unhold

Will follow up with comments

PR #1602, opened 2021-01-15
Source token signing keys from the database

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1569

**Release Note**

```release-note
**Potentially breaking!** Source token signing keys from the database. This completes the move of system token signing keys from environment variables to the database. This change attempts to be backward compatible, but server are encouraged to test changes in an isolated environment before upgrading production systems.
```

/assign @mikehelmick

### DIFF
--- a/cmd/server/assets/admin/info.html
+++ b/cmd/server/assets/admin/info.html
@@ -19,8 +19,23 @@
           <dt>Build ID</dt>
           <dd>{{.buildID}}</dd>
 
-          <dt>Build Tag</dt>
+          <dt>Build tag</dt>
           <dd>{{.buildTag}}</dd>
+
+          <dt>Token signing key</dt>
+          {{if $key := .activeTokenSigningKey}}
+            <dd class="mb-0">
+              kid: <span class="text-monospace user-select-all">{{$key.UUID}}</span>
+            </dd>
+            <dd class="mb-0">
+              key: <span class="text-monospace user-select-all">{{$key.KeyVersionID}}</span>
+            </dd>
+            <dd>
+              created: <span class="text-monospace user-select-all" data-timestamp="{{$key.CreatedAt.Format "1/02/2006 3:04:05 PM UTC"}}">{{$key.CreatedAt.Format "2006-02-01 15:04"}}</span>
+            </dd>
+          {{else}}
+            <dd>not configured</dd>
+          {{end}}
         </dl>
       </div>
     </div>
--- a/cmd/server/assets/admin/info.html
+++ b/cmd/server/assets/admin/info.html
@@ -22,16 +22,39 @@
           <dt>Build tag</dt>
           <dd>{{.buildTag}}</dd>
 
-          <dt>Token signing key</dt>
-          {{if $key := .activeTokenSigningKey}}
-            <dd class="mb-0">
-              kid: <span class="text-monospace user-select-all">{{$key.UUID}}</span>
-            </dd>
-            <dd class="mb-0">
-              key: <span class="text-monospace user-select-all">{{$key.KeyVersionID}}</span>
-            </dd>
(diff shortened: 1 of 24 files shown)


pkg/controller/rotation/handle_rotate.go
```diff
@@ -65,19 +66,19 @@ func (c *Controller) HandleRotate() http.Handler {
...
-			if err != nil {
+			if err != nil && !database.IsNotFound(err) {
 				merr = multierror.Append(merr, fmt.Errorf("failed to lookup existing signing key: %w", err))
 				result = observability.ResultError("FAILED")
 				return
 			}
-			if age, max := time.Now().UTC().Sub(existing.CreatedAt), c.config.TokenSigningKeyMaxAge; age < max {
-				logger.Debugw("token signing key does not require rotation", "age", age, "max", max)
-				return
+			if existing != nil {
```
sethvargo, line 74: This allows the rotator to create the first key if it does not exist

internal/routes/apiserver.go
```diff
@@ -49,6 +49,23 @@ func APIServer(
...
+	if _, err := db.ActiveTokenSigningKey(); err != nil {
+		if !database.IsNotFound(err) {
+			return nil, closer, fmt.Errorf("failed to lookup active token signing key: %w", err)
+		}
+
+		tokenSignerTyp, ok := tokenSigner.(keys.SigningKeyManager)
+		if !ok {
+			return nil, closer, fmt.Errorf("token signing key manager is not a signing key manager (is %T)", tokenSigner)
+		}
+		if _, err := db.RotateTokenSigningKey(ctx, tokenSignerTyp, cfg.TokenSigning.ParentKeyName(), database.System); err != nil {
```
mikehelmick, line 64: I think RotateTokenSigningKey needs some kind of locking mechanism (distributed).

It's possible that multiple container instances would try and do this at the same time.

sethvargo: Chatted about this sync. We're going to allow the rotation service to bootstrap the initial key.

sethvargo:
@mikehelmick updated, PTAL

cmd/server/assets/admin/info.html
```diff
@@ -19,8 +19,23 @@
           <dt>Build ID</dt>
           <dd>{{.buildID}}</dd>
 
-          <dt>Build Tag</dt>
+          <dt>Build tag</dt>
           <dd>{{.buildTag}}</dd>
+
+          <dt>Token signing key</dt>
```
mikehelmick, line 25: can we display the allowed keys as well - there should be a 24h + overlap where more than one key is valid?

pkg/config/token_signing_config.go
```diff
@@ -25,33 +26,79 @@ import (
...
 
-	TokenSigningKeys   []string `env:"TOKEN_SIGNING_KEY, required"`
+	// TokenSigningKeys is the parent token signing key (not the actual signing
+	// version). It is an array for backwards-compatibility, but in practice it
+	// should only have one element.
+	//
+	// Previously it was a list of all possible signing key versions, but those
+	// have moved into the database.
+	//
+	// TODO(sethvargo): Convert to string in 0.22+.
```
mikehelmick, line 38: we'll have to tackle the terraform changes as well...

cmd/server/assets/admin/info.html
sethvargo: Done

pkg/config/token_signing_config.go
sethvargo: They aren't actually that bad. This is the only change.

sethvargo:
@mikehelmick PTAL

mikehelmick (approved):

PR #1604, opened 2021-01-15
Return a success response on remote workers when no work is done

sethvargo:
Cloud Scheduler considers a 429 to be a "failure" and thus sets off our alerting. I think it also makes sense for the background jobs to return a successful status response in this case, because the request was wholly processed.

**Release Note**

```release-note
Return a success response on remote workers when no work is done
```

/assign @whaught


sethvargo:
@mikehelmick FYI

whaught (approved):

PR #1606, opened 2021-01-15
Restore user re-created metric and alert

sethvargo:
**Release Note**

```release-note
Restore user re-created metric and alert
```


pkg/controller/metrics.go
```diff
@@ -23,20 +23,20 @@ import (
 	"go.opencensus.io/stats/view"
 )
 
-const metricPrefix = observability.MetricRoot + "/server/login"
-
-var (
-	MFirebaseRecreates = stats.Int64(metricPrefix+"/fb_recreates", "firebase user recreates", stats.UnitDimensionless)
```
sethvargo, line 29: This appears to be an old, unused metric, so I removed it.

whaught: It looks like the behavior was removed in: https://github.com/google/exposure-notifications-verification-server/pull/902

Can we leave the metric - I think we want to instead keep that functionality

sethvargo: We have no associated alert for the metric. I can add it back in, but do you also want to create an alert for it?

whaught: https://github.com/google/exposure-notifications-verification-server/pull/1607 to bring it back.

We don't need an alert for it. I wanted it there to see (later) if this was a common problem, but it allows users to get unstuck in this situation

pkg/controller/healthz.go
```diff
@@ -53,7 +56,9 @@ func HandleHealthz(db *database.Database, h render.Renderer) http.Handler {
 				return
 			}
 		case "alerts":
-			// TODO(ych): fire a metric and configure an alert
+			// Alerts increments a metric that fires an alert after some interval. It
```
mikehelmick, line 59: I'm not convinced we should do this - because those endpoints are currently accessible to the public through the load balancer.

Anyone can page us at will...

mikehelmick (changes requested):

pkg/controller/healthz.go
sethvargo: Removed this and updated the OR to reflect the correct description.

sethvargo:
@whaught PTAL

whaught (approved):

PR #1607, opened 2021-01-15
Create firebase user if not exists

whaught:
## Proposed Changes

* When a user tries to reset their password for an account with no firebase, re-create the firebase user.
    * We've seen this a few times for users who are invited, but never accept the invitation and try to log-in much later
    *  I believe this behavior was lost in https://github.com/google/exposure-notifications-verification-server/pull/902

### DIFF
--- a/pkg/controller/login/reset_password.go
+++ b/pkg/controller/login/reset_password.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/internal/auth"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"go.opencensus.io/stats"
 )
 
 func (c *Controller) HandleShowResetPassword() http.Handler {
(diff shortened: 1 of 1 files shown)


pkg/controller/login/reset_password.go
```diff
@@ -73,8 +74,14 @@ func (c *Controller) HandleSubmitResetPassword() http.Handler {
 				return
 			}
 
-			controller.InternalError(w, r, c.h, err)
-			return
+			// Re-create the firebase user if it does not exist.
+			// We have seen this disappear if a user never logs-on after being invited
+			if created, _ := c.authProvider.CreateUser(ctx, user.Name, user.Email, "", false, nil); created {
+				stats.Record(ctx, controller.MFirebaseRecreates.M(1))
```
sethvargo, line 80: This will fire in dev, so I don't think it's what we want. I'll add something to the other PR

mikehelmick: we probably don't want an alert on single instances, but if we see a rash over these over a time period, that seems noteworthy

whaught: I don't think we alert on this metric?

whaught: but also feel free to cannibalize this change

sethvargo: I see. We should move this into the login controller's metrics then. I can do that as part of the other PR if that's easier.

sethvargo: Pulled into https://github.com/google/exposure-notifications-verification-server/pull/1606

/close

PR #1608, opened 2021-01-15
Add db lock to stats-puller service. Share 'Should<servicename>' func

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1512

## Proposed Changes

* Adds the db lock to the stats-puller service
* Shares the `SyncFoo` function each of these has defined
* Renames `Cleanup` to `Lock` except for CleanupStatus - that's in the db schema

**Release Note**

```release-note
Add locking to stats-puller service. Refactor cleanup -> lock
```

PR #1609, opened 2021-01-15
Put migration name back

whaught:
## Proposed Changes

* This was a find/replace mistake in https://github.com/google/exposure-notifications-verification-server/pull/1608. We don't want to change already-run migration IDs

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -90,7 +90,7 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


mikehelmick (approved):

sethvargo:
This is fine, but gorm only looks at the numbers, not the names iirc. That's why we have the test for strictly increasing numbres.

whaught:
> This is fine, but gorm only looks at the numbers, not the names iirc. That's why we have the test for strictly increasing numbres.

ah. I wasn't sure - figured better to leave it as it was

PR #1614, opened 2021-01-16
Add verification certificate key auto rotation

### LINKED ISSUE #1567: Proposal: Automatically rotate verification certificate signing keys

mikehelmick:
Fixes #1567

## Proposed Changes

* UI to enable/disable 
* Auto rotate job

New enable screen
image

Confirmation dialog
image

Status when enabled
image

New keys are created in inactive state to allow for propagation of public key
image

**Release Note**

```release-note
Allow for realm admins to enable automatic verification certificate key rotation. Keys are rotated every 30 days, and given 1h soak time for upstream key servers to pull public keys via public key discovery.
```


pkg/controller/rotation/handle_rotate.go
```diff
@@ -41,7 +41,7 @@ func (c *Controller) HandleRotate() http.Handler {
 
 		var merr *multierror.Error
```
whaught, line 43: should this file be handle_token_rotate and this function HandleTokenRotate

pkg/controller/rotation/handle_verification_rotate.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/hashicorp/go-multierror"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+)
+
+// HandleVerificationRotate handles verification certificate key rotation.
+func (c *Controller) HandleVerificationRotate() http.Handler {
+	type Result struct {
+		OK     bool    `json:"ok"`
+		Errors []error `json:"errors,omitempty"`
```
whaught, line 35: wait can we render `[]error` rather than `[]string`. I think Seth just did something with this

pkg/controller/rotation/rotation.go
```diff
@@ -22,7 +22,10 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
 )
 
-const rotationLock = "rotation"
+const (
+	tokenRotationLock        = "rotation"
```
whaught, line 26:
```suggestion
	tokenRotationLock        = "tokenRotationLock"
```

terraform/service_rotation.tf
```diff
@@ -223,3 +223,30 @@ resource "google_cloud_scheduler_job" "rotation-worker" {
     google_project_service.services["cloudscheduler.googleapis.com"],
   ]
 }
+
+resource "google_cloud_scheduler_job" "realm-key-rotation-worker" {
+  name             = "realm-key-rotation-worker"
+  region           = var.cloudscheduler_location
+  schedule         = "2,32 * * * *"
```
whaught, line 230: 2,32... those are oddly specific minutes

pkg/controller/rotation/handle_verification_rotate.go
whaught: https://github.com/google/exposure-notifications-verification-server/pull/1613

mikehelmick: yeah, that needs a rebase...

terraform/service_rotation.tf
mikehelmick: offsetting it from the token rotation (which is every 5 minutes)

whaught (approved):

pkg/database/database_util.go
```diff
@@ -277,6 +277,8 @@ func (i *TestInstance) NewDatabase(tb testing.TB, cacher cache.Cacher) (*Databas
 		tb.Fatalf("failed to load database configuration: %s", err)
 	}
 	db.keyManager = keys.TestKeyManager(tb)
+	db.config.CertificateSigningKeyRing = "certificates"
```
sethvargo, line 280: This introduced a data race

PR #1620, opened 2021-01-16
Build all cmd binaries at once

sethvargo:
This should dramatically improve our build times

**Release Note**

```release-note
Build all cmd binaries at once in CI
```


mikehelmick (approved):

whaught (approved):
Wow this seems simpler

PR #1621, opened 2021-01-16
Retrieve stats from key-server

whaught:

## Proposed Changes

* Client for the key-server stats endpoint
* Logic for composing the `Bearer` auth token
* Body of stats-puller calls Key server and stores the retrieved stats

**Release Note**

```release-note
Logic for executing the v1/stats request and storing results
```


pkg/controller/statspuller/statspull.go
```diff
@@ -16,23 +16,35 @@
...
+	h                      render.Renderer
 }
 
 // New creates a new stats-pull controller.
-func New(cfg *config.StatsPullerConfig, db *database.Database, h render.Renderer) *Controller {
-	return &Controller{
-		config: cfg,
-		db:     db,
-		h:      h,
+func New(cfg *config.StatsPullerConfig, db *database.Database, h render.Renderer) (*Controller, error) {
```
sethvargo, line 34: I’d recommend passing the client into the controller instead of relying on the controller to build from configuration. It’s a nicer separation, esp for testing, and it removes the need to return an error

pkg/controller/statspuller/handle_pull.go
```diff
@@ -53,6 +60,84 @@ func (c *Controller) HandlePullStats() http.Handler {
...
+			claims := &jwt.StandardClaims{
+				Audience:  audience,
+				ExpiresAt: now.Add(time.Minute).Unix(),
+				IssuedAt:  now.Unix(),
+				Issuer:    s.Issuer,
+			}
+			token := jwt.NewWithClaims(jwt.SigningMethodES256, claims)
+			token.Header["kid"] = s.KeyID
+
+			signedJWT, err := jwthelper.SignJWT(token, s.Signer)
```
whaught, line 108: @mikehelmick can you look through the signing stuff I've got here

pkg/controller/statspuller/signer.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+)
+
+type SignerInfo struct {
+	Signer crypto.Signer
+	KeyID  string
+	Issuer string
+}
+
+func (c *Controller) getSignerForRealm(ctx context.Context, realmID uint) (*SignerInfo, error) {
```
sethvargo, line 29: I think we need to optimize this function a bit and consume the public key cache.

pkg/controller/statspuller/signer.go
```diff
@@ -0,0 +1,75 @@
...
```
mikehelmick, line 29: seems like this could be shared w/ the cert api?

whaught: absolutely - it's currently nearly straight a cut/paste. I'm just playing to get this to work

mikehelmick (approved):

PR #1622, opened 2021-01-16
Keep schedules all in UTC timezone

whaught:
## Proposed Changes

* Put all the service schedules in UTC timezone
* Have the daily task run around the same time
    * appsync run at 11am instead of midnight is a better time for oncallers

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -387,7 +387,7 @@ resource "google_cloud_scheduler_job" "backup-database-worker" {
   name             = "backup-database-worker"
   region           = var.cloudscheduler_location
   schedule         = var.database_backup_schedule
-  time_zone        = "America/Los_Angeles"
+  time_zone        = "UTC"
   attempt_deadline = "1800s"
 
   retry_config {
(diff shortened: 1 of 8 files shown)


yegle:
Hmm, cron jobs should always use timezones that doesn't do daylight save...

Also can we make this configurable? People who host outside of US may have their preferred timezones.

whaught:
> Hmm, cron jobs should always use timezones that doesn't do daylight save...
> 
> Also can we make this configurable? People who host outside of US may have their preferred timezones.

We can go the other way and make them all UTC

mikehelmick:
Did you test this out, I feel like there was something weird with cloud scheduler time zones

whaught:
> Did you test this out, I feel like there was something weird with cloud scheduler time zones

No but @yegle is right and this is what Cloud Scheduler reccomends: https://cloud.google.com/scheduler/docs/configuring/cron-job-schedules#daylight_savings_time

mikehelmick (approved):

sethvargo (commented):
Wait, this doesn’t work. We had everything at UTC and it caused a perpetual diff. Also, if you try to look at the scheduler in the UI, it complains that it’s an invalid timezone. 

@whaught

PR #1624, opened 2021-01-16
legacy signing fixes

### LINKED ISSUE #1623: tokens signed w/ legacy key fail to validate if no DB keys exist`
/assign

mikehelmick:
Fixes #1623

## Proposed Changes

* allow for tokens to be signed w/ legacy keys until DB one is present
   * can be disabled by flag
   * to be removed in v0.21
* fix bug where a not found on DB backed key lookup returned error instead of validating w/ legacy key
* debug logging by default in some tools - makes life easier

**Release Note**

```release-note
Allow for legacy signing key config for tokens to be used during the upgrade to DB backed tokens.
```

### DIFF
--- a/pkg/controller/certapi/certapi.go
+++ b/pkg/controller/certapi/certapi.go
@@ -95,8 +95,9 @@ func (c *Controller) validateToken(ctx context.Context, verToken string) (string
 					return nil, fmt.Errorf("no key corresponds to kid %q", kid)
 				}
 				tokenSigningKey = &database.TokenSigningKey{KeyVersionID: keyID}
+			} else {
+				return nil, fmt.Errorf("failed to lookup token signing key: %w", err)
 			}
-			return nil, fmt.Errorf("failed to lookup token signing key: %w", err)
 		}
 
 		publicKey, err := c.pubKeyCache.GetPublicKey(ctx, tokenSigningKey.KeyVersionID, c.kms)
--- a/tools/get-certificate/main.go
+++ b/tools/get-certificate/main.go
@@ -18,6 +18,7 @@ package main
 import (
 	"context"
 	"flag"
+	"os"
 	"time"
 
 	"github.com/google/exposure-notifications-verification-server/internal/clients"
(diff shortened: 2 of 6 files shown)


pkg/controller/certapi/certapi.go
```diff
@@ -95,8 +95,9 @@ func (c *Controller) validateToken(ctx context.Context, verToken string) (string
 					return nil, fmt.Errorf("no key corresponds to kid %q", kid)
 				}
 				tokenSigningKey = &database.TokenSigningKey{KeyVersionID: keyID}
+			} else {
+				return nil, fmt.Errorf("failed to lookup token signing key: %w", err)
```
mikehelmick, line 99: this was the bug - it filled in legacy key ID, but then returned an error.

tools/get-certificate/main.go
```diff
@@ -41,6 +42,9 @@ func main() {
 
 	ctx, done := signalcontext.OnInterrupt()
 
+	if os.Getenv("LOG_LEVEL") == "" {
```
sethvargo, line 45: It’d be better just to use the debugger logger instead of setting the env. Just don’t use logging.NewFromEnv and use the NewDebug method (I think there’s one)

sethvargo (commented):
I’m not convinced the envvar buys us anything here? I think we should just fallback without the additional layer of indirection and remove in the next release

mikehelmick:
> I’m not convinced the envvar buys us anything here? I think we should just fallback without the additional layer of indirection and remove in the next release

meaning remove the flag?

sethvargo:
Yea, I think we just always fallback in this release.

tools/get-certificate/main.go
mikehelmick: this still allows the user to specify

mikehelmick:
> Yea, I think we just always fallback in this release.

done. Always on in this release.

sethvargo (approved):

Issue #1232, opened 2020-12-02
pha stats: expired claim attempt


mikehelmick:
### TL;DR

Number of times that expired (not invalid) codes were attempted to be claimed.

ideally an indicator of someone who go a real code but tried to use it too late.

sethvargo:
I see two use cases here:

1. You issued a code to one person and that one person did not claim it in time.

    I think this use case isn't solved by stats. The stats won't tell you _which_ user expired, and we intentionally don't have that data link.

2. You issued a bunch of codes that expired before they were claimed.

    While this data might be useful, it's not actionable. You can't re-issue the codes because there's no link back to the patient (by design).

sethvargo:
Duplicate of https://github.com/google/exposure-notifications-verification-server/issues/1497

sethvargo:
The backend piece of this is done. We opted to go for "failed" instead of "expired" which encompasses:

- codes that were definitely expired
- codes that don't exist in the database (could be super-expired or a typo)

Issue #1231, opened 2020-12-02
pha stats: claim histogram


mikehelmick:
### TL;DR

Histogram displaying the number of hours between code issue and code claim.

Rolling over the last 30 days.

sethvargo:
I'm not convinced this is valuable information for capturing or showing. We should talk about this more and decide if we actually want to do it.

PR #1629, opened 2021-01-19
Make seed script also verify codes and claim tokens

sethvargo:
This updates the seed script to produce a more probabilistic distribution of codes claimed, codes failed, tokens claimed, and tokens failed based on a random distribution. This can be an expensive operation, which is why it's guarded by the newly-added `-stats` flag.

As part of this change, the function signatures for verifying codes and claiming tokens were updated to accepted a time parameter. This is only used for testing and seeding. In production codepaths, the value of time is always `time.Now().UTC()`, which was the previously-hardcoded value in the function body.

**Release Note**

```release-note
Make seed script also optionally verify codes and claim tokens
```

### DIFF
--- a/pkg/database/token.go
+++ b/pkg/database/token.go
@@ -140,7 +140,7 @@ func (t *Token) Subject() *Subject {
 
 // ClaimToken looks up the token by ID, verifies that it is not expired and that
 // the specified subject matches the parameters that were configured when issued.
-func (db *Database) ClaimToken(authApp *AuthorizedApp, tokenID string, subject *Subject) error {
+func (db *Database) ClaimToken(t time.Time, authApp *AuthorizedApp, tokenID string, subject *Subject) error {
 	if err := db.db.Transaction(func(tx *gorm.DB) error {
 		var tok Token
 		if err := tx.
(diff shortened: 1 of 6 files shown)


sethvargo:
This is for #1231 and #1232 development

pkg/database/token.go
```diff
@@ -208,7 +208,7 @@ func (db *Database) ClaimToken(authApp *AuthorizedApp, tokenID string, subject *
 // The verCode can be the "short code" or the "long code" which impacts expiry time.
 //
 // The long term token can be used later to sign keys when they are submitted.
-func (db *Database) VerifyCodeAndIssueToken(authApp *AuthorizedApp, verCode string, acceptTypes api.AcceptTypes, expireAfter time.Duration) (*Token, error) {
+func (db *Database) VerifyCodeAndIssueToken(t time.Time, authApp *AuthorizedApp, verCode string, acceptTypes api.AcceptTypes, expireAfter time.Duration) (*Token, error) {
```
mikehelmick, line 211: I think this function should defensively put t into UTC

sethvargo: Done everywhere

pkg/database/token.go
```diff
@@ -312,16 +316,16 @@ func (db *Database) PurgeTokens(maxAge time.Duration) (int64, error) {
 
 // updateStatsCodeInvalid updates the statistics, increasing the number of codes
 // that were invalid.
-func (db *Database) updateStatsCodeInvalid(authApp *AuthorizedApp) {
-	now := timeutils.UTCMidnight(time.Now().UTC())
+func (db *Database) updateStatsCodeInvalid(t time.Time, authApp *AuthorizedApp) {
+	t = timeutils.UTCMidnight(t.UTC())
```
mikehelmick, line 320: you don't need to do this here since UTCMidnight does this.

https://github.com/google/exposure-notifications-server/blob/main/pkg/timeutils/midnight.go#L28

sethvargo: Done

mikehelmick (approved):

PR #1630, opened 2021-01-19
Ping when opening SQL connection

sethvargo:
Also reduce shadowing and increase total allowed timeout to 30s.

I've noticed an increase in test flakes, likely due to the increase in test coverage and additional overhead of creating more database instances.

**Release Note**

```release-note
Ping when opening SQL connection
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -149,12 +149,16 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
 	// opencenusus stats.
 	var rawSQL *sql.DB
 	if err := withRetries(ctx, func(ctx context.Context) error {
-		var err error
-		rawSQL, err = sql.Open("ocsql", c.ConnectionString())
+		s, err := sql.Open("ocsql", c.ConnectionString())
 		if err != nil {
+			return fmt.Errorf("failed to open sql connection: %w", err)
+		}
+
+		if err := s.Ping(); err != nil {
 			return retry.RetryableError(err)
 		}
-		db.statsCloser = ocsql.RecordStats(rawSQL, 5*time.Second)
+
+		rawSQL = s
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to create sql connection: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/database/database.go
```diff
@@ -149,33 +149,39 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
 	// opencenusus stats.
 	var rawSQL *sql.DB
 	if err := withRetries(ctx, func(ctx context.Context) error {
-		var err error
-		rawSQL, err = sql.Open("ocsql", c.ConnectionString())
+		s, err := sql.Open("ocsql", c.ConnectionString())
 		if err != nil {
+			return fmt.Errorf("failed to open sql connection: %w", err)
```
sethvargo, line 154: This is intentionally not retryable. "Open" doesn't test the connection, it just parses the string and creates a struct.

pkg/database/database.go
```diff
@@ -149,33 +149,39 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
...
+		}
+
+		if err := s.Ping(); err != nil {
 			return retry.RetryableError(err)
 		}
-		db.statsCloser = ocsql.RecordStats(rawSQL, 5*time.Second)
```
sethvargo, line 157: Moved this out of the loop.

pkg/database/database.go
```diff
@@ -149,33 +149,39 @@ func (db *Database) OpenWithCacher(ctx context.Context, cacher cache.Cacher) err
...
 	var rawSQL *sql.DB
 	if err := withRetries(ctx, func(ctx context.Context) error {
-		var err error
-		rawSQL, err = sql.Open("ocsql", c.ConnectionString())
+		s, err := sql.Open("ocsql", c.ConnectionString())
 		if err != nil {
+			return fmt.Errorf("failed to open sql connection: %w", err)
+		}
+
+		if err := s.Ping(); err != nil {
```
mikehelmick, line 157: does `s` need to be closed if ping fails?

sethvargo: Hmm yea, good point. I wonder if that's why we're exhausting... I refactored this a bit.

mikehelmick (approved):

PR #1633, opened 2021-01-19
Unwind stats when we delete verification codes

whaught:
## Proposed Changes

* When we fail to SMS a code, we invalidate it by deleting the code. This change also decrements the stats

**Release Note**

```release-note
Decrement stats for unsuccessful code-issuance
```

### DIFF
--- a/pkg/database/vercode.go
+++ b/pkg/database/vercode.go
@@ -112,11 +112,11 @@ func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
 func (v *VerificationCode) updateStats(scope *gorm.Scope, increment bool) {
 	date := timeutils.Midnight(v.CreatedAt)
 
-	var incString string
+	var inc rune
 	if increment {
-		incString = "+ 1"
+		inc = '+'
 	} else {
-		incString = "- 1"
+		inc = '-'
 	}
 
 	// If the issuer was a user, update the user stats for the day.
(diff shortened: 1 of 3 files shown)


pkg/database/vercode.go
```diff
@@ -106,16 +106,27 @@ func (v *VerificationCode) BeforeSave(tx *gorm.DB) error {
 // to update statistics about usage. If the executions fail, an error is logged
 // but the transaction continues. This is called automatically by gorm.
 func (v *VerificationCode) AfterCreate(scope *gorm.Scope) {
+	v.updateStats(scope, true)
+}
+
+func (v *VerificationCode) updateStats(scope *gorm.Scope, increment bool) {
 	date := timeutils.Midnight(v.CreatedAt)
 
+	var incString string
```
mikehelmick, line 115: simplification

```
sign := "+"
if !increment {
  sign = "-"
}
```

And then the 1 can be embedded in the statement still

whaught: sure

sethvargo (commented):
I'm pretty -1 on this. I think stats should be +1 only (e.g. we need a guard now to prevent a stat from going negative), and we advertise them as best-effort.

mikehelmick:
> I'm pretty -1 on this. I think stats should be +1 only (e.g. we need a guard now to prevent a stat from going negative), and we advertise them as best-effort.

I didn't have a chance to go back and and see where we're dispatching the increment - seems like delaying that should mitigate this issue

whaught:
> > I'm pretty -1 on this. I think stats should be +1 only (e.g. we need a guard now to prevent a stat from going negative), and we advertise them as best-effort.
> 
> I didn't have a chance to go back and and see where we're dispatching the increment - seems like delaying that should mitigate this issue

We could do it that way, but currently we have it on a GORM AfterCreate hook which feels like the most foolproof way of never missing the increment (hard to have fail-in-the-middle). This strategy mirrors our after the fact delete of the code itself. We only do this if RowsAffected > 1, so I don't forsee this going negative if we trust our transactions.

sethvargo:
I think it's okay if we miss an increment - all stats are advertised as best-effort.

My main concern is this is attempting to make "codes issued" match "people notified" which I'm pretty strongly opposed to. I'd rather solve this with education on what "codes issued" means.

Issue #1634, opened 2021-01-19
Onboarding redirect


mikehelmick:
### TL;DR

In the ENX redirect service, implement a top level redirect that prompts for onboarding.

For iOS this results in a redirect to the new ens://onboard spec #1610 and #1632

For Android, redirect to `market://search?q=exposure%20notifications`

### Design

**Proposal**

1. Accept requests at the top level ENX redirect domain (no parameters). i.e. `https://en.express`
    - iOS redirect to `ens://onboarding` and Android redirect to `market://search?q=exposure%20notifications`
    - this will require adding a new domain to the load balancer cert - and a new piece of config for the redirect service.
    - This must be done without causing an SSL outage.

2. Accept top level hits on the registered subdomains, but do the same onboarding flow.
    - iOS, redirect to `ens://onboarding`
    - Android - in the case the region is known, so if there is an associated mobile app redirect to that app in the play store, if it isn't known, use the market link above.

PR #1635, opened 2021-01-19
Add top-level onboarding domain to redirect service

sethvargo:
This adds the necessary configuration to support the ens://onboarding specification at the provided domain.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1634

**Release Note**

```release-note
Add optional onboarding support to redirect service. If `ONBOARDING_DOMAIN` is configured, the redirector service will redirect to the device-specific appropriate location to trigger onboarding when the host matches the value. This value can be configured via Terraform, but care should be taken to avoid an outage due to SSL certificate provisioning for the new domain.
```

/assign @mikehelmick

### DIFF
--- a/pkg/controller/redirect/helpers_test.go
+++ b/pkg/controller/redirect/helpers_test.go
@@ -275,12 +275,11 @@ func TestDecideRedirect(t *testing.T) {
 		Path: "/v",
 	}
 	q := relativePinURL.Query()
-	q.Set("c", "1234567890abcdef")
+	q.Set("c", "123456")
 	relativePinURL.RawQuery = q.Encode()
 
 	cases := []struct {
 		name         string
-		host         string
 		url          string
 		altURL       *url.URL
 		enxEnabled   bool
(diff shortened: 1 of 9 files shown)


pkg/controller/redirect/redirect.go
```diff
@@ -25,6 +25,14 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 )
 
+const (
+	// androidOnboardingRedirect and iosOnboardingRedirect are the URLs to
+	// redirect to for onboarding.
+	androidOnboardingRedirect = "market://search?q=exposure%20notifications"
+	iosOnboardingRedirect     = "ens://onboarding"
+	genericOnboardingRedirect = "https://www.google.com/covid19/exposurenotifications/"
```
sethvargo, line 33: @mikehelmick this wasn't in the spec, but we may want to add it. If we didn't detect ios or android (e.g. someone hit it in their browser), I decided to redirect to the generic EN landing page.

sethvargo:
/retest

503s from the Go proxy

pkg/controller/redirect/helpers_test.go
```diff
@@ -319,18 +319,26 @@ func TestDecideRedirect(t *testing.T) {
...
+			userAgent:    userAgentAndroid,
+			appStoreData: &appLinkNeither,
+			expected:     "market://search?q=exposure%20notifications",
+		},
 
 		// iOS
 		{
 			name:         "ios_both",
-			url:          "https://ios.example.com/store/moosylvania",
+			url:          "https://ios.example.com/store/moosylvania?c=123456",
```
mikehelmick, line 334: do we want the code in that example?

pkg/controller/redirect/redirect.go
mikehelmick: I don't think it needs to be in the spec - as that is for device handling

pkg/controller/redirect/helpers_test.go
sethvargo: The previous implementation did not redirect to the appstore with the code. We could add it, but that would be a behavior change (and it would be sharing the code with external servers - they would show up in Apple's App Store request logs). WDYT?

sethvargo: @mikehelmick ^

mikehelmick: right - that's what I'm worried about - I don't think it's useful to send the code to the store and have it in the logs

sethvargo: I see the confusion. This is the _incoming_ url. The outgoing url is below, sans code.

PR #1636, opened 2021-01-19
add auto rotation realm guide

mikehelmick:
## Proposed Changes

* Add realm admin auto rotate guide

**Release Note**

```release-note
DOCS: Realm admin guide, add instructions for enabling automatic signing key rotation.
```


sethvargo (approved):
/lgtm

PR #1638, opened 2021-01-20
Increment stats at the end of successfully issuing codes

whaught:
## Proposed Changes

* This is an alternative to https://github.com/google/exposure-notifications-verification-server/pull/1633
* Wait until the end of code issuance to increment stats - only increment for codes that are successful.
    * Batch issuance can increment stats by N, so fewer queries

/hold 

**Release Note**

```release-note
Increment code-issue stats at the end of issuance logic. This avoids recording [known] failures.
```

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -19,7 +19,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/api"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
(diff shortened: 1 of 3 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -86,5 +87,26 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*api.IssueCodeReq
...
+// recordStats increments stats for successfully issued codes
+func (c *Controller) recordStats(ctx context.Context, results []*IssueResult) {
+	// Count successes and update stats
+	var code *database.VerificationCode
+	count := 0
+	for _, result := range results {
+		if result.ErrorReturn != nil {
+			continue
+		}
+		code = result.VerCode
```
sethvargo, line 103: This is a little weird, since it's shadowing the `code`?

sethvargo (commented):
I like this approach much better than attempting to rollback, but I don't think `UpdateStats` should return an error, since that hints to callers that they should _handle_ the error, but stats are best-effort.

pkg/database/vercode.go
```diff
@@ -341,6 +281,79 @@ func (db *Database) DeleteVerificationCode(code string) error {
...
+	if v.IssuingUserID != 0 {
+		sql := fmt.Sprintf(`
+			INSERT INTO user_stats (date, realm_id, user_id, codes_issued)
+				VALUES ($1, $2, $3, 1)
+			ON CONFLICT (date, realm_id, user_id) DO UPDATE
+				SET codes_issued = user_stats.codes_issued + %d
+		`, issued)
+
+		if err := db.db.Exec(sql, date, v.RealmID, v.IssuingUserID).Error; err != nil {
+			multierror.Append(merr, fmt.Errorf("failed to update stats: %v", err))
```
sethvargo, line 305: I don't know if multierror is necessary here. I'd suggest we pull the logger out of the context at the beginning and just call `logger.Warnw` when appropriate.

sethvargo (approved):
/lgtm

I have a suggestion for making this a big clearer and more performant, but it can be a follow up

pkg/database/vercode.go
sethvargo: @whaught FYI

whaught: Sure. Will fold into an unrelated PR

PR #1639, opened 2021-01-20
Fix the query used by UpstreamUserRecreates alert.

yegle:


terraform/alerting/alerts.tf
```diff
@@ -263,11 +263,12 @@ resource "google_monitoring_alert_policy" "UpstreamUserRecreates" {
     condition_monitoring_query_language {
       duration = "600s"
       query    = <<-EOT
-      fetch generic_task :: custom.googleapis.com/opencensus/en-verification-server/user/upstream_user_recreate
+      fetch
+      generic_task :: custom.googleapis.com/opencensus/en-verification-server/user/upstream_user_recreate_count
       | align rate(5m)
       | every 1m
-      | group_by [resource.project_id],
```
sethvargo, line 269: I figured we'd want this for the server, not by realm. @whaught thoughts?

whaught: true. realm isn't really important: this is server-wide auth

yegle: I removed the group-by altogether. The previous group by field `resource.project_id` is not useful as our monitoring workspace contains only one project (it would be useful if we have a monitoring workspace with multiple projects).

whaught (approved):

PR #1642, opened 2021-01-20
Resolving nits

whaught:
## Proposed Changes

* Followup from https://github.com/google/exposure-notifications-verification-server/pull/1638
   * Fire away in the comments if we have any other nits to fix


sethvargo (approved):
/lgtm

PR #1643, opened 2021-01-20
Distinct certificates on load balancer

sethvargo:
**Release Note**

```release-note
Ensure distinct certificates on the redirect load balancer for root certificates.
```


mikehelmick (approved):
I don't _think_ this will cause any issues during upgrade - but we should put it in the release notes.

/hold

for release notes update

sethvargo:
/unhold

This is the load balancer for root (formerly www) that isn't actually used. Added a release note anyway.

PR #1647, opened 2021-01-20
Condition should contain a unit.

### LINKED ISSUE #1644: Error creating AlertPolicy: googleapi: Error 400
```text
Error creating AlertPolicy: googleapi: Error 400

Ignoring units for operation '>', which is combining two values, the first with
unit '1/s' and the second with no units. Units need to be given for neither or
both of the inputs to '>'. Units can be added to the second argument by the
`cast_units` function (example `cast_units(<expression>, "By/s")`) or, for
literals, by annotation (example `3.3 "GiBy"`).  Units can be removed by 
`cast_units(<expression>, "")`.
```

/assign @yegle

yegle:
Fixes #1644

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -268,7 +268,7 @@ resource "google_monitoring_alert_policy" "UpstreamUserRecreates" {
       | align rate(5m)
       | every 1m
       | group_by [], [val: sum(value.upstream_user_recreate_count)]
-      | condition val > 5 'Hz'
+      | condition val > 5 '1/s'
       EOT
       trigger {
         count = 1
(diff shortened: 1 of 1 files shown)


terraform/alerting/alerts.tf
```diff
@@ -267,8 +267,8 @@ resource "google_monitoring_alert_policy" "UpstreamUserRecreates" {
       generic_task :: custom.googleapis.com/opencensus/en-verification-server/user/upstream_user_recreate_count
       | align rate(5m)
       | every 1m
-      | group_by [], [val: aggregate(value.upstream_user_recreate_count)]
-      | condition val > 5
+      | group_by [], [val: sum(value.upstream_user_recreate_count)]
+      | condition val > 5 'Hz'
```
sethvargo, line 271: Hertz?

yegle: Hz is an alias of 1/s. But agree this is a bit confusing.

sethvargo (approved):
/lgtm

Issue #1640, opened 2021-01-20
Signed SMS payloads


mikehelmick:
### TL;DR

Implement signed SMS payloads.

### Design

**Proposal**

*  Per-realm SMS signing keys. These are distinct from the verification certificate signing keys

*  The signature is calculated over a preamble and the message body up to the keyword "Authorization:" Everything after that is the signature itself, including the kid to use from the public key set for signature verification.

/assign

Targeting v0.21 - may not make it until v0.22

sethvargo:
A few questions:

1. I assume we'll also want to hook into rotation. At what frequency should we rotate and for how long should a rotated key still be included in the allowed list?

1. The spec required some kind of server-side kill switch. I was thinking we could just implement system-admin and realm-admin revocation permissions. Does that make sense?

1. Where are we publishing the public keys? I remember the original proposal was to use the jwks endpoints, but this would be a _different_ type of key, so it'd require a new discovery endpoint, right?

mikehelmick:
1) Manual rotation only - we can probably reuse a lot of the UI from the existing signing keys UI

2) Server side kill switch is related to OS specific thing to stop verifying these messages - nothing here

3) Manual distribution of key to Google and Apple

PR #1649, opened 2021-01-21
add SMS signing keys at the DB layer

mikehelmick:
towards #1640

## Proposed Changes

* Add new SMSSiginingKey model
* Create ManagedKey interface
* Refactor realm key management to work on multiple type of managed keys

**Release Note**

```release-note
Add database model for managed keys for signing SMS messages.
```

### DIFF
--- a/pkg/database/managed_key.go
+++ b/pkg/database/managed_key.go
@@ -24,7 +24,16 @@ type ManagedKey interface {
 	// IsActive() returns true if this key is active
 	IsActive() bool
 
-	SetRealmID(id uint)
 	SetManagedKeyID(keyID string)
 	SetActive(active bool)
+
+	// These are expected to be static across all instances of an implementing type.
+	Table() string
+	Purpose() string
+}
+
+// RealmManagedKey indicates that this key is owned by a realm.
+type RealmManagedKey interface {
+	ManagedKey
+	SetRealmID(id uint)
 }
--- a/pkg/database/managed_key.go
+++ b/pkg/database/managed_key.go
@@ -1,0 +1,30 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+// ManagedKey is an interface that allows for a realm to manage signing keys
+// for different purposes.
+type ManagedKey interface {
+	// GetKID returns the public key version string
+	GetKID() string
+	// ManagedKeyID returns the reference to the key ID in the KMS.
+	ManagedKeyID() string
+	// IsActive() returns true if this key is active
+	IsActive() bool
+
+	SetRealmID(id uint)
+	SetManagedKeyID(keyID string)
+	SetActive(active bool)
+}
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2038,6 +2038,27 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 				return nil
 			},
 		},
+		{
+			ID: "00090-AddSMSSigningKeys",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`CREATE TABLE sms_signing_keys (
+						id BIGSERIAL,
+						created_at TIMESTAMP WITH TIME ZONE,
+						updated_at TIMESTAMP WITH TIME ZONE,
+						deleted_at TIMESTAMP WITH TIME ZONE,
+						realm_id INTEGER,
+						key_id TEXT,
+						active BOOLEAN,
+						PRIMARY KEY (id))`,
+					`CREATE INDEX IF NOT EXISTS idx_sms_signing_keys_realm ON sms_signing_keys (realm_id)`,
+				)
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`DROP TABLE IF EXISTS sms_signing_keys`)
+			},
+		},
 	}
 }
 
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2047,11 +2047,12 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 						created_at TIMESTAMP WITH TIME ZONE,
 						updated_at TIMESTAMP WITH TIME ZONE,
 						deleted_at TIMESTAMP WITH TIME ZONE,
-						realm_id INTEGER,
-						key_id TEXT,
-						active BOOLEAN,
+						realm_id INTEGER NOT NULL,
+						key_id TEXT NOT NULL,
+						active BOOLEAN NOT NULL DEFAULT false,
 						PRIMARY KEY (id))`,
-					`CREATE INDEX IF NOT EXISTS idx_sms_signing_keys_realm ON sms_signing_keys (realm_id)`,
+					`CREATE INDEX idx_sms_signing_keys_realm ON sms_signing_keys (realm_id)`,
+					`CREATE UNIQUE INDEX uix_sms_signing_keys_active ON sms_signing_keys (realm_id, active) WHERE (active IS TRUE)`,
 				)
 			},
 			Rollback: func(tx *gorm.DB) error {
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -692,10 +692,10 @@ func (db *Database) AbusePreventionEnabledRealmIDs() ([]uint64, error) {
 	return ids, nil
 }
 
-// GetCurrentSigningKey returns the currently active certificate signing key, the one marked
+// CurrentSigningKey returns the currently active certificate signing key, the one marked
 // active in the database. If there is more than one active, the most recently
 // created one wins. Should not occur due to transactional update.
-func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
+func (r *Realm) CurrentSigningKey(db *Database) (*SigningKey, error) {
 	var signingKey SigningKey
 	if err := db.db.
 		Where("realm_id = ?", r.ID).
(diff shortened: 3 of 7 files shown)


pkg/database/managed_key.go
```diff
@@ -0,0 +1,30 @@
...
+// for different purposes.
+type ManagedKey interface {
+	// GetKID returns the public key version string
+	GetKID() string
+	// ManagedKeyID returns the reference to the key ID in the KMS.
+	ManagedKeyID() string
+	// IsActive() returns true if this key is active
+	IsActive() bool
+
+	SetRealmID(id uint)
```
sethvargo, line 27: I think this function should be a separate interface, since some managed keys will not be realm specific, right? (Like system keys?) Or do you plan to use a realmID == 0 for those cases like we do for audit logs?

pkg/database/migrations.go
```diff
@@ -2038,6 +2038,27 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+		{
+			ID: "00090-AddSMSSigningKeys",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`CREATE TABLE sms_signing_keys (
+						id BIGSERIAL,
+						created_at TIMESTAMP WITH TIME ZONE,
+						updated_at TIMESTAMP WITH TIME ZONE,
+						deleted_at TIMESTAMP WITH TIME ZONE,
+						realm_id INTEGER,
```
sethvargo, line 2050: Some of these probably need a `NOT NULL` constraint

pkg/database/migrations.go
```diff
@@ -2038,6 +2038,27 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+						key_id TEXT,
+						active BOOLEAN,
```
sethvargo, line 2052: We should create an index that ensures only one value for `active` can exist in the scope of a `realm_id` (there's some other tables that have this, it's a conditional index)

pkg/database/realm.go
```diff
@@ -711,21 +711,51 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
 	return &signingKey, nil
 }
 
+// GetCurrentSMSSigningKey returns the currently active SMS signing key, the one marked
```
sethvargo, line 714: I know this is following the same naming pattern as `GetCurrentSigningKey`, but per https://golang.org/doc/effective_go.html#Getters, this should be called `CurrentSMSSigningKey()`.

pkg/database/realm.go
```diff
@@ -711,21 +711,51 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
...
+// created one wins. Should not occur due to transactional update.
+func (r *Realm) GetCurrentSMSSigningKey(db *Database) (*SMSSigningKey, error) {
+	var signingKey SMSSigningKey
+	if err := db.db.
+		Where("realm_id = ?", r.ID).
+		Where("active = ?", true).
+		Order("sms_signing_keys.created_at DESC").
+		First(&signingKey).
+		Error; err != nil {
+		if IsNotFound(err) {
```
sethvargo, line 725: This error should bubble up to the caller to check.

pkg/database/realm.go
```diff
@@ -711,21 +711,51 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
...
 
+// SetActiveSMSSigningKey sets a specific signing key to active=true for the realm,
+// and transactionally sets all other signing keys to inactive. It accepts the
+// database primary key ID but returns the KID of the now-active key.
+func (r *Realm) SetActiveSMSSigningKey(db *Database, id uint) (string, error) {
+	signingKey := &SMSSigningKey{}
+	return r.setActiveManagedSigningKey(db, id, "sms_signing_keys", "SMS", signingKey)
+}
+
+func (r *Realm) setActiveManagedSigningKey(db *Database, id uint, table, purpose string, signingKey ManagedKey) (string, error) {
```
sethvargo, line 749: `purpose` appears unused in the function body

pkg/database/realm.go
```diff
@@ -711,21 +711,51 @@ func (r *Realm) GetCurrentSigningKey(db *Database) (*SigningKey, error) {
...
+	}
+	return &signingKey, nil
+}
+
 // SetActiveSigningKey sets a specific signing key to active=true for the realm,
 // and transactionally sets all other signing keys to inactive. It accepts the
 // database primary key ID but returns the KID of the now-active key.
 func (r *Realm) SetActiveSigningKey(db *Database, id uint) (string, error) {
-	var signingKey SigningKey
+	signingKey := &SigningKey{}
```
sethvargo, line 737: Nit: I think it'd be better just to pass in `&signingKey` to the call.

pkg/database/realm.go
```diff
@@ -774,6 +804,23 @@ func (r *Realm) ListSigningKeys(db *Database) ([]*SigningKey, error) {
...
 }
 
+// ListSMSSigningKeys returns the non-deleted signing keys for a realm
+// ordered by created_at desc.
+func (r *Realm) ListSMSSigningKeys(db *Database) ([]*SMSSigningKey, error) {
+	var keys []*SMSSigningKey
+	if err := db.db.
+		Model(r).
+		Order("sms_signing_keys.created_at DESC").
+		Related(&keys).
```
sethvargo, line 814: Please don't use `Related`. Consider:

```go
var keys []*SMSSigningKey
if err := db.db.
  Model(&SMSSigningKey{}).
  Where("realm_id = ?", r.ID).
  Order("sms_signing_keys.created_at DESC").
  Find(&keys).
  Error; err != nil {
}
```

pkg/database/realm.go
```diff
@@ -1287,12 +1334,28 @@ func (r *Realm) SigningKeyID() string {
...
+}
+
+// CreateSMSSigningKeyVersion creates a new SMS signing key versino on the key manager
+// and saves a reference to the new key version in the database.
+func (r *Realm) CreateSMSSigningKeyVersion(ctx context.Context, db *Database) (string, error) {
+	newKey := &SMSSigningKey{}
+	return r.createdManagedSigningKey(ctx, db, r.SMSSigningKeyID(), "sms_signing_keys", "SMS", newKey)
+}
+
+func (r *Realm) createdManagedSigningKey(ctx context.Context, db *Database, keyID, table, purpose string, signingKey ManagedKey) (string, error) {
```
sethvargo, line 1358: Instead of passing in `purpose`, you can use `%T` on the `signingKey` param in the errors. It seems that's the only purpose for passing in purpose.

pkg/database/sms_signing_key.go
```diff
@@ -0,0 +1,76 @@
...
+var _ ManagedKey = (*SMSSigningKey)(nil)
+
+// SMSSigningKey represents a reference to a KMS backed signing key
+// version for SMS payload signing.
+type SMSSigningKey struct {
+	gorm.Model
+	Errorable
+
+	// A signing key belongs to exactly one realm.
+	RealmID uint `gorm:"index:realm"`
```
sethvargo, line 33: You don't need the `gorm` reference since you created this table manually. That's only used for automigrate.

pkg/database/sms_signing_key.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Reference to an exact version of a key in the KMS
+	KeyID  string
+	Active bool
+}
+
+// GetKID returns the 'kid' field value to use in signing JWTs.
+func (s *SMSSigningKey) GetKID() string {
+	return fmt.Sprintf("r%dv%d", s.RealmID, s.ID)
```
sethvargo, line 42: This doesn't seem to match `realm.SMSSigningKeyID()`?

sethvargo (changes requested):

pkg/database/managed_key.go
mikehelmick: Introduced RealmManagedKey - moved SetRealmID there.

this opens up extending ManagedKey to token signing keys in a follow up.

pkg/database/migrations.go
mikehelmick: please double check my index statement

pkg/database/realm.go
mikehelmick: I moved Table() and Purpose() to methods on the ManagedKey interface.

pkg/database/sms_signing_key.go
mikehelmick: SMSSigningKeyID is the KMS key - this is the KID for public consumption

pkg/database/realm.go
```diff
@@ -692,20 +692,33 @@ func (db *Database) AbusePreventionEnabledRealmIDs() ([]uint64, error) {
...
+}
+
+// CurrentSMSSigningKey returns the currently active SMS signing key, the one marked
+// active in the database. If there is more than one active, the most recently
+// created one wins. Should not occur due to transactional update.
+func (r *Realm) CurrentSMSSigningKey(db *Database) (*SMSSigningKey, error) {
+	var signingKey SMSSigningKey
+	if err := db.db.
+		Where("realm_id = ?", r.ID).
+		Where("active = ?", true).
```
sethvargo, line 718: Fun fact: this is now O(1) instead of O(n) because of that index.

pkg/database/realm.go
```diff
@@ -692,20 +692,33 @@ func (db *Database) AbusePreventionEnabledRealmIDs() ([]uint64, error) {
...
+		Order("sms_signing_keys.created_at DESC").
```
sethvargo, line 719: Shouldn't need order_by here

pkg/database/realm.go
```diff
@@ -1384,46 +1433,55 @@ func (r *Realm) CreateSigningKeyVersion(ctx context.Context, db *Database) (stri
...
+	signingKey := &SigningKey{}
+	return r.destroyManagedSigningKey(ctx, db, id, "signing_keys", "certificate", signingKey)
+}
+
+func (r *Realm) DestroySMSSigningKeyVersion(ctx context.Context, db *Database, id interface{}) error {
+	signingKey := &SMSSigningKey{}
+	return r.destroyManagedSigningKey(ctx, db, id, "sms_signing_keys", "SMS", signingKey)
+}
+
+func (r *Realm) destroyManagedSigningKey(ctx context.Context, db *Database, id interface{}, table, purpose string, signingKey ManagedKey) error {
```
sethvargo, line 1445: I think you can call Purpose() instead of passing it in

pkg/database/signing_key.go
```diff
@@ -40,6 +42,34 @@ func (s *SigningKey) GetKID() string {
...
+
+func (s *SigningKey) SetManagedKeyID(keyID string) {
+	s.KeyID = keyID
+}
+
+func (s *SigningKey) SetActive(active bool) {
+	s.Active = active
+}
+
+func (s *SigningKey) Table() string {
```
sethvargo, line 65: Does passing in as `Model` not work? It might not, I've never tried.

pkg/database/realm.go
mikehelmick: oops - missed the destroy method

pkg/database/signing_key.go
mikehelmick: will mess w/ that separate

sethvargo (approved):
/lgtm

PR #1650, opened 2021-01-21
Implement first pass at SMS signature algorithm

sethvargo:
I expect this to change over time as we reach consensus with Apple, but this is good to review and merge for now.

Part of https://github.com/google/exposure-notifications-verification-server/issues/1640

**Release Note**

```release-note
Implement first pass at SMS signature algorithm package
```

### DIFF
--- a/pkg/signatures/sms.go
+++ b/pkg/signatures/sms.go
@@ -26,6 +26,9 @@ import (
 )
 
 const (
+	// authPrefix is the beginning of the authorization bit.
+	authPrefix = " Authorization: "
+
 	// Dot represents a period, which is used as a separator in some signatures.
 	dot = "."
 )
@@ -49,7 +52,7 @@ func SMSSignature(signer crypto.Signer, keyID string, t time.Time, purpose SMSPu
 	}
 	sig := base64.RawStdEncoding.EncodeToString(b)
 
-	return keyID + dot + sig, nil
+	return authPrefix + keyID + dot + sig, nil
 }
 
 // smsSignatureString builds the string that is to be signed. The provided date
@@ -58,5 +61,5 @@ func SMSSignature(signer crypto.Signer, keyID string, t time.Time, purpose SMSPu
 // including any codes and links.
 func smsSignatureString(t time.Time, purpose SMSPurpose, phone, body string) string {
 	t = t.UTC()
-	return string(purpose) + dot + phone + dot + t.Format(project.RFC3339Date) + dot + body
+	return string(purpose) + dot + phone + dot + t.Format(project.RFC3339Date) + dot + body + authPrefix
 }
--- a/pkg/signatures/sms.go
+++ b/pkg/signatures/sms.go
@@ -1,0 +1,67 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


pkg/signatures/sms.go
```diff
@@ -0,0 +1,67 @@
...
+	signingString := smsSignatureString(t, purpose, phone, body)
+
+	digest := sha256.Sum256([]byte(signingString))
+	b, err := signer.Sign(rand.Reader, digest[:], nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to sign sms: %w", err)
+	}
+	sig := base64.RawStdEncoding.EncodeToString(b)
+
+	return authPrefix + t.Format(project.RFC3339Date) + dot + keyID + dot + sig, nil
```
sethvargo, line 57: I think @mikehelmick and I are in agreement that we should **not** include the date here.

pkg/signatures/sms.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"time"
+
+	"github.com/google/exposure-notifications-verification-server/internal/project"
+)
+
+const (
+	// authPrefix is the string to prepend to the authentication
+	// signature.
+	authPrefix = "Authentication: "
```
mikehelmick, line 31: I think this is intended to be part of what is signed.

pkg/signatures/sms.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	// SMSPurposeENReport is an SMS purpose for EN reporting.
+	SMSPurposeENReport SMSPurpose = "EN Report"
+)
+
+// SMSSignature returns the signature of the message uses the provided signer.
+func SMSSignature(signer crypto.Signer, keyID string, t time.Time, purpose SMSPurpose, phone, body string) (string, error) {
+	t = t.UTC()
+	signingString := smsSignatureString(t, purpose, phone, body)
```
mikehelmick, line 48: as above - I Think we want to add + " Authentication:" to the end of the body here (well, probably in the smsSignatureString message itself)

pkg/signatures/sms.go
mikehelmick: let's remove date, and I think also authPrefix (see other notes)

pkg/signatures/sms.go
mikehelmick: also - the trailing space will likely become problematic

pkg/signatures/sms.go
sethvargo: @mikehelmick or do we expect the SMS template itself to declare it like `blah blah Authorization:[[auth]]`?

mikehelmick: no - I think it needs to be inserted here. Because using SMS auth will be optional. If enabled, we should insert it.

pkg/signatures/sms.go
sethvargo: Done

pkg/signatures/sms.go
sethvargo: Done

pkg/signatures/sms.go
sethvargo: Then this function needs to return a _new_ SMS body, which feels like a really weird API.

mikehelmick: no, it can still return what should be appended to the body

sethvargo: Okay, done

mikehelmick (commented):
/lgtm

PR #1658, opened 2021-01-21
Add tests for api package

sethvargo:
Part of https://github.com/google/exposure-notifications-verification-server/issues/1399

**Release Note**

```release-note
Add tests for api package
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -17,6 +17,7 @@
 package api
 
 import (
+	"bytes"
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -156,9 +157,7 @@ func (p Padding) MarshalJSON() ([]byte, error) {
 // UnmarshalJSON is a custom JSON unmarshaler for padding.
 // The field is meaningless bytes, so this is just a passthrough.
 func (p *Padding) UnmarshalJSON(b []byte) error {
-	if l := len(b); l > 2 {
-		*p = b[1 : l-2] // remove outer quotes
-	}
+	*p = bytes.Trim(b, `"`)
```
sethvargo, line 160: @whaught found bug

whaught: this is much better

whaught: that said the padding is not interpreted and only used for verifying in the integration test

whaught (approved):

PR #1661, opened 2021-01-22
Emit stats as json on verification server + basic stats graph

whaught:

## Proposed Changes

* Converts saved stats back to json to emit
* Add key-server stats struct to cached RealmStats
* Will be interpreted on the stats page to display charts

Shows some basic stats:
keyserverstats

**Release Note**

```release-note
emit key-server stats as json for charting
```


internal/routes/server.go
```diff
@@ -409,6 +409,8 @@ func statsRoutes(r *mux.Router, c *stats.Controller) {
 
 	r.Handle("/realm/external-issuers.csv", c.HandleRealmExternalIssuersStats(stats.StatsTypeCSV)).Methods("GET")
 	r.Handle("/realm/external-issuers.json", c.HandleRealmExternalIssuersStats(stats.StatsTypeJSON)).Methods("GET")
+
+	r.Handle("/realm/key-server-stats.json", c.HandleKeyServerStats(stats.StatsTypeJSON)).Methods("GET")
```
sethvargo, line 413: This is already under the /stats parent route, so I think we can drop `-stats` from the path

sethvargo, line 413:
```suggestion
	r.Handle("/realm/key-server.json", c.HandleKeyServerStats(stats.StatsTypeJSON)).Methods("GET")
```

sethvargo (commented):
This seems fine. Devil's advocate question: why not merge these into the existing realm_stats JSON/CSV?

internal/routes/server.go
mikehelmick: broader point - should this be more under the realm stats that already exist?

sethvargo: See my review comment ;)

whaught: I think we'd just need to align dates for the Day field. It looks like the existing stats use timeutils.Midnight rather than UTCMidnight?

whaught: Just kidding `timeutils.Midnight(time.Now().UTC())` is the same thing

whaught: I've added in caching. At the moment I have them separate because they're stored in a different table so we'd need to re-do the json marshal to merge the two gorm structs to a single output struct. I figured the other stats were split out so this would be the easiest thing for now.

I also don't have a good solution for CSVing this because the distributions add an array dimension that's easy to do in json, but not CSV.

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -0,0 +1,118 @@
...
+    Key server stats
+  </div>
+  <div id="keyserver_chart_div" class="h-100 w-100" style="min-height:325px;">
+    <p class="text-center font-italic w-100 mt-5">Loading chart...</p>
+  </div>
+  <small class="card-footer d-flex justify-content-between text-muted">
+    <a href="#" data-toggle="modal" data-target="#keyserver-stats-modal">Learn more about this chart</a>
+    <span>
+      <span class="mr-1">Export as:</span>
+      <a href="/stats/realm/key-server.json">JSON</a>
```
sethvargo, line 15: No CSV?

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -0,0 +1,118 @@
...
+      let $div = $('#keyserver_chart_div');
+
+      if (!data.statistics) {
+        $div.find('p').text('No data yet.');
+        return;
+      }
+
+      var dataTable = new google.visualization.DataTable();
+      dataTable.addColumn('date', 'Date');
+      dataTable.addColumn('number', 'Total TEKs published');
```
sethvargo, line 78:
```suggestion
      dataTable.addColumn('number', 'TEKs published');
```

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -0,0 +1,118 @@
...
+      dataTable.addColumn('number', 'Requests with revisions');
```
sethvargo, line 79:
```suggestion
      dataTable.addColumn('number', 'Revision requests');
```

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -0,0 +1,118 @@
...
+      dataTable.addColumn('number', 'Request missing onset');
```
sethvargo, line 80:
```suggestion
      dataTable.addColumn('number', 'Requests missing onset');
```

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -0,0 +1,118 @@
...
+      let chart = new google.visualization.LineChart($div.get(0));
+      chart.draw(dataTable, options);
+    })
+    .fail(function(xhr, status, err) {
+      flash.error('Failed to render key-server stats: ' + err);
+    });
+  }
+</script>
+
+{{end}}
```
sethvargo, line 118: Nit: newline EOF

cmd/server/assets/realmadmin/stats.html
```diff
@@ -34,6 +34,8 @@ <h1>Realm stats</h1>
     </div>
 
     {{template "realmadmin/_stats_tokens" .}}
+
+    {{template "realmadmin/_stats_keyserver" .}}
```
sethvargo, line 38: iirc, this needs to be a realm-level configurable. We should only display this if the realm enabled it.

sethvargo (changes requested):
Everything looks good. I think we need to guard the display of the graph with the realm configurable as per APHL’s requirement.

internal/routes/server.go
sethvargo: I’d suggest joining them with a pipe character

cmd/server/assets/realmadmin/stats.html
whaught: the realm-enablement currently guards the stats-puller from storing data in the first place. Do we need it here too?

cmd/server/assets/realmadmin/_stats_keyserver.html
whaught: The distribution graphs (that I haven't added yet) have another array data dimension that's hard to format as CSV. Haven't done a csv marshaller yet

internal/routes/server.go
whaught: That works. esp since they're fixed-length. I might make that the next PR - this seems like a good place to stop.

cmd/server/assets/realmadmin/stats.html
whaught: added an explicit guard

sethvargo (approved):
/lgtm

PR #1662, opened 2021-01-22
removing unneeded Elevated500 alert

bschlaman:
## Proposed Changes

* remove Elevated500s alert
* remove associated Elevated500.md alerting playbook

**Release Note**

```release-note
removing unneeded alert
```


sethvargo (commented):
Why do we want to remove this? Elevated 500s indicate a problem with the service.

bschlaman:
> Why do we want to remove this? Elevated 500s indicate a problem with the service.

I believe the sentiment is that this is a duplicate of the SLO alerts and CloudSchedulerJobFailed.  @yuriatgoogle @yegle @mariliamelo is that correct?

yuriatgoogle:
Yes, that's correct.


*Yuri Grinshteyn*

Customer Reliability Engineering

thegrinch@google.com

PR #1668, opened 2021-01-23
add UI for managing SMS signing keys

mikehelmick:
towards #1640 

## Proposed Changes

* Add realm setting for enabling Authenticated SMS
* Add key management

image

**Release Note**

```release-note
Realm admins can create/rotate SMS signing keys and enable authenticated SMS.
```

### DIFF
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -141,6 +141,12 @@ aria-expanded="false" aria-label="Toggle navigation">
               {{t $.locale "nav.signing-keys"}}
             </a>
           {{end}}
+          {{if $currentMembership.Can rbac.SettingsRead}}
+            {{$showRealmMenu = true}}
+            <a class="dropdown-item {{if .currentPath.IsDir "/realm/smskeys"}}active{{end}}" href="/realm/smskeys">
+              {{t $.locale "nav.sms-signing-keys"}}
+            </a>
+          {{end}}
           {{if $currentMembership.Can rbac.StatsRead}}
             {{$showRealmMenu = true}}
             <a class="dropdown-item {{if .currentPath.IsDir "/realm/stats"}}active{{end}}" href="/realm/stats">
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -140,10 +140,7 @@ aria-expanded="false" aria-label="Toggle navigation">
             <a class="dropdown-item {{if .currentPath.IsDir "/realm/keys"}}active{{end}}" href="/realm/keys">
               {{t $.locale "nav.signing-keys"}}
             </a>
-          {{end}}
-          {{if .enableAuthenticatedSMS}}
-            {{if $currentMembership.Can rbac.SettingsRead}}
-              {{$showRealmMenu = true}}
+            {{if .enableAuthenticatedSMS}}
               <a class="dropdown-item {{if .currentPath.IsDir "/realm/smskeys"}}active{{end}}" href="/realm/smskeys">
                 {{t $.locale "nav.sms-signing-keys"}}
               </a>
--- a/cmd/server/assets/header.html
+++ b/cmd/server/assets/header.html
@@ -141,11 +141,13 @@ aria-expanded="false" aria-label="Toggle navigation">
               {{t $.locale "nav.signing-keys"}}
             </a>
           {{end}}
-          {{if $currentMembership.Can rbac.SettingsRead}}
-            {{$showRealmMenu = true}}
-            <a class="dropdown-item {{if .currentPath.IsDir "/realm/smskeys"}}active{{end}}" href="/realm/smskeys">
-              {{t $.locale "nav.sms-signing-keys"}}
-            </a>
+          {{if .enableAuthenticatedSMS}}
+            {{if $currentMembership.Can rbac.SettingsRead}}
+              {{$showRealmMenu = true}}
+              <a class="dropdown-item {{if .currentPath.IsDir "/realm/smskeys"}}active{{end}}" href="/realm/smskeys">
+                {{t $.locale "nav.sms-signing-keys"}}
+              </a>
+            {{end}}
           {{end}}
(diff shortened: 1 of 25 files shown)


cmd/server/assets/header.html
```diff
@@ -141,6 +141,12 @@
               {{t $.locale "nav.signing-keys"}}
             </a>
           {{end}}
+          {{if $currentMembership.Can rbac.SettingsRead}}
```
sethvargo, line 144: Nit, it'd probably be better to just lump this in with the existing "SettingsRead" rbac check on 138

internal/routes/server.go
```diff
@@ -393,6 +400,16 @@ func realmkeysRoutes(r *mux.Router, c *realmkeys.Controller) {
 	r.Handle("/keys/activate", c.HandleActivate()).Methods("POST")
 }
 
+// realmSMSkeysRoutes are the realm key routes.
+func realmSMSkeysRoutes(r *mux.Router, c *smskeys.Controller) {
+	r.Handle("/smskeys", c.HandleIndex()).Methods("GET")
```
sethvargo, line 405: Can we use `/sms-keys` instead please? It's a friendlier route. Will have to update any href links.

pkg/controller/smskeys/create_test.go
```diff
@@ -0,0 +1,110 @@
...
+	t.Run("create_first", func(t *testing.T) {
+		t.Parallel()
+
+		c, err := smskeys.New(ctx, cfg, harness.Database, harness.Cacher, h)
+		if err != nil {
+			t.Fatal(err)
+		}
+		handler := c.HandleCreateKey()
+
+		realm := database.NewRealmWithDefaults("test")
```
sethvargo, line 68: The harness has a default realm. Use FindRealm(1) instead (it's faster)

pkg/controller/smskeys/destroy_test.go
```diff
@@ -0,0 +1,134 @@
...
+		ctx = controller.WithSession(ctx, &sessions.Session{})
+		ctx = controller.WithMembership(ctx, &database.Membership{
+			Realm:       realm,
+			User:        &database.User{},
+			Permissions: rbac.SettingsWrite,
+		})
+
+		r := httptest.NewRequest("PUT", "/", nil)
+		r = r.Clone(ctx)
+		r = mux.SetURLVars(r, map[string]string{"id": fmt.Sprintf("%d", toDestroy)})
```
sethvargo, line 103: TIL

pkg/controller/smskeys/smskeys.go
```diff
@@ -0,0 +1,49 @@
...
+// Controller has handlers HTTP actions realted to SMS signing key management.
+type Controller struct {
+	config         *config.ServerConfig
+	db             *database.Database
+	h              render.Renderer
+	publicKeyCache *keyutils.PublicKeyCache
+}
+
+// New creates a new Controller
+func New(ctx context.Context, config *config.ServerConfig, db *database.Database, cacher cache.Cacher, h render.Renderer) (*Controller, error) {
```
sethvargo, line 37: Consider passing in the public key cache as a parameter instead of building it in the controller. That removes the need to pass in a context to New and removes the need to return an error.

pkg/database/sms_signing_key.go
```diff
@@ -39,7 +39,7 @@ type SMSSigningKey struct {
 
 // GetKID returns the 'kid' field value to use in signing JWTs.
 func (s *SMSSigningKey) GetKID() string {
-	return fmt.Sprintf("r%dmv%d", s.RealmID, s.ID)
+	return fmt.Sprintf("r%dv%ds", s.RealmID, s.ID)
```
sethvargo, line 42: Should we be more explicit and just say "sms" in here?

pkg/controller/smskeys/render.go
```diff
@@ -0,0 +1,72 @@
...
+func (c *Controller) redirectShow(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	http.Redirect(w, r, "/realm/smskeys", http.StatusSeeOther)
+}
+
+func (c *Controller) renderShow(ctx context.Context, w http.ResponseWriter, r *http.Request, realm *database.Realm) {
+	m := controller.TemplateMapFromContext(ctx)
+	m.Title("Authenticated SMS Signing keys")
+	m["realm"] = realm
+
+	keys, err := realm.ListSMSSigningKeys(c.db)
```
sethvargo, line 36: This should be cached (if we don't have a ListSMSSigningKeysCached, we should make one)

pkg/controller/smskeys/activate.go
```diff
@@ -0,0 +1,68 @@
...
+		var form FormData
+		if err := controller.BindForm(w, r, &form); err != nil {
+			flash.Error("Failed to process form: %v", err)
+			c.renderShow(ctx, w, r, currentRealm)
+			return
+		}
+
+		kid, err := currentRealm.SetActiveSMSSigningKey(c.db, form.SigningKeyID)
+		if err != nil {
+			flash.Error("Unable to set active SMS signing key: %v", err)
```
sethvargo, line 60: These should really follow the same form as the other errors:

```go
currentRealm.AddError("", fmt.Errorf("Unable to set ...: %w", err))
w.WriteHeader(http.StatusUnprocessableEntity)
// ...
```

sethvargo (changes requested):

pkg/controller/smskeys/render.go
mikehelmick: I don't think we want to cache in this instance - this is the admin screen.
We do cache the public keys, so we're not hitting KMS on every refresh.

pkg/controller/smskeys/smskeys.go
mikehelmick: did same for realmkeys

pkg/database/sms_signing_key.go
mikehelmick: sure

PR #1670, opened 2021-01-23
Redraw stats charts on window resize

whaught:
## Proposed Changes

* The charts don't resize automatically and need to be resized with the window
* The resize event fires as the window is dragged, delay we only redraw every 300ms

### DIFF
--- a/cmd/server/assets/realmadmin/_stats_codes.html
+++ b/cmd/server/assets/realmadmin/_stats_codes.html
@@ -177,6 +177,10 @@
 
       dashboard.bind(filter, realmChart);
       dashboard.draw(dataTable);
+      redrawCharts.push({
+        chart: dashboard,
+        data: dataTable,
+      });
     })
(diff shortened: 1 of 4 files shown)


mikehelmick (approved):

sethvargo:
We used to do this, and I reverted it. I was able to rate-limit myself rather quickly just by dragging by window. We need to add some kind of intelligent debouncing so that we don't redraw once for each pixel moved.

sethvargo:
@whaught ^

whaught:
> We used to do this, and I reverted it. I was able to rate-limit myself rather quickly just by dragging by window. We need to add some kind of intelligent debouncing so that we don't redraw once for each pixel moved.

In this PR, the redraw event sets a `pendingRedraw=true` bit and then a timeout to fire in 300ms for the actual redraw. We also compare to prior horizontal width. The idea was to discard all of the intermediate events or vertical changes so we don't attempt this for every pixel.

PR #1674, opened 2021-01-24
Move and cache SMS provider up one level

sethvargo:
For batch, this creates N providers (and N TCP connections) unnecessarily. This also adds a local in-memory cache of the provider for 5 minutes to alleviate lookups during heavy load. This is a local in-memory cache, since we can't serialize TCP connections into Redis.

This will become more important for SMS signing keys, since creating the signer is expensive.

**Release Note**

```release-note
Use the same SMS provider for all messages, cache locally for 5 minutes to improve performance
```

/assign @whaught

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -16,12 +16,15 @@ package issueapi
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
+	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/api"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
+	"github.com/google/exposure-notifications-verification-server/pkg/sms"
 	"go.opencensus.io/tag"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controller/issueapi/issue.go
```diff
@@ -70,23 +74,31 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*api.IssueCodeReq
...
-		}
+	defer c.recordStats(ctx, results)
 
-		wg.Add(1)
-		go func(request *api.IssueCodeRequest, r *IssueResult) {
-			defer wg.Done()
-			c.SendSMS(ctx, request, r, realm)
-		}(requests[i], result)
+	// Send SMS messages if there's an SMS provider.
+	smsProvider, err := c.smsProviderFor(ctx, realm)
```
whaught, line 80: ah one sms provider here... why didn't I think of this.

whaught (approved):

pkg/controller/issueapi/issue.go
whaught: this is lgtm, but there is an unhappy unit test

whaught: I'm now wondering if a pool of connections (rather than one or always making new ones) is what we really want

sethvargo: Go's HTTP client re-uses connections, and Twilio is HTTP/2. Given we didn't see the system fall over with batch upload when creating entirely new connections for each request, I don't anticipate we need connection pooling for the size/number of requests. Connection pooling tends to introduce a bunch of complexity (see database layer with idle timeouts, max timeouts, etc) that seems like unnecessary overhead for this until we actually need it 😄

sethvargo:
@whaught fixed up the failing test.

whaught (approved):

PR #1675, opened 2021-01-25
Code issue to claim distribution stat

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1231

## Proposed Changes

* Adds new DB array for issue-claim age distribution stat
* Adds running average issue-claim age stat
* Marshal to CSV and json
* Record bucket when updating code-claim stats
* Record running avarage

**Release Note**

```release-note
Add code issue to claim age average and distribution stat
```

### DIFF
--- a/pkg/controller/issueapi/validate_code.go
+++ b/pkg/controller/issueapi/validate_code.go
@@ -39,6 +39,7 @@ func (c *Controller) BuildVerificationCode(ctx context.Context, request *api.Iss
 		RealmID:           realm.ID,
 		IssuingExternalID: request.ExternalIssuerID,
 		TestType:          strings.ToLower(request.TestType),
+		IssuedAt:          now,
 		ExpiresAt:         now.Add(realm.CodeDuration.Duration),
 		LongExpiresAt:     now.Add(realm.LongCodeDuration.Duration),
 	}
--- a/pkg/controller/issueapi/validate_code.go
+++ b/pkg/controller/issueapi/validate_code.go
@@ -39,7 +39,6 @@ func (c *Controller) BuildVerificationCode(ctx context.Context, request *api.Iss
 		RealmID:           realm.ID,
 		IssuingExternalID: request.ExternalIssuerID,
 		TestType:          strings.ToLower(request.TestType),
-		IssuedAt:          now,
 		ExpiresAt:         now.Add(realm.CodeDuration.Duration),
 		LongExpiresAt:     now.Add(realm.LongCodeDuration.Duration),
 	}
--- a/pkg/database/realm_stats.go
+++ b/pkg/database/realm_stats.go
@@ -34,6 +34,11 @@ var _ icsv.Marshaler = (RealmStats)(nil)
 // RealmStats represents a logical collection of stats of a realm.
 type RealmStats []*RealmStat
 
+var claimDistributionBuckets = []time.Duration{
+	time.Minute, 5 * time.Minute, 25 * time.Minute, 30 * time.Minute, time.Hour,
+	2 * time.Hour, 3 * time.Hour, 6 * time.Hour, 12 * time.Hour, 24 * time.Hour, 336 * time.Hour,
+}
+
 // RealmStat represents statistics related to a user in the database.
 type RealmStat struct {
 	Date    time.Time `gorm:"column:date; type:date; not null;"`
--- a/pkg/database/realm_stats.go
+++ b/pkg/database/realm_stats.go
@@ -31,9 +31,6 @@ import (
 
 var _ icsv.Marshaler = (RealmStats)(nil)
 
-var codeClaimBuckets []time.Duration = []time.Duration{
-	time.Minute, 5 * time.Minute, 15 * time.Minute, 30 * time.Minute, time.Hour, 2 * time.Hour, 3 * time.Hour, 6 * time.Hour, 12 * time.Hour, 24 * time.Hour}
-
 // RealmStats represents a logical collection of stats of a realm.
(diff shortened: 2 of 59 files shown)


pkg/database/vercode_test.go
```diff
@@ -447,7 +447,7 @@ func TestStatDates(t *testing.T) {
 	db, _ := testDatabaseInstance.NewDatabase(t, nil)
 	realm := NewRealmWithDefaults("Test Realm")
 
-	now := time.Now()
+	now := time.Now().UTC()
```
whaught, line 450: This was a bug. This test fails in the evening.

pkg/database/token.go
```diff
@@ -336,23 +336,57 @@ func (db *Database) updateStatsCodeInvalid(t time.Time, authApp *AuthorizedApp)
...
+			Error; err != nil && !IsNotFound(err) {
+			return err
+		}
+
+		existing.CodesClaimed++
+		if len(existing.CodeClaimDistribution) == 0 {
+			existing.CodeClaimDistribution = make([]int32, len(claimDistributionBuckets))
+		}
+
+		sinceIssue := t.Sub(vc.IssuedAt)
```
whaught, line 365: We could try avoid the missing data by instead trying to subtract from expiry (assuming it hasn't changed) but adding explicit IssuedAt feels more foolproof

pkg/database/token.go
```diff
@@ -397,14 +397,31 @@ func (db *Database) updateStatsTokenInvalid(t time.Time, authApp *AuthorizedApp)
 func (db *Database) updateStatsTokenClaimed(t time.Time, authApp *AuthorizedApp) {
 	t = timeutils.UTCMidnight(t)
 
-	realmSQL := `
+	if err := db.db.Transaction(func(tx *gorm.DB) error {
+		var existing RealmStat
+		if err := tx.
```
sethvargo: I'm not sure I see how this is different from the existing implementation?

pkg/database/vercode_test.go
sethvargo: Nice. I've hit a few of those when I run tests late at night too.

pkg/controller/issueapi/validate_code.go
```diff
@@ -39,6 +39,7 @@ func (c *Controller) BuildVerificationCode(ctx context.Context, request *api.Iss
 		RealmID:           realm.ID,
 		IssuingExternalID: request.ExternalIssuerID,
 		TestType:          strings.ToLower(request.TestType),
+		IssuedAt:          now,
```
sethvargo, line 42: Wouldn't this just be `CreatedAt` (from the gorm.Model)?

pkg/database/realm_stats.go
```diff
@@ -21,17 +21,24 @@ import (
...
+	"github.com/lib/pq"
 )
 
 var _ icsv.Marshaler = (RealmStats)(nil)
 
 // RealmStats represents a logical collection of stats of a realm.
 type RealmStats []*RealmStat
 
+var claimDistributionBuckets = []time.Duration{
+	time.Minute, 5 * time.Minute, 25 * time.Minute, 30 * time.Minute, time.Hour,
```
sethvargo, line 38: Do we really expect people to care about 25 vs 30 min?

pkg/database/token.go
```diff
@@ -336,23 +336,57 @@ func (db *Database) updateStatsCodeInvalid(t time.Time, authApp *AuthorizedApp)
...
-		db.logger.Errorw("failed to update authorized app stats", "error", err)
+		db.logger.Errorw("failed to update authorized app stats code invalid", "error", err)
 	}
 }
 
 // updateStatsCodeClaimed updates the statistics, increasing the number of codes
 // claimed.
-func (db *Database) updateStatsCodeClaimed(t time.Time, authApp *AuthorizedApp) {
-	t = timeutils.UTCMidnight(t)
+func (db *Database) updateStatsCodeClaimed(t time.Time, authApp *AuthorizedApp, vc *VerificationCode) {
```
sethvargo, line 345: I think this function could be simplified by updating the two stats independently, potentially even updating CodeClaimDistribution in another function.

Similarly, this will be faster and less error-prone to do it in straight SQL:

```sql
WITH arr AS (
  -- this is the fun part. you could build this with case statements or use a 
  -- range function: https://www.postgresql.org/docs/9.3/rangetypes.html
)
INSERT INTO realm_stats(date, realm_id, code_claim_distribution
  VALUES ($1, $2, arr)
ON CONFLICT (date, realm_id) DO UPDATE
  SET code_claim_distribution = array_agg(UNNEST(arr) + UNNEST(realm_stats.code_claim_distribution))
```

pkg/database/realm_stats.go
sethvargo: In general from a stats perspective, these buckets are a little weird.

pkg/controller/issueapi/validate_code.go
whaught: Yes. I forgot about the gorm.Model fields. it was late

pkg/database/realm_stats.go
whaught: that's supposed to be 15m. True log scale would be more statistically sound

pkg/database/token.go
whaught: the existing implementation does a blind-write. I wanted to read the data so I can manipulate the existing values. Arrays are hard to increment an index in pure SQL and I have this average calc too.

sethvargo (approved):

PR #1677, opened 2021-01-25
separating notification channels -paging and non-paging

yuriatgoogle:
Fixes #

## Proposed Changes

* Added paging and non-paging notification channels
* Specified paging alerts for SLO and latency violations
* Specified paging and non-paging alerts 

**Release Note**

```release-note
Creates separate paging and non-paging alert mechanisms.
```

### DIFF
--- a/terraform/alerting/notifications.tf
+++ b/terraform/alerting/notifications.tf
@@ -21,5 +21,5 @@ resource "google_monitoring_notification_channel" "non-paging" {
     null_resource.manual-step-to-enable-workspace
   ]
 
-  for_each = var.alert-notification-channel-nonpaging
+  for_each = var.alert-notification-channel-non-paging
 }
--- a/terraform/alerting/notifications.tf
+++ b/terraform/alerting/notifications.tf
@@ -11,7 +11,7 @@ resource "google_monitoring_notification_channel" "paging" {
   for_each = var.alert-notification-channel-paging
 }
 
-resource "google_monitoring_notification_channel" "nonpaging" {
+resource "google_monitoring_notification_channel" "non-paging" {
   provider     = google-beta
   project      = var.project
   display_name = "Non-paging Notification Channel"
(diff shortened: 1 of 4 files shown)


terraform/alerting/notifications.tf
```diff
@@ -1,12 +1,25 @@
...
   labels       = each.value.labels
   depends_on = [
     null_resource.manual-step-to-enable-workspace
   ]
 
-  for_each = var.alert-notification-channels
+  for_each = var.alert-notification-channel-paging
+}
+
+resource "google_monitoring_notification_channel" "nonpaging" {
```
sethvargo, line 14: Nit: It's standard to use dashes for separators in resource names (e.g. this should be `"non-paging"`)

yuriatgoogle: Addressed, thank you.

terraform/alerting/notifications.tf
```diff
@@ -1,12 +1,25 @@
...
+  provider     = google-beta
+  project      = var.project
+  display_name = "Non-paging Notification Channel"
+  type         = each.key
+  labels       = each.value.labels
+  depends_on = [
+    null_resource.manual-step-to-enable-workspace
+  ]
+
+  for_each = var.alert-notification-channel-nonpaging
```
mariliamelo, line 24: should this be `non-paging` as well?

terraform/alerting/variables.tf
```diff
@@ -40,12 +40,30 @@ variable "alert-notification-channels" {
...
+        channel_name = "#paging-channel"
+        auth_token   = "paging-token"
       }
     }
   }
-  description = "Notification channels"
+  description = "Paging notification channel"
+}
+
+variable "alert-notification-channel-nonpaging" {
```
mariliamelo, line 51: `non-paging` ?

mariliamelo (commented):
Let's use only `non-paging` then

terraform/alerting/notifications.tf
yuriatgoogle: Yes, thank you. Done.

terraform/alerting/variables.tf
yuriatgoogle: Done, thank you.

PR #1688, opened 2021-01-26
My dockerfiles work with this change

whaught:
## Proposed Changes

* My docker builds fail to retrieve assets. This gives them all owner as they copy
* Caveat: I have no idea what I'm doing

/hold


sethvargo (approved):
@whaught can you make the same changes in the key server?

whaught:
/hold cancel

I'll do the same for key

PR #1691, opened 2021-01-26
Key server stats marshal to CSV

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1512

## Proposed Changes

* Unpins key-serve dep to main to use https://github.com/google/exposure-notifications-server/pull/1307
* Adds CSV routes and marshaller option

**Release Note**

```release-note
Add CSV format for key-server stats file
```

PR #1692, opened 2021-01-26
csv link in the ui

whaught:
## Proposed Changes

* https://github.com/google/exposure-notifications-verification-server/pull/1691 created the routes to do CSV marshal and this sticks the link into the UI
    * (sorry for the absurdly small PRs - I missed a spot here)
* `target="_blank"` for new tab when downloading as a file

### DIFF
--- a/cmd/server/assets/apikeys/show.html
+++ b/cmd/server/assets/apikeys/show.html
@@ -86,8 +86,8 @@
         </span>
         <span>
           <span class="mr-1">Export as:</span>
-          <a href="/stats/realm/api-keys/{{$authApp.ID}}.csv" class="mr-1">CSV</a>
-          <a href="/stats/realm/api-keys/{{$authApp.ID}}.json">JSON</a>
+          <a href="/stats/realm/api-keys/{{$authApp.ID}}.csv" target="_blank" class="mr-1">CSV</a>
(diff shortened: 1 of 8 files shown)


cmd/server/assets/apikeys/show.html
```diff
@@ -86,8 +86,8 @@ <h1>{{$authApp.Name}} API key</h1>
         </span>
         <span>
           <span class="mr-1">Export as:</span>
-          <a href="/stats/realm/api-keys/{{$authApp.ID}}.csv" class="mr-1">CSV</a>
-          <a href="/stats/realm/api-keys/{{$authApp.ID}}.json">JSON</a>
+          <a href="/stats/realm/api-keys/{{$authApp.ID}}.csv" target="_blank" class="mr-1">CSV</a>
+          <a href="/stats/realm/api-keys/{{$authApp.ID}}.json" target="_blank">JSON</a>
```
sethvargo, line 90: Why target blank? The CSV should trigger as a file download.

whaught: That's true it may be redundant for CSV, but for JSON it takes over the window

sethvargo (approved):
/lgtm

PR #1698, opened 2021-01-26
Seven day totals

whaught:

## Proposed Changes

* The distribution tables now sum the last 7 days from that slider stat date
    * This should provide smoother charts esp if there is day-of-the-week variance in user behavior

**Release Note**

```release-note
Show distributions as 7 day sums
```


sethvargo (approved):
/lgtm

PR #1700, opened 2021-01-27
Move redrawer to static js. Apply to the remaining charts

whaught:
## Proposed Changes

* Moves the redraw charts function to static js so it'll be cached. I think we can do a lot more of that with some of these charts as things settle down
* Apply redraw to the users and keys charts too

**Release Note**

```release-note
Redraw all charts with window resize. Move some js to application.js
```

### DIFF
--- a/cmd/server/assets/static/js/application.js
+++ b/cmd/server/assets/static/js/application.js
@@ -1021,3 +1021,34 @@ function showSuccessfulCode(request, code, line) {
   $row.append($('<td/>').text(code.uuid));
   $successTableBody.append($row);
 }
+
+function redrawCharts(chartsData, timeout) {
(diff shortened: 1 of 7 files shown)


cmd/server/assets/static/js/application.js
```diff
@@ -1021,3 +1021,34 @@ function showSuccessfulCode(request, code, line) {
...
+  function redraw() {
+    let c;
+    for (c of chartsData) {
+      if (c.options) {
+        c.options.animation = null;
+      }
+      c.chart.draw(c.data, c.options);
+    }
+  }
+}
```
sethvargo, line 1054: Newline

sethvargo (approved):
/lgtm

PR #1708, opened 2021-01-27
Fix bad column name for new stats

whaught:
## Proposed Changes

* This column should be named correctly
* Made another migration line so we can rerun on dev


mikehelmick (approved):

mikehelmick (approved):

sethvargo (commented):
We already ran the migration though...

whaught:
> We already ran the migration though...

This makes a new migration 00093

PR #1712, opened 2021-01-28
Refactor Terraform secrets and key management

sethvargo:
This simplifies our key and secret management management at the Terraform level. We've had a few partial outages when deploying new features because it was previously very easy to mess up. This refactors things to make it clearer which secrets a service needs and to depend on the service's service account having the permission as a dependency.

It also fixes a continuous diff we've been getting with Cloud Run.

**Release Note**

```release-note
Simplify key and secret management in Terraform. Note: this will cause a _very_ large Terraform diff when applying, including deleting IAM bindings. We recommend running `terraform apply` twice to ensure convergence.
```


whaught (approved):
I'm not a terraform expert, but I like the local var grouping and deleted code.

PR #1713, opened 2021-01-28
Add a unit test for issue-claim stat update

whaught:
## Proposed Changes

* Adds a test to check that the bucketing and average calculation are being saved properly
* Updates a condition with smoothing (off by one)


sethvargo (approved):
/lgtm

PR #1718, opened 2021-01-28
Add graph controls and fix axes

mikehelmick:
## Proposed Changes

* Add tickmarks to histogram sliders
* Fix the vertical axis height so that the whole graph doesn't redraw as you use slider
   * TODO - make this work after changing smoothing

image

**Release Note**

```release-note
Key server histograms have better controls and don't resize as days change
```

### DIFF
--- a/cmd/server/assets/realmadmin/_stats_keyserver.html
+++ b/cmd/server/assets/realmadmin/_stats_keyserver.html
@@ -27,7 +27,7 @@
   </div>
   <div class="slidecontainer" id="tek_slide_div">
   </div>
-
+  <datalist id="tektickmarks"></datalist>
   <div class="card-body">
     <h5 class="card-title">Onset-to-upload distribution <span class="sum-title">(7 day sum)</span></h5>
   </div>
@@ -36,7 +36,7 @@
   </div>
   <div class="slidecontainer" id="onset_slide_div">
   </div>
-
+  <datalist id="onset_slide_ticks"></datalist>
   <small class="card-footer d-flex justify-content-between text-muted">
     <a href="#" data-toggle="modal" data-target="#keyserver-stats-modal">Learn more about this chart</a>
     <span>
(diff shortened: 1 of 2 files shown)


cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -100,10 +100,26 @@ <h5 class="modal-title">Key server stats</h5>
...
       $slider.val(data.length);
       $onsetSlider.attr('min', Math.min(smoothing, data.length));
       $onsetSlider.attr('max', data.length);
+      $onsetSlider.attr('value',data.length);
+      $onsetSlider.attr('step', 1);
       $onsetSlider.val(data.length);
+  
+      $("#tektickmarks").empty();
+      for (i = Math.min(smoothing, data.length); i < data.length; i++) {
+        let opt = '<option value="' + i + '" label="' + data[i].day.substring(0,10) + '"></option>';
```
whaught, line 114: I think its preferred to make an option and mess with it's features one-by-one
```
let $opt = $('<option/>');
$opt.val(i);
$opt.attr('label', substring);

foo.append($opt);
```

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -100,10 +100,26 @@ <h5 class="modal-title">Key server stats</h5>
...
+        $("#tektickmarks").append(opt);
+      }
+
+      $("#onset_slide_ticks").empty();
```
whaught, line 118: every time you do `$('#onset_slide_ticks')` jquery is retrieving it from the DOM. I'd just make this once and reference a variable here.

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -308,18 +338,26 @@ <h5 class="modal-title">Key server stats</h5>
...
 
     if (!data.length) {
       $div.find('p').text('No data yet.');
       return;
     }
     stats = data;
 
+    // calculate the vaxis max.
+    for (i = 0; i < stats.length - smoothing; i++) {
+      getOnsetDataTable(i);
```
whaught, line 358: ah now 'getOnsetDataTable' creates a bunch of new google.visualization.DataTable() to be immediately thrown away (to populate the global variable). This is getting messy.

If we want to do this maybe we should cache the results so we don't need to rebuild these as we slide the slider.

whaught (approved):

cmd/server/assets/realmadmin/_stats_keyserver.html
mikehelmick: added caches - redraws are faster now!

whaught (approved):

PR #1724, opened 2021-01-29
Styling cards for the statistics page

whaught:
## Proposed Changes

* Give every chart a boostrap card
* Show download links on headers
* Float right the smoothing options
* Give each chart its own explanation

boxes

**Release Note**

```release-note
Styling updates for statistics page & about links
```


sethvargo (approved):
/lgtm

but can we add a little horizontal padding (maybe px-3) to the sliders in a follow-up

PR #1734, opened 2021-01-29
Move SMS template into SMS tab

sethvargo:
This is a straight copy-paste of HTML and as little changes as possible for tests and controller implementation.

**Release Note**

```release-note
Move SMS templates into the SMS tab (originally under codes)
```


whaught (approved):
I had been considering doing this too. WA at one point said it was confusing.

PR #1738, opened 2021-01-29
Feedback on chart titles

whaught:
## Proposed Changes

* This addresses user-feedback on clarifying these chart titles
*  I'm not certain that these are definitely clearer, but feel free to suggest

### DIFF
--- a/cmd/server/assets/realmadmin/_stats_keyserver.html
+++ b/cmd/server/assets/realmadmin/_stats_keyserver.html
@@ -17,7 +17,7 @@
 <div class="card shadow-sm mb-3">
   <div class="card-header">
     <span class="oi oi-bar-chart mr-2 ml-n1"></span>
-    Publish requests (total number of TEKs published)
+    TEK Publish Metrics
   </div>
   <div id="keyserver_chart_div" class="h-100 w-100" style="min-height:325px;">
     <p class="text-center font-italic w-100 mt-5">Loading chart...</p>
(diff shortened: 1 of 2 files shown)


cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -17,7 +17,7 @@ <h6 class="float-right">
 <div class="card shadow-sm mb-3">
   <div class="card-header">
     <span class="oi oi-bar-chart mr-2 ml-n1"></span>
-    Publish requests
+    Publish requests (total number of TEKs published)
```
mikehelmick, line 20: this chat I think we need to separate - because the units are the same.

How about

`TEK Publish Metrics`

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -43,7 +43,7 @@ <h6 class="float-right">
 <div class="card shadow-sm mb-3">
   <div class="card-header">
     <span class="oi oi-bar-chart mr-2 ml-n1"></span>
-    TEK age distribution <span class="sum-title">(7 day sum)</span>
+    TEK number of days-old distribution <span class="sum-title">(7 day sum)</span>
```
mikehelmick, line 46: EN Days Active Before Upload?

cmd/server/assets/realmadmin/_stats_keyserver.html
```diff
@@ -134,14 +134,14 @@ <h5 class="modal-title">Publish requests by OS</h5>
   <div class="modal-dialog modal-dialog-centered">
     <div class="modal-content">
       <div class="modal-header">
-        <h5 class="modal-title">TEK age distribution</h5>
+        <h5 class="modal-title">TEK number of days-old distribution</h5>
```
mikehelmick, line 137: EN Days Active Before Upload?

mikehelmick (approved):

PR #1740, opened 2021-01-30
UTC time for test

sethvargo:
/assign @whaught 

**Release Note**

```release-note
Use UTC time in stats-puller test
```


mikehelmick (approved):

whaught:
Huh I was just griping about this in another test... and then made the same mistake

PR #1742, opened 2021-01-30
Set align rate to 1min

mariliamelo:
This should only notify once for every time the cloud scheduler job fails.


sethvargo (approved):
/lgtm

worth a shot

PR #1746, opened 2021-02-01
Test render of all the code pages

whaught:
## Proposed Changes

* Render code and assert expiry button is shown
* Render code + click expire + assert expiry button not shown
* Render index

**Release Note**

```release-note
Basic coverage for /codes pages
```


sethvargo (approved):
/lgtm

PR #1747, opened 2021-02-01
Fallback to a sentinel string if the redis auth is unset

sethvargo:
**Release Note**

```release-note
Fallback to a sentinel string if the redis auth is unset. This fixes an issue where Secret Manager fails to create the secret because it has a null value with Redis authentication is disabled.
```


mikehelmick:
/hold

services shouldn't have the redis auth env vars set if redis auth is disabled

whaught (approved):

PR #1751, opened 2021-02-02
don't pass redis password when auth disabled

mikehelmick:
## Proposed Changes

* Ensures that redis PW isn't passed when auth disabled

**Release Note**

```release-note
Allow for redis auth to be disabled and terraform to apply successfully
```

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -191,7 +191,7 @@ export FIREBASE_STORAGE_BUCKET="${data.google_firebase_web_app_config.default.st
 export CACHE_TYPE="REDIS"
 export CACHE_REDIS_HOST="${google_redis_instance.cache.host}"
 export CACHE_REDIS_PORT="${google_redis_instance.cache.port}"
-export CACHE_REDIS_PASSWORD="secret://${google_secret_manager_secret_version.redis-auth.id}"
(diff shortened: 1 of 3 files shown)


sethvargo (commented):
There's another redis_passwird field and the local env file too, but this approach is better

mikehelmick:
updated

terraform/main.tf
```diff
@@ -191,7 +191,7 @@ export FIREBASE_STORAGE_BUCKET="${data.google_firebase_web_app_config.default.st
 export CACHE_TYPE="REDIS"
 export CACHE_REDIS_HOST="${google_redis_instance.cache.host}"
 export CACHE_REDIS_PORT="${google_redis_instance.cache.port}"
-export CACHE_REDIS_PASSWORD="secret://${google_secret_manager_secret_version.redis-auth.id}"
+export CACHE_REDIS_PASSWORD=var.redis_enable_auth ? "secret://${google_secret_manager_secret_version.redis-auth.id}" : ""
```
sethvargo, line 194: There's also Rate Limit

sethvargo (changes requested):

sethvargo (approved):

PR #1754, opened 2021-02-02
Add more admin tests

sethvargo:
This increases the test coverage on the admin pages significantly. There's still quite a few cases that are very difficult to test though.

**Release Note**

```release-note
Add more admin tests
```


pkg/controller/admin/users_list.go
```diff
@@ -29,6 +29,12 @@ func (c *Controller) HandleUsersIndex() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
+		session := controller.SessionFromContext(ctx)
```
whaught, line 32: Do we need to duplicate this for every handler? Middleware shouldn't let you past RequireAuth without session?

sethvargo: We don't, and it's a lot of repetition, but it's specifically checking to make sure we don't panic if the middleware is not installed correctly.

whaught (approved):

PR #1762, opened 2021-02-03
Double frequency for stats

whaught:
## Proposed Changes

* These are very low-cost calls, but could keep things up-to-date if a call fails
* Key-server only updates this once per hour, most of our distributions show daily


sethvargo (approved):
/lgtm

PR #1763, opened 2021-02-03
Fix data race

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/database/config.go
+++ b/pkg/database/config.go
@@ -96,6 +96,50 @@ func (c *Config) ConnectionString() string {
 	return strings.Join(p, " ")
 }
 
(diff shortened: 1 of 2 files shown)


mikehelmick (approved):

pkg/database/config.go
```diff
@@ -96,6 +96,50 @@ func (c *Config) ConnectionString() string {
 	return strings.Join(p, " ")
 }
 
+// clone creates a deep copy of the configuration.
```
whaught, line 99: I worry about forgetting to update this if we ever add a field to `Config`. Thoughts about using a deepcopy library? Maybe github.com/jinzhu/copier

Issue #1771, opened 2021-02-04
VS Portal Session does not expire if page is closed


bschlaman:
### TL;DR
While having the VS portal UI open, the session will expire after SESSION_IDLE_TIMEOUT minutes of inactivity (20 min default).  However, if the page is closed, the session seems to remain active for an indefinite amount of time.

**Expected behavior**
When navigating to the portal after it was closed out for more than 20 min, the user should be prompted to log in again.

**Observed behavior**
The login screen flashes for a moment, and the user is automatically logged in.

### Reproduction
1. Log in normally
2. Close out of the portal
3. Wait >20min (can be overnight)
4. Navigate to the login page, user is automatically logged in

**Environment**
- OS: Windows, Linux
- Browser: Chrome, Chromium

**Additional information**
Not sure if this happens every time or just some of the time.

whaught:
I don't believe the session will remain active indefinitely - `SESSION_DURATION` sets the ttl to 20h by default. Let me check to see if we're bumping activity on the session before doing the idle check

bschlaman:
I'm pretty sure this has happened with time >20hrs, although I can't confirm that for sure.  I closed out of a session earlier today, so I'll test this tomorrow after 20hrs have passed.

PR #1772, opened 2021-02-04
Check session idleness in login

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1771

## Proposed Changes

* Checks the session for idleness even if auth is not required. This is important because requireSession defers a bump to session freshness (so the login page currently refreshes the session).

**Release Note**

```release-note
Ensure the login page doesn't refresh session freshness
```


sethvargo (approved):
/lgtm

I wonder if we could separate idle checking from auth entirely.

Issue #1770, opened 2021-02-04
Terraform deployment: `stats-puller` fails, missing `KEY_SERVER_URL` value


krazykid:
### Question
I'm trying to deploy the verification server with terraform and it errors out with the following error:

```
google_cloud_run_service.stats-puller: Still creating... [10s elapsed]

Error: Error waiting to create Service: resource is in failed state "Ready:False", message: Cloud Run error: Container failed to start. Failed to start and then listen on the port defined by the PORT environment variable. Logs for this revision might contain more information.

Logs URL:
https://console.cloud.google.com/logs/viewer?project=ma-verification-qvtby51btd&resource=cloud_run_revision/service_name/stats-puller/revision_name/stats-puller-jd2v6&advancedFilter=resource.type%3D%22cloud_run_revision%22%0Aresource.labels.service_name%3D%22stats-puller%22%0Aresource.labels.revision_name%3D%22stats-puller-jd2v6%22
```

I looked at the Cloud Run console and the service logs, and see the following:

![Screen Shot 2021-02-04 at 9 59 53 AM](https://user-images.githubusercontent.com/955049/106910971-bc660c80-66cf-11eb-899b-2bf1962776bd.png)

and:

![Screen Shot 2021-02-04 at 10 00 18 AM](https://user-images.githubusercontent.com/955049/106911035-ca1b9200-66cf-11eb-8cf7-791d299c39f4.png)

I'm not sure where or how to set this value (environment variable? TF var?). Or what its value should be. Any help would be greatly appreciated.

sethvargo:
You can set the environment variable in Terraform. If you're using a module, for example

```hcl
service_environment = {
  stats-puller = {
    KEY_SERVER_URL = "https://exposure.key.apollo-project.org"
  }
}
```

whaught:
Specifically this environment variable points to the location of the associated key-server which Verification can use to retrieve statistics about code publish. This URL is set on the server, but each realm must opt-in via realm settings for it to be active. The variable is marked `required` with no default value, but if you are just testing and don't intend to pull key-server statistics, you could put any string value in (although that service would fail if any realm opts in).

I'll find a spot to add this to documentation.

krazykid:
@whaught, I'm sorry, but I still don't track what you mean by the `key-server`. Is this the `server` Cloud Run service in  `exposure-notifications-verification-server` Cloud Run stack? Or is it a service in `exposure-notifications-server` Cloud Run stack?

[...]

PR #1776, opened 2021-02-05
Update example tf

whaught:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1770

## Proposed Changes

* Update terraform example

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -195,29 +195,29 @@ Terraform module.
     inputs/parameters to the module definition. See the `variables.tf` file for
     the full list of configuration options.
 
-2.  (Optional if alerting is enabled): Manually create Google Cloud Monitoring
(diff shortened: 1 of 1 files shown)


terraform/README.md
```diff
@@ -186,29 +195,29 @@ Terraform module.
     inputs/parameters to the module definition. See the `variables.tf` file for
     the full list of configuration options.
 
-1.  (Optional if alerting is enabled): Manually create Google Cloud Monitoring
```
sethvargo, line 189: You can leave these all as "1". Markdown does the right thing.

sethvargo (approved):
/hold

terraform/README.md
whaught: hmm must be my markdown vs plugin

sethvargo (approved):
/lgtm

PR #1782, opened 2021-02-08
Add custom tooltips for key-server stats

whaught:
## Proposed Changes

* Custom tooltips for the key-server stats charts
    * Specifically shows the % of total for each bar

Screen Shot 2021-02-08 at 10 48 12 AM

**Release Note**

```release-note
Custom chart tooltips with % of total
```

### DIFF
--- a/cmd/server/assets/realmadmin/_stats_keyserver.html
+++ b/cmd/server/assets/realmadmin/_stats_keyserver.html
@@ -309,11 +309,17 @@
 
       dataTable.addRow([utcDate(row.day),
         row.publish_requests.unknown,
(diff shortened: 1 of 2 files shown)


sethvargo (commented):
We trust the source, but this is _technically_ vulnerable to XSS. If someone could get the JSON endpoint to serve a string for `count`, that would be rendered as HTML on the page. Can we sanitize the input with jQuery?

whaught:
> We trust the source, but this is _technically_ vulnerable to XSS. If someone could get the JSON endpoint to serve a string for `count`, that would be rendered as HTML on the page. Can we sanitize the input with jQuery?

Changed them over to `parseInt` those fields

sethvargo (approved):

PR #1785, opened 2021-02-09
[WIP] codes_claimed by issuer

whaught:
## Proposed Changes

* This feature is experimental and requires privacy review & potentially checks for  a minimum group size to prevent de-anonymized data.
* This PR is just looking at what it takes to add these fields

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -1597,6 +1597,7 @@ func (r *Realm) ExternalIssuerStats(db *Database) (ExternalIssuerStats, error) {
 			$1 AS realm_id,
 			d.issuer_id AS issuer_id,
 			COALESCE(s.codes_issued, 0) AS codes_issued
+			COALESCE(s.codes_claimed, 0) AS codes_claimed
 		FROM (
 			SELECT
 				d.date AS date,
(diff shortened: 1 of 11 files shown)


pkg/database/token.go
```diff
@@ -414,7 +414,46 @@ func (db *Database) updateStatsCodeClaimed(t time.Time, authApp *AuthorizedApp)
...
+				ON CONFLICT (date, realm_id, issuer_id) DO UPDATE
+					SET codes_claimed = external_issuer_stats.codes_claimed + 1`
+
+		if err := db.db.Exec(sql, midnight, vc.RealmID, vc.IssuingExternalID).Error; err != nil {
+			db.logger.Warnw("failed to update external-issuer stats", "error", err)
+		}
+	}
+
+	// Update the per-realm stats.
+	if vc.RealmID != 0 {
```
whaught, line 447: This should already be happening somewhere, but this is what we do for `realm_stats.tokens_claimed` and `realm_stats.codes_issued`

whaught: I grepped for it wrong. we do CodesClaimed++ below

pkg/database/realm.go
```diff
@@ -1597,6 +1597,7 @@ func (r *Realm) ExternalIssuerStats(db *Database) (ExternalIssuerStats, error) {
 			$1 AS realm_id,
 			d.issuer_id AS issuer_id,
 			COALESCE(s.codes_issued, 0) AS codes_issued
```
sethvargo, line 1599: Need comma

sethvargo (commented):
Okay, so it's not _terrible_

pkg/database/realm.go
whaught: which means it also needs a test... looking at that too

PR #1786, opened 2021-02-09
Ensure stats queries are  run in tests

whaught:
## Proposed Changes

* Run the stats queries in test to ensure properly formed SQL

### DIFF
--- a/pkg/database/vercode_test.go
+++ b/pkg/database/vercode_test.go
@@ -451,6 +451,11 @@ func TestStatDates(t *testing.T) {
 			ID: 100,
 		},
 	}
+	app := &AuthorizedApp{
+		Model: gorm.Model{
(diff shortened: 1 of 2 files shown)


pkg/database/vercode_test.go
```diff
@@ -498,6 +508,11 @@ func TestStatDates(t *testing.T) {
 			if f := stats[0].Date.Format(project.RFC3339Date); f != test.statDate {
 				t.Errorf("[%d] expected stat.Date = %s, expected %s", i, f, test.statDate)
 			}
+
+			_, err := user.Stats(db, realm)
```
sethvargo, line 512: Nit, `if _, err := ...; err != nil` keeps the error out of the parent scope.

sethvargo (approved):

PR #1792, opened 2021-02-10
Use a chromedp test for bulk-issue

whaught:
## Proposed Changes

* This matches how our other tests work
* Removes test renderer
* I did this as an experiment to try out `jba/templatecheck`. It's nifty, but ultimately not that useful because it can't assert like chromedp does (and we're already dependant on the chome harness). Long term it's not worth the extra dep


sethvargo (approved):
Can we also remove the interface abstraction in another PR?

whaught:
> Can we also remove the interface abstraction in another PR?

We sure can. Thanks for putting up with the experimenting flip/flop

PR #1801, opened 2021-02-11
Switch linter to GitHub Actions

sethvargo:
This also fixes a number of linter errors that were not being found or not being reported by the previous CI system.

**Release Note**

```release-note
Switch linter CI to GitHub Actions
```

### DIFF
--- a/internal/clients/e2e.go
+++ b/internal/clients/e2e.go
@@ -110,8 +110,8 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 	curDayInterval := timeToInterval(now)
 	nextInterval := curDayInterval
 
-	teks := make([]verifyapi.ExposureKey, 14)
-	for i := 0; i < len(teks); i++ {
+	teks := make([]verifyapi.ExposureKey, 0, 14)
+	for i := 0; i < cap(teks); i++ {
 		key, err := util.RandomExposureKey(nextInterval, maxInterval, 0)
 		if err != nil {
 			return fmt.Errorf("not enough entropy: %w", err)
@@ -268,7 +268,13 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 				return nil, err
 			}
 
-			httpResp, err := client.Post(cfg.KeyServer, "application/json", &b)
+			httpReq, err := http.NewRequestWithContext(ctx, "POST", cfg.KeyServer, &b)
+			if err != nil {
+				return nil, err
+			}
+			httpReq.Header.Set("Content-Type", "application/json")
+
+			httpResp, err := client.Do(httpReq)
 			if err != nil {
 				result = enobs.ResultNotOK
 				return nil, fmt.Errorf("error making request to publish teks: %w", err)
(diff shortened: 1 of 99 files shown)


internal/clients/e2e.go
```diff
@@ -110,7 +110,7 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 	curDayInterval := timeToInterval(now)
 	nextInterval := curDayInterval
 
-	teks := make([]verifyapi.ExposureKey, 14)
+	teks := make([]verifyapi.ExposureKey, 0, 14)
```
mikehelmick, line 113: the 0 here seems wrong since the slice isn't being built up w/ append

sethvargo: It is appended to later though (L310)

mikehelmick: this actually breaks the test though

line 114 would have gone through the loop 14 times

not it is zero since the len(teks) is 0 and not 14 anymore

mikehelmick (changes requested):
for fix of RunEndToEnd

internal/clients/e2e.go
sethvargo: Switch to `cap`.

scripts/presubmit.sh
```diff
@@ -52,60 +52,15 @@ OUT="$(go test -i -tags=performance,e2e ./... 2>&1)" || {
...
-
-
 echo "↹ Verify tabs"
 make tabcheck || {
   echo "✋ Found tabs in html."
   exit 1
 }
 
 
-echo "🌌 Verify and tidy module"
```
sethvargo, line 90: This is handled implicitly now by `readonly` on the linter.

pkg/database/realm.go
```diff
@@ -456,7 +462,14 @@ func (r *Realm) validateSMSTemplate(label, t string) {
 		r.AddError("smsTextTemplate", fmt.Sprintf("when expanded, the result message is too long (%v characters). The max expanded message is %v characters", l, SMSTemplateExpansionMax))
 		r.AddError(label, fmt.Sprintf("when expanded, the result message is too long (%v characters). The max expanded message is %v characters", l, SMSTemplateExpansionMax))
 	}
+}
 
+// enxRedirectDomain returns the configured ENX redirect domain for this realm.
+func (r *Realm) enxRedirectDomain() string {
```
sethvargo, line 468: Small feature addition (package private) to support running more tests in parallel.

pkg/controller/issueapi/issue.go
```diff
@@ -106,7 +106,9 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*api.IssueCodeReq
 			wg.Add(1)
 			go func(request *api.IssueCodeRequest, r *IssueResult) {
 				defer wg.Done()
-				c.SendSMS(ctx, realm, smsProvider, smsSigner, keyID, request, r)
+				if err := c.SendSMS(ctx, realm, smsProvider, smsSigner, keyID, request, r); err != nil {
```
sethvargo, line 109: @whaught double check me here. It looks like we might have been silently discarding the error returned from sending the SMS?

sethvargo (commented):
/unhold

whaught (approved):

pkg/controller/issueapi/issue.go
whaught: Well this is interesting because we set r.ErrorReturn on the inner call. I might double check this with a followup

PR #1803, opened 2021-02-11
Move tabcheck into GitHub Actions

sethvargo:
**Release Note**

```release-note
Move tabcheck into GitHub Actions
```

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -37,7 +37,7 @@ fi
 
 
 echo "📚 Fetch dependencies"
-OUT="$(go get -t -tags=performance,e2e ./... 2>&1)" || {
+OUT="$(go get -t ./... 2>&1)" || {
   echo "✋ Error fetching dependencies"
   echo "\n\n${OUT}\n\n"
   exit 1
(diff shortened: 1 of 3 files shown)


scripts/presubmit.sh
```diff
@@ -37,28 +37,21 @@ fi
...
 echo "📚 Fetch test dependencies"
-OUT="$(go test -i -tags=performance,e2e ./... 2>&1)" || {
+OUT="$(go test -i ./... 2>&1)" || {
   echo "✋ Error fetching test dependencies"
   echo "\n\n${OUT}\n\n"
   exit 1
 }
 
 
-echo "↹ Verify tabs"
```
mikehelmick, line 55: should we preserve all of these checks as client side executable (can be different script)?

mikehelmick (commented):
/lgtm

but see comment for possible followups

scripts/presubmit.sh
sethvargo: They already are 😄 

The CI instruction is `make tabcheck` which still exists and can be run locally. Same thing with `make lint`. It's a personal pet peeve of mine to have to rely on a CI system to get feedback.

PR #1804, opened 2021-02-11
Handle sms errors in the function

whaught:
## Proposed Changes

* Seth found that this function attempts to [both handle an error and return it](https://github.com/google/exposure-notifications-verification-server/pull/1801#discussion_r574837896) which is ambiguous as to what the caller should do.
    * The new code might swallow the `api.ErrSMSQueueFull` errorReturn and needs the `.WithCode(api.ErrSMSFailure)` to give the proper enum error code
    * Instead we fully handle the errors inside sendSMS and don't return the error.
    * Move the anonymous function to a fully named `doSend` for readability

### DIFF
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -37,6 +37,7 @@ type IssueResult struct {
 	obsResult   tag.Mutator
 }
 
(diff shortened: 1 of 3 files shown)


sethvargo (approved):
/lgtm

sethvargo (approved):

PR #1809, opened 2021-02-12
Run go tests on GitHub Actions

sethvargo:
/hold

Gonna be WIP for awhile.

**Release Note**

```release-note
Run go tests on GitHub Actions
```


sethvargo:
@mikehelmick @whaught this is ready for review

pkg/controller/user/update_test.go
```diff
@@ -177,7 +177,7 @@ func TestHandleUpdate(t *testing.T) {
 		t.Parallel()
 
 		browserCtx := browser.New(t)
-		taskCtx, done := context.WithTimeout(browserCtx, 30*time.Second)
+		taskCtx, done := context.WithTimeout(browserCtx, 120*time.Second)
```
mikehelmick, line 180: feels like all of these should be a var somewhere so next time we can change it once...

don't have to be in this PR

mikehelmick (approved):

Issue #1777, opened 2021-02-05
Improve cloud scheduler job alerts.


mariliamelo:
### TL;DR
We should have alerts that consistently tracks the forward progress of our cloud scheduler jobs.

### Design

We want to make sure the system alerts if the following jobs aren't being executed according to the expectation.

E.g:
JWKS - importing at least once ever X minutes
database backup - successfully runs at least once every 12 hours

Mike to define the requirements for the other jobs.

mariliamelo:
Here is the list of cloud scheduler jobs we have:

Job | Frequency | When to be alerted | page Y/N |
-----|-----|-----|-----
appsync-worker | Once a day | TBD | TBD |
backup-database-worker | Every 6 hours |  TBD | TBD |
cleanup-worker |  Every hour | TBD | TBD |
docker-mirror-worker | Once a day | TBD | TBD | 
e2e-default-workflow | Every 5-10 mins | TBD | Yes | 
e2e-revise-workflow | Every 5 - 10 mins | TBD | Yes |
e2e-enx-redirect-workflow | Every 5-10 mins|  TBD | Yes
modeler-worker | Once a day | TBD | TBD |
realm-key-rotation-worker | Every 30 mins | TBD | TBD |
rotation-worker | Every 5 mins | TBD | TBD |
stats-puller-worker | Every 30 mins | TBD | TBD |

Every job will be retried 3 times. 

I'm separating the to be alerted and page columns in case we want to file a ticket instead of paging someone for a given job failure. We can customize to only alert when it is needed to page as well.

mikehelmick:
my quick suggestions

* appsync-worker - run 4x a day. Alert if it fails twice in a row.
*  backup-database-worker - alert on all failures, page, playbook should have us run a manual backup
* cleanup worker - alert if no success for 6 consecutive hours in a row 
* docker-mirror-worker - email on failures, do not page.
* e2e - page on each failure, but not if the scheduler fails to reach thyte container
* modeler - run 4x a day, alert if it fails twice ina. row
* relam-key-rotation-worker - change to every 15 minutes, alert if fails twice in a row
* rotation worker - change to every 30 minutes, alert if fails twice in a row
* stats-puller - change to 3x an hour - 10,20,30 after the hour. Alert if fails 3x in a row

sethvargo:
In general, I think "fails twice in a row" is going to be harder to track than "didn't succeed in X minutes".

> docker-mirror-worker - email on failures, do not page.

[...]

PR #1811, opened 2021-02-12
Update retry_count and database backup to synchronous

mariliamelo:
Part of #1777

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -406,7 +406,7 @@ resource "google_cloud_scheduler_job" "backup-database-worker" {
   attempt_deadline = "1800s"
 
   retry_config {
-    retry_count = 1
+    retry_count = 3
   }
 
   http_target {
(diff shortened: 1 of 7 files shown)


terraform/database.tf
```diff
@@ -418,7 +418,7 @@ resource "google_cloud_scheduler_job" "backup-database-worker" {
         fileType  = "SQL"
         uri       = "gs://${google_storage_bucket.backups.name}/database/${google_sql_database.db.name}",
         databases = [google_sql_database.db.name]
-        offload   = true,
+        offload   = false,
```
sethvargo, line 421: I'm not sure this is a good idea unless we _really_ bump the timeout. It can take 15+min

mariliamelo: The timeout now is 30 mins. This is the best way to have the retry logic work without relying on external services. (Cloud SQL or GCS) to verify the backup.

mariliamelo: Also, 30mins is the max attempt_deadline allowed: https://cloud.google.com/scheduler/docs/reference/rpc/google.cloud.scheduler.v1#:~:text=The%20failed%20attempt%20can%20be,15%20seconds%20and%2024%20hours.

sethvargo (approved):

PR #1813, opened 2021-02-13
Some simple cleanups to the codebase

vagababov:
This is just a subset, since everything in one pass would be a big PR.

- Use standard constants for HTTP things.
- Collapse some test cases for shorter files
- Some code reflow

/assign @whaught

### DIFF
--- a/pkg/controller/form.go
+++ b/pkg/controller/form.go
@@ -34,8 +34,5 @@ func BindForm(w http.ResponseWriter, r *http.Request, data interface{}) error {
 	// don't need to be captured. By default schema decoder is very struct.
 	decoder.IgnoreUnknownKeys(true)
 
-	if err := decoder.Decode(data, r.PostForm); err != nil {
-		return err
-	}
-	return nil
+	return decoder.Decode(data, r.PostForm)
 }
--- a/pkg/controller/middleware/email_verified_test.go
+++ b/pkg/controller/middleware/email_verified_test.go
@@ -169,7 +169,7 @@ func TestRequireEmailVerified(t *testing.T) {
 			w.Flush()
 
 			if got, want := w.Code, tc.code; got != want {
-				t.Errorf("Status = %d, want: %d", got, want)
+				t.Errorf("Expected %d to be %d", got, want)
 			}
 		})
 	}
--- a/pkg/controller/middleware/email_verified_test.go
+++ b/pkg/controller/middleware/email_verified_test.go
@@ -45,7 +45,7 @@ func TestRequireEmailVerified(t *testing.T) {
 	requireEmailVerified := middleware.RequireEmailVerified(authProvider, h)
 
 	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(200)
+		w.WriteHeader(http.StatusOK)
 	})
 
 	cases := []struct {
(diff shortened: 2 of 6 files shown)


pkg/controller/form.go
```diff
@@ -34,8 +34,5 @@ func BindForm(w http.ResponseWriter, r *http.Request, data interface{}) error {
 	// don't need to be captured. By default schema decoder is very struct.
 	decoder.IgnoreUnknownKeys(true)
 
-	if err := decoder.Decode(data, r.PostForm); err != nil {
-		return err
```
sethvargo, line 38: We should actually probably wrap this error in some meaningful context

pkg/controller/middleware/email_verified_test.go
```diff
@@ -54,87 +54,79 @@ func TestRequireEmailVerified(t *testing.T) {
...
-			code:       400,
-		},
-		{
-			name:          "optional_verified",
-			emailVerified: true,
-			membership: &database.Membership{
-				Realm: &database.Realm{
-					EmailVerifiedMode: database.MFAOptional,
-				},
+	}{{
```
sethvargo, line 57: The previous format is better for avoiding unnecessary diffs

pkg/controller/middleware/email_verified_test.go
```diff
@@ -169,7 +160,7 @@ func TestRequireEmailVerified(t *testing.T) {
 			w.Flush()
 
 			if got, want := w.Code, tc.code; got != want {
-				t.Errorf("expected %d to be %d", got, want)
```
sethvargo, line 172: We use this format for test failures across the project and I'd prefer to keep it like this. Is there a strong reason to change it?

pkg/controller/middleware/firewall_test.go
```diff
@@ -44,91 +45,79 @@ func TestProcessFirewall(t *testing.T) {
...
-			name: "single_reject_xff",
-			ctx: controller.WithRealm(ctx, &database.Realm{
-				AllowedCIDRsServer: []string{"1.2.3.4/32"},
-			}),
-			remoteAddr: "1.2.3.4",          // xff is preferred over remote ip
-			xff:        "5.6.7.8, 1.2.3.4", // Only trusts the first value in xff
-			code:       401,
-		},
-	}
+	}{{
```
sethvargo, line 48: Please revert to old style

pkg/controller/middleware/mfa_test.go
```diff
@@ -50,113 +51,103 @@ func TestRequireMFA(t *testing.T) {
...
-			code:       400,
-		},
-		{
-			name:        "optional_verified",
-			mfaVerified: true,
-			membership: &database.Membership{
-				Realm: &database.Realm{
-					MFAMode: database.MFAOptional,
-				},
+	}{{
```
sethvargo, line 54: Please revert to old style

sethvargo (changes requested):

pkg/controller/middleware/email_verified_test.go
mikehelmick: agree on this - this style of literal slice is all over the codebases, let's stay consistent

pkg/controller/form.go
vagababov: well probably not part of this pr :)

pkg/controller/middleware/email_verified_test.go
vagababov: Sure...

pkg/controller/middleware/email_verified_test.go
vagababov: The more or less google standard go is "got before want".  Besides this particular message (even if keeping the order the same) "expected 200 to be 400" is not the best message.

sethvargo: It's still "got before want" in both the init and print. 

We use this same error output for most HTTP status tests throughout the entire project, and I think consistency is more important. 

While we also aren't bound by Google's internal recommendations, I don't think this is in violation.

sethvargo (changes requested):
One remaining item

pkg/controller/form.go
sethvargo: Then leave it as-is and do it as a follow-up?

pkg/controller/middleware/email_verified_test.go
vagababov: Kind of, but it reads weird. Anyway. While not bound it's still better to follow the general style.
🤷 — I was going to fix all of those one by one. Note that unit test errors are also expected to be uppercased in general — since they are whole sentences.

vagababov:
OK pushed the old error, but still uppercased it (I think that's the right thing).
I'll work on a sed to uppercase everything else same way separately, if you don't mind.

pkg/controller/form.go
vagababov: Not sure why? The code is idempotent as is, but 4 times more readable (4:1 line ratio )

sethvargo:
Hey @vagababov 

It looks like most of this was wrapped up in the other PR that we merged. Do you still want to keep this one open?

vagababov:
Let's close it. I'll send more more targeted prs.

PR #1815, opened 2021-02-14
Use standard Golang net/http provided constant for HTTP methods

vagababov:
No reason not to use standard ones.


mikehelmick (approved):
/ok-to-test
/lgtm

Issue #1819, opened 2021-02-15
Move test timeout to a project variable or envvar


sethvargo:
https://github.com/google/exposure-notifications-verification-server/pull/1809#discussion_r575574878

/assign

PR #1826, opened 2021-02-15
Make per-test timeout a project function

sethvargo:
**Release Note**

```release-note
Make per-test timeout a project function
```


sethvargo:
Fixes #1819

mikehelmick (approved):

PR #1831, opened 2021-02-16
admin API needs signing config to get key manager settings

mikehelmick:
## Proposed Changes

* add signing config to admin api
* 
**Release Note**

```release-note
Admin API gets signing config in terraform
```


mikehelmick:
we'll need to cut another point release after this

sethvargo (approved):

sethvargo:
> we'll need to cut another point release after this

Given the functionality is off by default, I think we can wait a few days in case other items pop up.

Issue #1830, opened 2021-02-16
authenticated SMS - fail open


mikehelmick:
### TL;DR

When authenticated SMS is enabled, but the crypto key is unavailable for whatever reason, it would be best to fail open and send the message without the signature appended.

PR #1834, opened 2021-02-16
Introduce configurable Authenticated SMS failure modes

sethvargo:
This introduces a runtime configurable to control how the system behaves when signing an authenticated SMS fails. The default behavior is to "fail open" (continue on error), meaning if the signing fails for any reason, the system still continues and delivers the SMS. Operators can also configure the system to "fail closed" (fail on error) to halt and NOT send SMS unless signing succeeds. To control the behavior, operators must set the `SMS_FAIL_CLOSED` environment variable on the `server` and `adminapi` services. Set the value to `true` to fail closed. The default value is false (fail open).

Regardless of the configuration, an error is logged and a metric is incremented when an SMS fails to sign. There is also a new alerting policy for when more than 5 SMS messages fail to sign in a 60s window by realm. This ensures that, even if the system is configured to fail open, operators are aware of the issue and can respond accordingly. The alert is non-paging since there is rarely action an operator can take if the upstream key management system is unavailable.

The double negative is weird (fail closed false), but that is to ensure the default behavior is false even when envconfig is not used (since Go's default value for bool is false)

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1830

**Release Note**

```release-note
Introduce configurable Authenticated SMS failure modes. The default behavior is to "fail open" (continue on error). Operators can configure the system to "fail closed" (halt on error) by setting the `SMS_FAIL_CLOSED` environment variable on the `server` and `adminapi` services. We recommend leaving the default configuration. Regardless of the configuration, this also introduces a new non-paging alert to inform operators when an out-of-threshold number of failures occur while signing SMS messages, per realm. See the new playbook for more information.
```

### DIFF
--- a/pkg/config/admin_server_config.go
+++ b/pkg/config/admin_server_config.go
@@ -120,3 +120,7 @@ func (c *AdminAPIServerConfig) ObservabilityExporterConfig() *observability.Conf
 func (c *AdminAPIServerConfig) IsMaintenanceMode() bool {
 	return c.MaintenanceMode
 }
+
+func (c *AdminAPIServerConfig) GetAuthenticatedSMSFailClosed() bool {
+	return c.SMSSigning.FailClosed
+}
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -248,6 +248,38 @@ resource "google_monitoring_alert_policy" "UpstreamUserRecreates" {
   ]
 }
 
+resource "google_monitoring_alert_policy" "AuthenticatedSMSFailure" {
+  project      = var.project
+  combiner     = "OR"
+  display_name = "AuthenticatedSMSFailure"
+  conditions {
+    display_name = "Failed attempts to sign an SMS for authenticated SMS"
+    condition_monitoring_query_language {
+      duration = "60s"
+      query    = <<-EOT
+      fetch
+      generic_task :: custom.googleapis.com/opencensus/en-verification-server/api/issue/authenticated_sms_failure_count
+      | align rate(5m)
+      | every 1m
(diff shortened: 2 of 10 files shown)


terraform/alerting/alerts.tf
```diff
@@ -248,6 +248,38 @@ resource "google_monitoring_alert_policy" "UpstreamUserRecreates" {
...
 }
 
+resource "google_monitoring_alert_policy" "AuthenticatedSMSFailure" {
+  project      = var.project
+  combiner     = "OR"
+  display_name = "AuthenticatedSMSFailure"
+  conditions {
+    display_name = "Failed attempts to sign an SMS for authenticated SMS"
+    condition_monitoring_query_language {
+      duration = "60s"
```
sethvargo, line 258: @mariliamelo @yegle can yall check my MQL here please? It's non-paging, so I don't think we need to get it 100% accurate on the first go (can tune over time), but I'd like to be at least _close_ 😄

pkg/config/admin_server_config.go
```diff
@@ -120,3 +120,7 @@ func (c *AdminAPIServerConfig) ObservabilityExporterConfig() *observability.Conf
 func (c *AdminAPIServerConfig) IsMaintenanceMode() bool {
 	return c.MaintenanceMode
 }
+
+func (c *AdminAPIServerConfig) GetAuthenticatedSMSFailClosed() bool {
```
mikehelmick, line 124: just `AuthenticatedSMSFailClosed` ?

sethvargo: I was trying to match the existing things in the interface

mikehelmick: yeah - that's fine.

Issue #1832, opened 2021-02-16
Enable authenticated SMS in E2E runner


sethvargo:

mikehelmick:
SMS in general...

PR #1839, opened 2021-02-17
Send authenticated SMS in e2e-runner

sethvargo:
This updates the e2e-runner to send an SMS credential as part of its test. This feature is on by default, but it can be disabled by setting `E2E_SKIP_SMS` to any truthy value on the e2e-runner service.

Operators must configure the e2e realm with [Twilio Test Credentials](https://www.twilio.com/docs/iam/test-credentials) via the verification server UI. Note that the system will not actually send an SMS, it uses Twilio's magic numbers and test credentials to simulate all aspects of the system and API calls but stops short of actually delivering an SMS.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1832

**Release Note**

```release-note
Simulate sending SMS in e2e-runner and enable authenticated SMS for the e2e test realm. Operators will either need to opt out of the e2e-runner service testing SMS paths by setting `E2E_SKIP_SMS` (not recommended) or configure the e2e test realm with [Twilio Test Credentials](https://www.twilio.com/docs/iam/test-credentials). See the updated production guide for screenshots and examples. We recommend configuring the e2e-test-realm _before_ upgrading the service to minimize alerts.
```


cmd/e2e-runner/main.go
```diff
@@ -109,7 +111,18 @@ func realMain(ctx context.Context) error {
 		}
 	}()
 
-	// Create the enx-redirect client if the URL was specified.
+	// Verify that SMS is configured on the realm
+	if !project.SkipE2ESMS {
+		has, err := resp.Realm.HasSMSConfig(db)
+		if err != nil {
+			return fmt.Errorf("failed to check if realm has sms config: %w", err)
```
mikehelmick, line 118: I think this is ok - we'll still alert that the request failed.

the request will fail because the container fails to start.

internal/clients/e2e.go
```diff
@@ -145,6 +145,11 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 				ExternalIssuerID: adminID,
 			}
 
+			// Add a phone number unless SMS is skipped.
+			if !project.SkipE2ESMS {
+				codeReq.Phone = "+18558361987"
```
mikehelmick, line 150: we should comment on the significance of this number, maybe put it in a const

internal/envstest/bootstrap.go
```diff
@@ -67,11 +72,27 @@ func Bootstrap(db *database.Database) (*BootstrapResponse, error) {
...
 	realm.AllowedTestTypes = database.TestTypeNegative | database.TestTypeConfirmed | database.TestTypeLikely
 	realm.AllowBulkUpload = true
+	realm.UseAuthenticatedSMS = true
 	if err := db.SaveRealm(realm, database.SystemTest); err != nil {
 		return &resp, fmt.Errorf("failed to save realm: %w", err)
 	}
 	resp.Realm = realm
 
+	// Ensure an SMS signing key exists.
+	if _, err := realm.CurrentSMSSigningKey(db); err != nil {
```
mikehelmick, line 82: does the e2e runner have key admin capabilities?

sethvargo: Good call. We could alternatively remove this and move it into the admin setup guide with the existing manual steps. What do you think? I'm kinda worried about giving e2e permission to manage keys.

mikehelmick: my vote is move it to the docs

internal/clients/e2e.go
sethvargo: Done

internal/envstest/bootstrap.go
sethvargo: Done

mikehelmick (dismissed):

mikehelmick (approved):

PR #1844, opened 2021-02-17
Improve cloud scheduler job alerts

mariliamelo:
Part of #1777. This will temporarily improve the alerts before we break it in different alerts with different thresholds.


sethvargo (approved):
/lgtm

PR #1850, opened 2021-02-22
Make e2e-runner publish a real client

sethvargo:
This gives us all the JSON parsing + automatic fallback and nice error messages when something fails. This commit changes the required configuration for `KEY_SERVER` from the full URL to the /publish endpoint to *just* the URL to the key server. Where previously you may have configured `KEY_SERVER=https://foo.bar/v1/publish`, please re-configure with `KEY_SERVER=https://foo.bar`. The system attempts to maintain backwards compatibility by parsing the URL, but this may be removed at a later date.

**Release Note**

```release-note
Upgrade e2e-runner's HTTP client. This changes the required configuration for `KEY_SERVER` from the full URL to the `/publish` endpoint to _just_ the URL to the key server. Where previously you may have configured `KEY_SERVER=https://foo.bar/v1/publish`, please re-configure with `KEY_SERVER=https://foo.bar`. The system attempts to maintain backwards compatibility by parsing the URL, but this may be removed at a later date.
```

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -192,7 +192,7 @@ func handleEndToEnd(cfg *config.E2ERunnerConfig, h *render.Renderer) http.Handle
 		ctx := r.Context()
 
 		if err := clients.RunEndToEnd(ctx, cfg); err != nil {
-			renderJSONError(w, r, h, err)
+			renderJSONError(w, r, h, fmt.Errorf("failed to run end-to-end: %w", err))
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


sethvargo:
The goal here is to remove all instances of:

```text
invalid character '<' looking for beginning of value
```

which sporadically occur across environments. Our best guess is that the server is sometimes sending an HTML response when it should be sending JSON, however, we don't know the actual response body. We have logic for handling this edge case in the clients package, but the e2e runner was not using the clients package to interact with the key server.

mikehelmick (approved):

PR #1852, opened 2021-02-23
Refactor user tests to use less chromedp

sethvargo:
This refactors the controller/user package tests to use less chromedp. We relied on chromedp a lot for testing rendering, but it's really expensive and should only be used to test really complex CUJs which rely on the UI. I profiled the I/O and found a few other optimizations we can make, including:

- Introduce `BadDatabase` onto the test harness. Previously tests where we rely on injecting the faulty database would spin up a real database, then replace it with the injected one. The harness now exports a good database and `BadDatabase`, which is the same fault-injected one.

- Introduce some test helpers for building form and json requests. I identified a ton of repetition (in controller/user, but also every other test package) for building a request with context + url.Values + content-type headers. That is now centralized in a helper. Only used in controller/user for now to keep the diff small.

With these changes, the controller/user tests went from 14.8s -> 6.2s on my laptop, and they consume 480% less CPU.

**Release Note**

```release-note
Refactor user tests to use less chromedp
```

/assign @mikehelmick


pkg/controller/user/create_test.go
```diff
@@ -151,43 +106,50 @@ func TestHandleCreate(t *testing.T) {
...
+			"name":  []string{"Test User"},
+			"email": []string{"you@example.com"},
+		})
+		handler.ServeHTTP(w, r)
 
-			chromedp.WaitVisible(`body#apikeys-show`, chromedp.ByQuery),
-			chromedp.Value(`#apikey-value`, &apiKey, chromedp.ByQuery),
-		); err != nil {
-			t.Fatal(err)
+		if got, want := w.Code, 303; got != want {
```
mikehelmick, line 126: http.StatusSeeOther

pkg/controller/user/delete_test.go
```diff
@@ -125,40 +97,32 @@ func TestHandleDelete(t *testing.T) {
...
+		w, r := envstest.BuildFormRequest(ctx, t, http.MethodDelete, "/", nil)
+		r = mux.SetURLVars(r, map[string]string{
+			"id": fmt.Sprintf("%v", user.ID),
+		})
+		handler.ServeHTTP(w, r)
 
-			chromedp.WaitVisible(`body#users-index`, chromedp.ByQuery),
-		); err != nil {
-			t.Fatal(err)
+		if got, want := w.Code, 303; got != want {
```
mikehelmick, line 117: StatusSeeOther

pkg/controller/user/import_test.go
```diff
@@ -15,40 +15,54 @@
...
+		ctx = controller.WithMembership(ctx, &database.Membership{
+			Realm:       realm,
+			User:        testUser,
+			Permissions: rbac.UserWrite,
+		})
+
+		w, r := envstest.BuildFormRequest(ctx, t, http.MethodGet, "/", nil)
+		handler.ServeHTTP(w, r)
+
+		if got, want := w.Code, 200; got != want {
```
mikehelmick, line 64: StatusOK

pkg/controller/user/index_test.go
```diff
@@ -108,20 +80,41 @@ func TestHandleIndex(t *testing.T) {
...
+			User:        testUser,
+			Permissions: rbac.UserRead,
+		})
+
+		w, r := envstest.BuildFormRequest(ctx, t, http.MethodGet, "/", &url.Values{
+			"q": []string{"@example.com"},
+		})
+		handler.ServeHTTP(w, r)
+
+		if got, want := w.Code, 200; got != want {
```
mikehelmick, line 98: StatusOK

pkg/controller/user/index_test.go
```diff
@@ -108,20 +80,41 @@ func TestHandleIndex(t *testing.T) {
...
+		ctx = controller.WithMembership(ctx, &database.Membership{
+			Realm:       realm,
+			User:        testUser,
+			Permissions: rbac.UserRead,
+		})
+
+		w, r := envstest.BuildFormRequest(ctx, t, http.MethodGet, "/", nil)
+		handler.ServeHTTP(w, r)
+
+		if got, want := w.Code, 200; got != want {
```
mikehelmick, line 116: StatusOK

pkg/controller/user/update_test.go
```diff
@@ -173,81 +124,48 @@ func TestHandleUpdate(t *testing.T) {
...
+		})
+		r = mux.SetURLVars(r, map[string]string{
+			"id": fmt.Sprintf("%d", testUser.ID),
+		})
+		handler.ServeHTTP(w, r)
 
-			if err := chromedp.Run(taskCtx, actions...); err != nil {
-				t.Fatal(err)
-			}
+		if got, want := w.Code, 303; got != want {
```
mikehelmick, line 150: StatusSeeOther

PR #1853, opened 2021-02-23
Only test verification server with google tags

sethvargo:
**Release Note**

```release-note
NONE
```


mikehelmick (approved):
this is fine here - we can't do this on the key server

PR #1866, opened 2021-02-25
Drop support for crypto/dsa in public key parsing

sethvargo:
The system only uses ecdsa keys, so this should not affect anyone.

**Release Note**

```release-note
Drop support for crypto/dsa in public key parsing. The EN system only supports ecdsa keys, so this will not affect the system.
```


sethvargo:
It produces a warning on Go 1.16

mikehelmick (approved):

PR #1872, opened 2021-02-26
Remove deprecated TOKEN_SIGNING_KEYS

sethvargo:
⚠️ 🛑 Please review with caution 🛑 ⚠️ 

The new value is TOKEN_SIGNING_KEY and it must point to the parent key. Token signing keys have moved into the database since v0.22.

You MUST upgrade to v0.22 for at least 24 hours before applying this release.

**Release Note**

```release-note
Remove backwards-compatibility for `TOKEN_SIGNING_KEY`. Specifying multiple values has been deprecated since 0.21 and token signing keys have moved into the database. If you are on an older version, you **MUST upgrade to v0.22 for at least 24h before applying this update.** You should ensure that the value of `TOKEN_SIGNING_KEY` points to a key (not a key version) in the service environment before applying this upgrade. Since v0.22, `TOKEN_SIGNING_KEY` accepted a key version or a parent key. This release only accepts a parent key.
```

### DIFF
--- a/pkg/config/token_signing_config.go
+++ b/pkg/config/token_signing_config.go
@@ -44,6 +44,9 @@ func (t *TokenSigningConfig) Validate() error {
 	if strings.Contains(t.TokenSigningKey, ",") {
 		return fmt.Errorf("TOKEN_SIGNING_KEY can only contain one element")
 	}
+	if strings.Contains(t.TokenSigningKey, "/cryptoKeyVersions") {
+		return fmt.Errorf("TOKEN_SIGNING_KEY must be the path to a parent crypto key (not the version)")
+	}
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


pkg/config/token_signing_config.go
```diff
@@ -28,76 +28,21 @@ type TokenSigningConfig struct {
...
 }
 
-// ParentKeyName returns the name of the parent key.
-func (t *TokenSigningConfig) ParentKeyName() string {
-	// Validation prevents this slice from being empty.
-	value := t.TokenSigningKeys[0]
-
-	// This is Google-specific, but that's the only platform where we can
-	// meaningfully detect this.
-	if idx := strings.Index(t.TokenSigningKeys[0], "/cryptoKeyVersions"); idx != -1 {
```
sethvargo, line 60: Should we move this to validation and fail if we detect that substring now?

mikehelmick: yes - I think so

sethvargo: Done

mikehelmick (approved):

PR #1878, opened 2021-02-26
Do not use embedded fs in dev mode

sethvargo:
This allows reloading asset files

**Release Note**

```release-note
Do not use embedded fs in dev mode
```


sethvargo:
⚠️ This does make it impossible to set `DEV_MODE` in production. I think that's probably good and by design.

mikehelmick (approved):

PR #1881, opened 2021-02-28
Upgrade to Firebase auth 8.2.9

sethvargo:
[Changes to auth since 8.2.1](https://firebase.google.com/support/release-notes/js):

- API update for an internal package. There is no change in the public API.
- Fixed a bug where UTF-8 encoded strings were not decoded in getIdTokenResult. See https://github.com/firebase/firebase-js-sdk/issues/4174.
- Added the useEmulator() function and emulatorConfig to the firebase package externs.

**Release Note**

```release-note
Upgrade to Firebase Auth version 8.2.9.
```

### DIFF
--- a/assets/server/firebase.html
+++ b/assets/server/firebase.html
@@ -1,7 +1,7 @@
 {{define "firebase"}}
 {{- if .firebase}}
-<script src="https://www.gstatic.com/firebasejs/8.2.1/firebase-app.js"></script>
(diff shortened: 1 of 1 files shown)


sethvargo:
FWIW, GitHub Actions and Prow are now 1:1 with test duration.

mikehelmick (approved):

PR #1886, opened 2021-02-28
change chart combos

mikehelmick:
## Proposed Changes

Update stats visualizations based on user input

* remove tokens claimed/tokens invalid - that chart has been confusing
* add tokens claimed and publish requests to make code chart (top of page) -- IF key server metrics are enabled
* separate out revisions and missing onset from TEKs published (different units)
    * don't show revisions at all (to our knowledge, nobody is using that functionality)
* add total publish requests and missing onset as overlays on the publish bar chart

image

image

**Release Note**

```release-note
* The tokens claimed/invalid chart has been removed (data is still in exports)
* Add total publish requests to the codes/issued claim chart (if key server stats are enabled)
* Separate out revisions and missing onset from TEKs published
* Add total publish requests and missing onset as overlays on the publish bar chart
```

### DIFF
--- a/assets/server/realmadmin/_stats_codes.html
+++ b/assets/server/realmadmin/_stats_codes.html
@@ -3,7 +3,7 @@
 <div class="card shadow-sm mb-3">
   <div class="card-header">
     <span class="oi oi-bar-chart mr-2 ml-n1"></span>
-    Codes issued, claimed, &amp; invalid
+    Codes Issued &amp; Usage
   </div>
   <div id="dashboard_div">
     <div id="realm_chart_div" class="h-100 w-100" style="min-height:325px;">
@@ -17,6 +17,12 @@
       <span class="mr-1">Export as:</span>
       <a href="/stats/realm.csv" class="mr-1">CSV</a>
       <a href="/stats/realm.json" target="_blank">JSON</a>
+      {{if .hasKeyServerStats}}
+      (legacy code data) |
+      <span class="mr-1">Composite Data:</span>
+      <a href="/stats/composite.csv" class="mr-1">CSV</a>
+      <a href="/stats/composite.json" target="_blank">JSON</a>  
+      {{end}}
     </span>
   </small>
 </div>
@@ -68,41 +74,57 @@
   <div class="modal-dialog modal-dialog-centered">
     <div class="modal-content">
       <div class="modal-header">
-        <h5 class="modal-title">Codes issued, claimed, &amp; invalid</h5>
+        <h5 class="modal-title">Codes Issued &amp; Usage</h5>
         <button type="button" class="close" data-dismiss="modal" aria-label="Close">
           <span aria-hidden="true">&times;</span>
         </button>
       </div>
       <div class="modal-body mb-n3">
         <p>
-          This graph reflects the total number of codes issued and claimed
-          for this realm, grouped by UTC day.
+          This graph reflects the total number of codes issued and usage
+          of those codes for this realm, grouped by UTC day.
         </p>
 
-        <strong>Issued</strong>
+        <strong>Codes Issued</strong>
         <p>
           This line tracks the total number of codes issued. Codes can be
           issued via the web interface or via the API. Both types of codes
           are included in this count.
         </p>
 
-        <strong>Claimed</strong>
+        <strong>Codes Claimed</strong>
         <p>
           This line tracks the total number of codes claimed. Codes can only
           be claimed by the end user using the API. Typically the iOS or
           Android application is responsible for claiming the code.
         </p>
 
-        <strong>Invalid</strong>
+        <strong>Invalid Codes</strong>
         <p>
           This line tracks the total number of codes that were rejected by the
           system. This includes codes that were incorrectly entered
           (typographical error) and codes that have expired.
         </p>
 
+       
+        <strong>Tokens Claimed</strong>
+        <p>
+          This line tracks the total number of users that successfully
+          claimed a verification code and then consented to key release.
+        </p>
+
+        {{if .hasKeyServerStats}}
+        <strong>Publish Requests</strong>
+        <p>
+          The number of users that followed the flow all the way through
+          and successfully completed uploading of temporary exposure keys.          
+        </p>
+        {{end}}
+
         <hr>
 
         <strong>Inferring this data</strong>
+        <p><em>All statistics are captured on a best effort basis.</em></p>
         <ul>
           <li>
             <em>Issued codes</em> does not necessarily correspond to the number
@@ -119,6 +141,17 @@
             The delta between <em>codes issued</em> and <em>codes claimed</em>
             can be used as a rudimentary measure of adoption.
           </li>
+          <li>
+            The delta between codes claimed and tokens claimed indicates users
+            that enter a valid verification code but don't get through the consent
+            to share data screen.
+          </li>
+          {{if .hasKeyServerStats}}
+          <li>
+            The delta between tokens claimed and publish requests indicates that
+            a user got past the consent screen but keys were not uploaded.
+          </li>
+          {{end}}
         </ul>
       </div>
     </div>
@@ -173,10 +206,12 @@
 
   function drawRealmCharts() {
     $.ajax({
-      url: '/stats/realm.json',
+      url: '/stats/composite.json',
       dataType: 'json',
     })
     .done(function(data, status, xhr) {
+      console.log(data);
+
       if (data.statistics) {
         $issueAgeSlider.attr('min', Math.min(smoothing, data.statistics.length));
         $issueAgeSlider.attr('max', data.statistics.length);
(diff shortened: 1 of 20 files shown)


assets/server/realmadmin/_stats_codes.html
```diff
@@ -173,10 +206,12 @@ <h5 class="modal-title">Mean time until code claimed</h5>
 
   function drawRealmCharts() {
     $.ajax({
-      url: '/stats/realm.json',
+      url: '/stats/composite.json',
       dataType: 'json',
     })
     .done(function(data, status, xhr) {
+      console.log(data);
```
sethvargo, line 213: Debug

assets/server/realmadmin/_stats_codes.html
```diff
@@ -212,12 +247,20 @@ <h5 class="modal-title">Mean time until code claimed</h5>
...
     let dataTable = new google.visualization.DataTable();
     dataTable.addColumn('date', 'Date');
-    dataTable.addColumn('number', 'Issued');
-    dataTable.addColumn('number', 'Claimed');
-    dataTable.addColumn('number', 'Invalid');
+    dataTable.addColumn('number', 'Codes Issued');
+    dataTable.addColumn('number', 'Codes Claimed');
+    dataTable.addColumn('number', 'Invalid Codes');
+    dataTable.addColumn('number', 'Tokens Claimed');
+    {{if .hasKeyServerStats}}
```
sethvargo, line 254: Can we do this based on the JavaScript response instead?

pkg/controller/stats/composite.go
```diff
@@ -0,0 +1,300 @@
...
+	Stats   []*jsonCompositeStatStats `json:"statistics"`
+}
+
+type jsonCompositeStatStats struct {
+	Date time.Time                   `json:"date"`
+	Data *jsonCompositeStatStatsData `json:"data"`
+}
+
+type jsonCompositeStatStatsData struct {
+	// Fields that come from the realm stats
```
sethvargo, line 62: Why not just embed both of those stats structs?

assets/server/realmadmin/_stats_codes.html
mikehelmick: if we did, it would involve scanning every row and seeing if key server stats exist.
I think this is cleaner.

pkg/controller/stats/composite.go
mikehelmick: I copied the structure here: https://github.com/google/exposure-notifications-verification-server/blob/main/pkg/database/realm_stats.go#L125

so that I could minimize javascript changes by having the exact same json structure

sethvargo: If you move this into the database package though, couldn't you do the following?

```go
type jsonCompositeStatStatsData struct {
  jsonRealmStatStats
  // ...
}
```

I'm worried about things getting out of sync.

assets/server/realmadmin/_stats_codes.html
sethvargo: Javascript is cheap, over a very small data set (N is usually < 100), and computed by the client instead of the server. Relying on information from the template makes it more difficult to move this into the application bundle in the future.

I also don't think you need to scan every row. I think you just need to scan the _last_ row. If there are key server stats in the last row, it means stats are enabled. If there aren't stats in the last row, it means that key server stats aren't enabled, _or_ they have recently been disabled. In either case, I think you arrive at the desired behavior. Am I missing something?

assets/server/realmadmin/_stats_codes.html
```diff
@@ -17,6 +17,12 @@
       <span class="mr-1">Export as:</span>
       <a href="/stats/realm.csv" class="mr-1">CSV</a>
       <a href="/stats/realm.json" target="_blank">JSON</a>
+      {{if .hasKeyServerStats}}
+      (legacy code data) |
+      <span class="mr-1">Composite Data:</span>
```
sethvargo, line 22: Since composite _includes_ realm stats, I think this should be an if/else instead of showing both.

internal/routes/server.go
```diff
@@ -438,6 +438,9 @@ func statsRoutes(r *mux.Router, c *stats.Controller) {
 
 	r.Handle("/realm/key-server.csv", c.HandleKeyServerStats(stats.TypeCSV)).Methods(http.MethodGet)
 	r.Handle("/realm/key-server.json", c.HandleKeyServerStats(stats.TypeJSON)).Methods(http.MethodGet)
+
+	r.Handle("/composite.csv", c.HandleComposite(stats.TypeCSV)).Methods(http.MethodGet)
```
sethvargo, line 442: This should still be nested under `realm`. I want to future-proof a stats API that is for the system.

internal/routes/server.go
```diff
@@ -438,6 +438,9 @@ func statsRoutes(r *mux.Router, c *stats.Controller) {
...
+	r.Handle("/composite.json", c.HandleComposite(stats.TypeJSON)).Methods(http.MethodGet)
```
sethvargo, line 443: If we keep the composite API, we need to document it (and its API support level) in docs/api.md.

pkg/controller/stats/composite.go
mikehelmick: dun. exported the one from the database, the other comes from the key server.

pkg/controller/stats/composite.go
```diff
@@ -0,0 +1,302 @@
...
+		for _, rs := range realmStats {
+			day := &CompositeDay{
+				Day:        rs.Date,
+				RealmStats: rs,
+			}
+			stats = append(stats, day)
+			statsMap[rs.Date] = day
+		}
+
+		if c.db.HasKeyServerStats(currentRealm.ID) {
```
sethvargo, line 240: I think this is adding two database calls when there should really only be 1. We should just try to get them and then check if the error is IsNotFound.

pkg/controller/stats/composite.go
```diff
@@ -0,0 +1,302 @@
...
+			days, err := c.db.ListKeyServerStatsDaysCached(ctx, currentRealm.ID, c.cacher)
+			if err != nil {
+				controller.InternalError(w, r, c.h, err)
+				return
+			}
+
+			needsSort := false
```
sethvargo, line 247: This feels like an unnecessary optimization. `sort.Slice` is O(n) for a sorted slice.

assets/server/realmadmin/_stats_codes.html
mikehelmick: or they haven't shown up yet for today...

mikehelmick: there's a race here and not all days will have key server status (min threshold)

pkg/database/key_server_stats.go
```diff
@@ -92,6 +92,19 @@ func (kssd *KeyServerStatsDay) BeforeSave(tx *gorm.DB) error {
...
 
+func (kssd *KeyServerStatsDay) TotalPublishRequests() int64 {
+	var sum int64
+	for _, v := range kssd.PublishRequests {
+		sum += v
+	}
+	return sum
+}
+
+func (db *Database) HasKeyServerStats(realmID uint) bool {
```
sethvargo, line 103: I'd suggest we drop this and have callers check the result of GetKeyServerStats. There's also an edge case here because this checks the live data, but then the code later checks the cached data.

pkg/database/realm_stats.go
```diff
@@ -65,6 +65,14 @@ type RealmStat struct {
 	CodeClaimMeanAge DurationSeconds `gorm:"column:code_claim_mean_age; type:bigint; not null; default: 0;"`
 }
 
+func (s *RealmStat) CodeClaimAgeDistributionAsStrings() []string {
```
sethvargo, line 68: Comment

pkg/database/key_server_stats.go
```diff
@@ -92,6 +92,19 @@ func (kssd *KeyServerStatsDay) BeforeSave(tx *gorm.DB) error {
 	return kssd.ErrorOrNil()
 }
 
+func (kssd *KeyServerStatsDay) TotalPublishRequests() int64 {
```
sethvargo, line 95: Comment

sethvargo (commented):
It might be a stupid question, but why don't we just include this information in the existing realm stats. Sure, it would be zeroed if key server stats weren't enabled, but I think that's okay?

assets/server/realmadmin/_stats_codes.html
sethvargo: You could also export a `hasKeyServerStats` field in the _top-level_ JSON which mirrors the config for the realm on the server.

pkg/controller/stats/composite.go
mikehelmick: I think it's actually a different thing - we should probably cache key server stats enabled on a realm though

pkg/database/key_server_stats.go
mikehelmick: GetKeyServerStats is checking for the existence of the row in the KeyServerStats table

The listing of the stats is reading KeyServerStatsDay

When we disable key server stats, we don't garbage collect old data - so presence of old data doesn't indicate if we should show it or not.

pkg/database/key_server_stats.go
```diff
@@ -111,17 +144,20 @@ func (db *Database) SaveKeyServerStats(stats *KeyServerStats) error {
...
+		Delete(&kss).
 		Error
 }
 
 // ListKeyServerStats retrieves the key-server statistics configuration for all realms
 func (db *Database) ListKeyServerStats() ([]*KeyServerStats, error) {
 	var stats []*KeyServerStats
 	if err := db.db.
-		Model(&KeyServerStatsDay{}).
+		Model(&KeyServerStats{}).
```
mikehelmick, line 160: this was a bug

pkg/database/database.go
```diff
@@ -368,6 +371,10 @@ func callbackPurgeCache(ctx context.Context, cacher cache.Cacher, namespace, tab
 			return
 		}
 
+		for _, f := range scope.Fields() {
+			log.Printf("field: %v value: %v", f.DBName, f.Field.Interface())
```
sethvargo, line 375: Debug?

pkg/database/key_server_stats.go
```diff
@@ -128,9 +144,12 @@ func (db *Database) SaveKeyServerStats(stats *KeyServerStats) error {
 
 // DeleteKeyServerStats disables gathering key-server statistics and removes the entry
 func (db *Database) DeleteKeyServerStats(realmID uint) error {
+	kss := KeyServerStats{
```
sethvargo, line 147:
```suggestion
	kss := &KeyServerStats{
```

pkg/database/key_server_stats.go
```diff
@@ -128,9 +144,12 @@ func (db *Database) SaveKeyServerStats(stats *KeyServerStats) error {
...
+		RealmID: realmID,
+	}
 	return db.db.Unscoped().
-		Where("realm_id = ?", realmID).
-		Delete(&KeyServerStats{}).
+		Set("gorm:delete_option", "RETURNING *").
+		Delete(&kss).
```
sethvargo, line 152:
```suggestion
		Delete(kss).
```

sethvargo (approved):
/lgtm

PR #1890, opened 2021-03-01
Add Indonesia (id) locale

dwisiswant0:
Fixes #

## Proposed Changes

* Add Indonesia (id) language

**Release Note**

```release-note
 - Indonesia (id) language translation
```

### DIFF
--- a/internal/i18n/locales/id/default.po
+++ b/internal/i18n/locales/id/default.po
@@ -1,0 +1,228 @@
+msgid ""
+msgstr ""
+"Language: id\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#
+# navigation menus
+# ----------------
+
+msgid "nav.issue-code"
+msgstr "Kode masalah"
+
+msgid "nav.bulk-issue-codes"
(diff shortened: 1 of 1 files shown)


dwisiswant0:
@googlebot I signed it!

mikehelmick:
Thanks for the contribution!

/ok-to-test
/hold

Checking on engagement with Indonesia.

sethvargo (dismissed):

internal/i18n/locales/id/default.po
```diff
@@ -0,0 +1,228 @@
...
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#
+# navigation menus
+# ----------------
+
+msgid "nav.issue-code"
+msgstr "Kode masalah"
```
sethvargo, line 14: Hey @dwisiswant0 thank you for the PR. I think this translation might be incorrect. "Issue" is a verb in this case (synonyms: create, generate, make). I think this translation to "trouble".

sethvargo: (For all instances of "issue"). I'm not a native speaker though, so please correct me if I'm wrong

dwisiswant0: Thanks for review this, looks like your context is correct. Should be resolved as per 63b4fee.

sethvargo:
Thank you!

Issue #1901, opened 2021-03-02
Set custom log retention thru terraform


bschlaman:
### TL;DR
Allow changing the default log retention via terraform

**Proposal**
Propose that the value be configurable per-service

**Resources**
Documentation on custom retention [here](https://cloud.google.com/logging/docs/storage#custom-retention)

bschlaman:
Per-service retention might not be possible; might have to be done on a per-bucket bases

PR #1902, opened 2021-03-02
Allow customizing log retention period

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/1901

Not marking auto-merge because I want to test manually.

**Release Note**

```release-note
Allow customizing global log retention period for all services in the project. The default value is 14 days. **Note: this differs from the unconfigured value of 30 days!**. To retain the existing behavior, set `log_retention_period` to `30` in the Terraform configuration. However, we strongly recommend using a 14-day retention period instead.
```

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -251,6 +251,12 @@ variable "enable_lb_logging" {
   EOT
 }
 
+variable "log_retention_period" {
+  type        = number
+  default     = 14
+  description = "Number of days to retain logs for all services in the project"
+}
+
 // Note: in Cloud Run/Knative, there are two kinds of annotations.
 // - Service level annotations: applies to all revisions in the service. E.g.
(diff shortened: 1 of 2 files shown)


terraform/variables.tf
```diff
@@ -251,6 +251,12 @@ variable "enable_lb_logging" {
   EOT
 }
 
+variable "log_retention_period" {
+  type        = number
+  default     = 30
```
mikehelmick, line 256: I think we should make the default 14 - we should also note it impacts ALL resources in the project

sethvargo: Updated and updated release notes.

mikehelmick (approved):

sethvargo:
I've tested this and confirm it works as expected. Enabling auto-merge.

PR #1905, opened 2021-03-04
Retry and parallelize stats puller

sethvargo:
This updates the stats-puller to retry the HTTP request 3 times. This will make us more resilient to temporary network blips. It also parallelizes pulling (default 5).

**Release Note**

```release-note
Add client-side retry logic and parallelize stats puller. The default parallelize is 5, but it can be customized with `STATS_PULLER_MAX_WORKERS`. There is also a **behavior change**. The stats-puller previously always returned success (but logged errors on failure). This changes the puller to return a non-200 response code if there are still failures after all retries have executed.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	go.opencensus.io v0.22.6
 	go.uber.org/zap v1.16.0
 	golang.org/x/oauth2 v0.0.0-20210216194517-16ff1888fd2e // indirect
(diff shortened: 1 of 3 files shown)


sethvargo:
https://github.com/google/exposure-notifications-verification-server/pull/1905/files?w=1 is a nicer review diff

whaught (approved):

mikehelmick (approved):

PR #1907, opened 2021-03-05
Set alert interval to 10 mins

mariliamelo:
This will create a cloud scheduler job failed alert if there have been errors in the last 10 minutes.

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -192,15 +192,15 @@ resource "google_monitoring_alert_policy" "CloudSchedulerJobFailed" {
     condition_monitoring_query_language {
       duration = "0s"
       # NOTE: The query below will be evaluated every 30s. It will look at the latest point that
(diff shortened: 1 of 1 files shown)


mikehelmick (approved):

sethvargo:
I think this will still alert if there have been any errors in 10min. We're seeing an alert once per 5min interval where any alerts have occurred during that interval. I think this would just increase the time before we are alerted to 10min.

mariliamelo:
No, we will alert if 100% of log messages were errors in the last 10 minutes. Right now we alert if 100% of log messages were errors in the last 5 minutes. Since we only run about once every 5 minutes, we end up alerting for every run.

sethvargo:
We run once every 5min, but retry 3 times. It failed and then immediately succeeded 3 seconds later... I don't think the MQL is behaving the way the comment advertises.

mariliamelo:
I see. I'll investigate this.

PR #1909, opened 2021-03-10
Set binauth annotation

sethvargo:
**Release Note**

```release-note
Set Binary Authorization service annotations on Cloud Run services.
```

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -279,6 +279,7 @@ locals {
     "run.googleapis.com/vpc-access-egress" : "private-ranges-only"
   }
   default_service_annotations = {
+    "run.googleapis.com/binary-authorization" : "default"
     "run.googleapis.com/ingress" : "all"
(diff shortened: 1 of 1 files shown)


mikehelmick (approved):

terraform/variables.tf
```diff
@@ -279,6 +279,7 @@ locals {
     "run.googleapis.com/vpc-access-egress" : "private-ranges-only"
   }
   default_service_annotations = {
+    "run.googleapis.com/binary-authorization" : "default"
```
steren, line 282: I wonder if using this requires "run.googleapis.com/launch-stage" : "ALPHA", let me ask for confirmation

steren: All good, I got confirmation that Beta will work.

sethvargo: Yea, it applied cleanly on two envs

PR #1913, opened 2021-03-11
Allow Cloud Run services to respond in the default timeout

sethvargo:
The default value is 5 minutes. I'm not sure why were configured these? Especially for the long-running jobs, I think we should allow the container to run for more than 25s.

**Release Note**

```release-note
Allow Cloud Run services to respond in the default timeout (5 minutes).
```


yuriatgoogle:
I had no input in configuring these, so I'm going to recluse myself from approving this.

mikehelmick (approved):

PR #1918, opened 2021-03-11
e2e-default should report client-side http errors

ethanmdavidson:
## Proposed Changes

* e2e runner default workflow should fail any time httpClient returns an error, instead of only when server responds with error.

**Release Note**

```release-note
e2e runner now fails on httpClient errors (e.g. timeout)
```

### DIFF
--- a/internal/clients/e2e.go
+++ b/internal/clients/e2e.go
@@ -269,7 +269,7 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 			}
 			publishResp, err := keyServerClient.Publish(ctx, publishReq)
 			defer logger.Debugw("publish", "request", publishReq, "response", publishResp)
-			if publishResp.ErrorMessage != "" {
+			if err != nil || publishResp.ErrorMessage != "" {
 				result = enobs.ResultNotOK
 				logger.Errorw("failed to publish teks", "error", err, "keys", teks)
 				return nil, fmt.Errorf("publish API error: %+v", publishResp)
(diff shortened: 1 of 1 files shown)


mikehelmick (dismissed):

internal/clients/e2e.go
```diff
@@ -269,7 +269,7 @@ func RunEndToEnd(ctx context.Context, cfg *config.E2ERunnerConfig) error {
 			}
 			publishResp, err := keyServerClient.Publish(ctx, publishReq)
 			defer logger.Debugw("publish", "request", publishReq, "response", publishResp)
-			if publishResp.ErrorMessage != "" {
+			if err != nil || publishResp.ErrorMessage != "" {
```
sethvargo, line 272: I think we need to take an approach like L244 and return a different error structure

sethvargo (commented):
That test failure seems legitimate

ethanmdavidson:
The test is failing because there's no key server. I tried using [enkstest](https://github.com/google/exposure-notifications-server/blob/main/internal/enkstest/server.go) but it seems like that can't be used cause it's internal to the other repo (I'm a go noob so maybe I'm just doing something wrong though). What's the best way to get this test passing?

sethvargo:
Ah fun. So, the reason the test is failing is because , without your fix, we swallow the failed request and continue along.

@mikehelmick thoughts on making `enkstest` public so we can spin up a key server as part of the e2e integration test?

mikehelmick:
> Ah fun. So, the reason the test is failing is because , without your fix, we swallow the failed request and continue along.
> 
> @mikehelmick thoughts on making `enkstest` public so we can spin up a key server as part of the e2e integration test?

no problem with that.

https://github.com/google/exposure-notifications-server/pull/1454

to make that package public

sethvargo:
@ethanmdavidson I'm going to pull your commit into another PR tomorrow that spins up a key server.

ethanmdavidson:
sounds good. Do you want me to close this?

sethvargo:
Hey @ethanmdavidson 

Let's leave it open because it serves as a reminder for me to fix this up. When I open the other PR, I'll include your commit and have it auto-close this one. Thank you again for the contribution!

PR #1919, opened 2021-03-12
restrict redirects to allowed paths

mikehelmick:
## Proposed Changes

* Ensure redirects only go to necessary URLs post redirect

**Release Note**

```release-note
Re-auth login redirect uses allowlist for post auth redirects.
```

### DIFF
--- a/pkg/controller/login/reauth.go
+++ b/pkg/controller/login/reauth.go
@@ -21,21 +21,34 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 )
 
+var (
+	allowedRedirects = map[string]struct{}{
+		"login/register-phone": struct{}{},
+	}
+)
+
+func redirectAllowed(r string) bool {
+	_, ok := allowedRedirects[r]
+	return ok
+}
+
 func (c *Controller) HandleReauth() http.Handler {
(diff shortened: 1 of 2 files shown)


pkg/controller/login/reauth.go
```diff
@@ -21,21 +21,34 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 )
 
+var (
+	allowedRedirects = map[string]struct{}{
+		"login/register-phone": struct{}{},
+	}
+)
+
+func redirectAllowed(r string) bool {
```
sethvargo, line 30: Can you document this and why it's important?

pkg/controller/login/reauth.go
```diff
@@ -21,21 +21,34 @@ import (
...
+}
+
 func (c *Controller) HandleReauth() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 
 		// No session redirect for reauth
 
-		if r := r.FormValue("redir"); r != "" {
+		if r := r.FormValue("redir"); redirectAllowed(r) {
```
sethvargo, line 41: I think we need a test for this.

sethvargo (approved):

PR #1927, opened 2021-03-19
Rename CleanupStatus to LockStatus

sethvargo:
Also fix an issue where modeler would return a non-200 when the lock was already held, causing Cloud Scheduler to fail.

**Release Note**

```release-note
Rename CleanupStatus to LockStatus and fix an issue where modeler would return a non-200 when the lock was already held, causing Cloud Scheduler to fail.
```


sethvargo:
@mikehelmick CI is happy now

mikehelmick (approved):

PR #1929, opened 2021-03-20
Switch to forward-progress alerting

sethvargo:
This changes most of our background jobs to alert when forward-progress is not achieved instead of on specific failures. This did require we increase the frequency of the modeler and appsync workers because of how Cloud Monitoring calculates windows.

Part of https://github.com/google/exposure-notifications-verification-server/issues/1777:

- [x] appsync-worker - appsync runs every 4h, alert after 2 failures
- [ ] backup-database-worker - alert on all failures, page, playbook should have us run a manual backup
- [x] cleanup worker - cleanup runs every 1h, alert after 4 failures
- [x] ~docker-mirror-worker - email on failures, do not page.~ no longer in use
- [ ] e2e - page on each failure, but not if the scheduler fails to reach the container
- [x] modeler - modeler runs every 4h, alert after 2 failures
- [x] realm-key-rotation-worker - realm-key-rotation runs every 15m, alert after 2 failures
- [x] rotation worker - rotation runs every 30m, alert after 2 failures
- [x] stats-puller - stats-puller runs every 15m, alert after 2 failures

There's a 5min buffer on the end to give us enough time in case a later attempt _just_ succeeded to avoid the alert.

**Release Note**

```release-note
Switch to forward-progress alerting for most background jobs. See the updated ForwardProgressFailed.md documentation for more information.
```


pkg/controller/rotation/metrics.go
```diff
@@ -59,14 +59,14 @@ func init() {
 			Aggregation: view.Count(),
 		},
 		{
-			Name:        metricPrefix + "/token_success",
```
sethvargo, line 62: Had to rename these so we can use the same MQL query for .success (they are new metrics I just created in the last PR, so no prod impact).

mikehelmick (approved):

PR #1931, opened 2021-03-21
Re-add name, remove playbooks test

sethvargo:
I fixed this to get the tests to pass, but then I realized it's the wrong change. We can't have multiple alerting policies with the same display name. Unfortunately there's no way to make the test continue to work since the alerting policies are now defined dynamically.

**Release Note**

```release-note
NONE
```


mikehelmick (approved):

yegle:
Is there a way to actually evaluate the config file, as opposed to simply parse it semantically?

sethvargo:
@yegle the packages that actually parse Terraform are private :/

Issue #1928, opened 2021-03-20
User initiated report


mikehelmick:
### TL;DR

add an API for user initiated report

### Design

**Proposal**

* UserReport API on apiserver
* Phone number requires, SMS only - 90d de-duplication
* requires 256 bytes of nonce so that same device claims code
* cleanup changes

mikehelmick:
Still todo

- Specific statistics for self report

PR #1945, opened 2021-03-24
add key revisions back to key server chart iff user report is enabled

mikehelmick:
Towards #1928

## Proposed Changes

* add revisions back to key sever chart

**Release Note**

```release-note
Number of revision request are shown on key server stats if user initiated reporting is enabled. This is the only EN Express scenario where key revision could occurr.
```

### DIFF
--- a/assets/server/realmadmin/_stats_keyserver.html
+++ b/assets/server/realmadmin/_stats_keyserver.html
@@ -1,5 +1,7 @@
 {{define "realmadmin/_stats_keyserver"}}
 
+{{$realm := .currentMembership.Realm}}
+
 <h1>Key server stats</h1>
 <p>
   The data below shows realm statistics and visualizations gathered from the key server.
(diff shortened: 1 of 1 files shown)


assets/server/realmadmin/_stats_keyserver.html
```diff
@@ -295,6 +297,9 @@ <h5 class="modal-title">Onset-to-upload distribution</h5>
     dataTable.addColumn('number', 'Total');
     dataTable.addColumn({type:'string', role:'tooltip', p: {html: true}});
     dataTable.addColumn('number', 'Missing Onset Date');
+    {{if $realm.AllowsUserReport}}
```
sethvargo, line 300: I'd really prefer if we did this at the javascript layer

mikehelmick: done, although it makes the JS more complicated

sethvargo: Yea, it just enables us to move the JS into a CDN in the future 😄

sethvargo (approved):

PR #1955, opened 2021-03-26
allow admin /api/issue to issue user-report codes

mikehelmick:
Towards #1928 

## Proposed Changes

* Add realm setting to enable "admin" API user-reports
  * these don't require a nonce
  * do require a phone number and still de-dupe against the HMAC user_report phone index

**Release Note**

```release-note
All admin actions to issue user-report level verification codes, that can result in SELF_REPORT
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2181,7 +2181,11 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`ALTER TABLE realm_stats
 						ALTER COLUMN user_reports_issued SET NOT NULL,
 						ALTER COLUMN user_reports_claimed SET NOT NULL,
-						ALTER COLUMN user_report_tokens_claimed SET NOT NULL`)
+						ALTER COLUMN user_report_tokens_claimed SET NOT NULL`,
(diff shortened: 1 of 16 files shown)


pkg/database/migrations.go
```diff
@@ -2181,7 +2181,11 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`ALTER TABLE realm_stats
 						ALTER COLUMN user_reports_issued SET NOT NULL,
 						ALTER COLUMN user_reports_claimed SET NOT NULL,
-						ALTER COLUMN user_report_tokens_claimed SET NOT NULL`)
+						ALTER COLUMN user_report_tokens_claimed SET NOT NULL`,
+					`ALTER TABLE user_reports
```
sethvargo, line 2185: This needs to be a different migration since this migration has already been run

sethvargo (changes requested):

pkg/database/migrations.go
mikehelmick: fixed.

sethvargo (approved):

PR #1957, opened 2021-03-27
bootstrap e2e runner on first run instead of startup

mikehelmick:
## Proposed Changes

* bootstrap e2e runner on first run

**Release Note**

```release-note
e2e-runner will bootstrap the e2e test realm on first invocation instead of on startup. This prevents deploy failures when new columns are added to the realm table during an upgrade. 
```

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 	"time"
 
@@ -35,10 +36,17 @@ import (
 	"github.com/google/exposure-notifications-verification-server/internal/project"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/middleware"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
 
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
+)
+
+var (
+	mu            sync.Mutex
+	setupComplete bool = false
+	closeFunc     func() error
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/e2e-runner/main.go
```diff
@@ -174,6 +169,54 @@ func realMain(ctx context.Context) error {
 	return srv.ServeHTTPHandler(ctx, mux)
 }
 
+func bootstrap(ctx context.Context, db *database.Database, cfg *config.E2ERunnerConfig) mux.MiddlewareFunc {
+	logger := logging.FromContext(ctx).Named("e2erunner.bootstrap")
+	bootstrapFn := func() error {
```
sethvargo, line 174: Wouldn’t this be simpler with sync.Once?

mikehelmick: it's not a true "once" see, the note on line 201

sethvargo: I see

cmd/e2e-runner/main.go
```diff
@@ -174,6 +169,56 @@ func realMain(ctx context.Context) error {
...
+				has, err := resp.Realm.HasSMSConfig(db)
+				if err != nil {
+					return fmt.Errorf("failed to check if realm has sms config: %w", err)
+				}
+				if !has {
+					return fmt.Errorf("realm does not have sms config, configure it or set E2E_SKIP_SMS to continue")
+				}
+			}
+
+			cfg.VerificationAdminAPIKey = resp.AdminAPIKey
```
sethvargo, line 198: This is shared state, but it's only guarded by the mutex in bootstrap. The handlers also access and in some cases modify this resource. It might be fine, but I'm not 100% familiar with how the middleware is handled when there's shared state with a child handler? 

If we're going to do this, we might be better storing the config on the request context using context.WithValue inside the middleware?

I'm worried this is going to cause a subtle race condition that will be impossible to debug.

mikehelmick: The handlers don't get to run unless this has successfully run.

I could put more guards around it, but it's not strictly necessary. We already live with our config structs being pointers to shared state across multiple controllers and handlers.

sethvargo (approved):

PR #1958, opened 2021-03-28
fix os chart stack + nil ptr

mikehelmick:
## Proposed Changes

* fix nil ptr in composite.go
* fix series settings when self report is off

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

PR #1982, opened 2021-04-06
only redirect for known host headers

mikehelmick:
## Proposed Changes

* only redirect for known host headers

**Release Note**

```release-note
Change load balancer https redirect rules to only match known host names, otherwise redirect to UI server.
```

### DIFF
--- a/terraform/verification-lb.tf
+++ b/terraform/verification-lb.tf
@@ -60,7 +60,7 @@ resource "google_compute_url_map" "urlmap-http" {
     for_each = length(var.server_hosts) > 0 ? [1] : []
 
     content {
-      name            = "server"
+      name = "server"
       default_url_redirect {
         host_redirect  = var.server_hosts[0]
         https_redirect = true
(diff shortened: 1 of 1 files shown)


terraform/verification-lb.tf
```diff
@@ -46,9 +46,81 @@ resource "google_compute_url_map" "urlmap-http" {
...
+        host_redirect  = var.adminapi_hosts[0]
+        https_redirect = true
+        strip_query    = true
+      }
+    }
+  }
+
+  // any other hosts, just get redirected to the UI server
   default_url_redirect {
-    strip_query    = false
```
sethvargo, line 50: I'm not sure we want to strip the query string - that could break an existing http link

sethvargo (changes requested):

terraform/verification-lb.tf
mikehelmick: I only put it on this one, because this is when the incoming host header is not valid, so I think it's best to also strip the query. 

I'll make the known hosts false.

sethvargo (approved):

PR #1989, opened 2021-04-06
add some UX fixes

mikehelmick:
## Proposed Changes

* don't let greyed out fields be changed
* alert if user exists

**Release Note**

```release-note
Don't let greyed out fields be changed in signing keys.
Alert when creating a user that already exists.
Fix ENX upgrade SMS template text when user-report is enabled.
```

### DIFF
--- a/pkg/controller/realmadmin/settings_modify.go
+++ b/pkg/controller/realmadmin/settings_modify.go
@@ -233,7 +233,8 @@ func (c *Controller) HandleSettings() http.Handler {
 			if c.config.Features.EnableUserReport && form.AllowUserReport {
 				currentRealm.AddUserReportToAllowedTestTypes()
 				currentRealm.AllowAdminUserReport = form.AllowAdminUserReport
-			} else {
+			}
+			if !c.config.Features.EnableUserReport {
 				currentRealm.AllowAdminUserReport = false
 			}
 
--- a/pkg/controller/realmadmin/settings_modify.go
+++ b/pkg/controller/realmadmin/settings_modify.go
@@ -230,12 +230,13 @@ func (c *Controller) HandleSettings() http.Handler {
 			currentRealm.AllowedTestTypes = form.AllowedTestTypes
 			currentRealm.RequireDate = form.RequireDate
 			currentRealm.AllowBulkUpload = form.AllowBulkUpload
-			if c.config.Features.EnableUserReport && form.AllowUserReport {
-				currentRealm.AddUserReportToAllowedTestTypes()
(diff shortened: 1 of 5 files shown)


pkg/controller/user/create.go
```diff
@@ -111,6 +111,9 @@ func (c *Controller) HandleCreate() http.Handler {
 			return
 		}
 
+		if existing != nil && existing.ID != 0 {
+			flash.Alert("A user with this email address already exists, so the exiting name has been used.")
```
sethvargo, line 115: I intentionally didn't do this because it leaks information. This tells the admin with UserWrite permissions that the person they just added was already a member of another realm.

pkg/controller/realmadmin/settings_modify.go
```diff
@@ -233,7 +233,8 @@ func (c *Controller) HandleSettings() http.Handler {
 			if c.config.Features.EnableUserReport && form.AllowUserReport {
 				currentRealm.AddUserReportToAllowedTestTypes()
 				currentRealm.AllowAdminUserReport = form.AllowAdminUserReport
-			} else {
+			}
+			if !c.config.Features.EnableUserReport {
```
sethvargo, line 237: This will have the (potentially intended) side-effect of disabling user-report in the following scenario:

1. Enabled at system level
2. Realm enables
3. Disabled at system level
4. Realm updates something else on that page

Is that intended?

pkg/controller/realmkeys/save.go
```diff
@@ -61,8 +61,11 @@ func (c *Controller) HandleSave() http.Handler {
 		}
 
 		// Update settings.
-		currentRealm.CertificateIssuer = form.Issuer
-		currentRealm.CertificateAudience = form.Audience
+		if !currentRealm.UseRealmCertificateKey {
```
sethvargo, line 64: Should this be an error instead?

sethvargo (changes requested):

pkg/controller/user/create.go
mikehelmick: fair point - removed.

pkg/controller/realmadmin/settings_modify.go
mikehelmick: good catch - fixed for what I intended it to be.

pkg/controller/realmkeys/save.go
mikehelmick: I went back and forth on this - I went w/ the easier thing since this is unlikely to happen to a human not trying to get around this.

sethvargo (approved):

PR #1991, opened 2021-04-06
add warning about UR and auth SMS

mikehelmick:
## Proposed Changes

* add warning on settings page

**Release Note**

```release-note
Add a warning on realm settings page that when user report is enabled, authenticated SMS should be enabled too.
```

### DIFF
--- a/assets/server/realmadmin/_form_codes.html
+++ b/assets/server/realmadmin/_form_codes.html
@@ -87,6 +87,16 @@
     </div>
 
     {{if .features.EnableUserReport}}
+    <label>User report</label>
+
+    {{if and $realm.AllowsUserReport (not $realm.UseAuthenticatedSMS)}}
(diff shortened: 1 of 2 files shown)


assets/server/realmadmin/_form_codes.html
```diff
@@ -87,6 +87,16 @@
     </div>
 
     {{if .features.EnableUserReport}}
+    <label>User report</label>
+
+    {{if and $realm.AllowsUserReport (not $realm.UseAuthenticatedSMS)}}
+    <div class="alert alert-warning" role="alert">
+      This realm has user report enabled, but authenticated SMS disabled. It is recommended
```
whaught, line 94: "It is recommended to enable authenticated SMS is enabled as"

mikehelmick: fixed

whaught (approved):

Issue #1970, opened 2021-04-02
Login pages need translations


mikehelmick:
### TL;DR

User login pieces need to be translated

There are a few other items.

PR #1999, opened 2021-04-07
Add translations for email verify pages

sethvargo:
I think this is the last "login" page that needs translations.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/1970

**Release Note**

```release-note
Add translations for email verify pages.
```

### DIFF
--- a/assets/server/login/verify-email-check.html
+++ b/assets/server/login/verify-email-check.html
@@ -15,13 +15,20 @@
 
     <div class="d-flex vh-100">
       <div class="d-flex w-100 justify-content-center">
-        <div class="col-sm-6">
-
-          <div class="card mb-3 shadow-sm">
-            <div class="card-header">Email verification</div>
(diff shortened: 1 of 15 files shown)


assets/server/login/verify-email-check.html
```diff
@@ -31,32 +38,28 @@
...
+        showError();
+        return;
       }
 
       firebase.auth().applyActionCode(code)
         .then(function(resp) {
-          window.location.assign("/");
-        }).catch(function(error) {
-          flash.clear();
-          flash.error("Invalid email verification code. "
```
whaught, line 45: No flash?

sethvargo: It's rendered in HTML on L28. Go's templating language doesn't work when you try to embed the translated string into a javascript string.

whaught (approved):

PR #2014, opened 2021-04-09
Cleanup e2e API keys that have existed for > 24 hours

sethvargo:
The e2e-runner does self cleanup, but sometimes the job times out or fails. In those cases, there's no opportunity for cleanup and we gradually leak API keys over time.

This introduces a function for getting the e2e-realm (if one exists), and adds a new task to the cleanup job to delete e2e keys that have existed for more than 24 hours.

**Release Note**

```release-note
Cleanup e2e API keys that have existed for > 24 hours.
```

### DIFF
--- a/pkg/database/authorized_app.go
+++ b/pkg/database/authorized_app.go
@@ -474,3 +474,25 @@ func (db *Database) PurgeAuthorizedApps(maxAge time.Duration) (int64, error) {
 		Delete(&AuthorizedApp{})
 	return result.RowsAffected, result.Error
 }
+
+// PurgeE2EAuthorizedApps purges any API keys for the e2e-runner that have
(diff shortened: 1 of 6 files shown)


whaught (approved):

pkg/database/authorized_app.go
```diff
@@ -474,3 +474,24 @@ func (db *Database) PurgeAuthorizedApps(maxAge time.Duration) (int64, error) {
...
+	if err != nil {
+		if IsNotFound(err) {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("failed to lookup e2e realm: %w", err)
+	}
+
+	result := db.db.
+		Unscoped().
+		Where("realm_id = ? AND created_at < ?", realm.ID, deleteBefore).
```
mikehelmick, line 494: I think this should always leave the most recent in place

PR #2019, opened 2021-04-12
Lower Cloud KMS database-encrypter rotation to 90d

sethvargo:
**Release Note**

```release-note
Lower Cloud KMS database-encrypter rotation to 90d
```

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -186,7 +186,7 @@ manager, provide this value directly in the environment.
 
 ### Database encryption keys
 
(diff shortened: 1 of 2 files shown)


whaught (approved):
Lower? Aren't we raising it from 30 here?

sethvargo:
> Lower? Aren't we raising it from 30 here?

We're lowering the frequency (every 30d to 90d). I do see how this is confusing though.

PR #2022, opened 2021-04-12
Track API key "last used"

sethvargo:
This isn't an exact timestamp tracker, since we use some pretty aggressive caching to avoid a write on every read. However, this will give realm admins a very clear signal of whether an API key is still being used.

The UI is hidden behind a feature flag because I want to capture this data for a full release until displaying it in the UI. Otherwise it could lead to misleading information about an API having "never" been used.

**Release Note**

```release-note
Track API key "last used"
```

### DIFF
--- a/assets/server/apikeys/index.html
+++ b/assets/server/apikeys/index.html
@@ -49,8 +49,11 @@
             <tr>
               <th scope="col" width="40"></th>
               <th scope="col">App</th>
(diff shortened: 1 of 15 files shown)


sethvargo:
screenshot-20210412-154835@2x

whaught (approved):

PR #2024, opened 2021-04-12
Do not store firebase cookie and verify ID token is < 5min old

sethvargo:
I believe this will address the issue where a user is logged out and then quickly logged back in. We're using server-side sessions, but Firebase is still setting a cookie. So even when the session expires, Firebase re-authenticates automatically. This follows the instructions at https://firebase.google.com/docs/auth/admin/manage-cookies to disable persisting the Firebase cookie. There's also a server-side check that the auth token was created within the past 5 minutes for any Firebase cookies that might already be out in the wild.

**Release Note**

```release-note
Do not store firebase cookie and verify ID token is < 5min old.
```

### DIFF
--- a/assets/server/firebase.html
+++ b/assets/server/firebase.html
@@ -17,6 +17,9 @@
 
   window.addEventListener('load', () => {
     firebase.initializeApp(firebaseConfig);
+
+    // Disable the firebase cookie - we manage the ID token on the server side.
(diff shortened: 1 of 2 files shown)


sethvargo:
@whaught I _think_ this will fix the magical re-auth that happens

whaught (approved):

PR #2031, opened 2021-04-15
Add secret models and structure to info page

sethvargo:
This is just the models; trying to split this up to make it easier to review.

screenshot-20210415-171255@2x

Note that there won't be any secrets in the info page for now, since nothing is actually connected.

**Release Note**

```release-note
Add secret models and structure to info page.
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2245,6 +2245,30 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 					`DROP INDEX IF EXISTS idx_authorized_apps_last_used_at`)
 			},
 		},
+		{
+			ID: "00103-AddSecrets",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`CREATE TABLE secrets (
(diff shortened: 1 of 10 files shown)


pkg/database/migrations.go
```diff
@@ -2245,6 +2245,30 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+			ID: "00103-AddSecrets",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`CREATE TABLE secrets (
+  					id BIGSERIAL PRIMARY KEY,
+  					type TEXT NOT NULL,
+  					reference TEXT NOT NULL,
+  					active BOOL DEFAULT FALSE NOT NULL,
+  					created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  					updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
```
sethvargo, line 2258: updated_at isn't used, but gorm always tries to set it if created_at exists 🙄

whaught (approved):

PR #2032, opened 2021-04-15
Extract key/secret bootstrapping into functions for seeding

sethvargo:
This also adds the secrets.SecretVersionManager on the rotation controller but does not use it yet.

**Release Note**

```release-note
Extract key/secret bootstrapping from rotation controller into functions for seeding.
```

### DIFF
--- a/tools/seed/main.go
+++ b/tools/seed/main.go
@@ -32,13 +32,16 @@ import (
 	firebaseauth "firebase.google.com/go/auth"
 	"github.com/google/exposure-notifications-verification-server/pkg/api"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
+	"github.com/google/exposure-notifications-verification-server/pkg/controller/rotation"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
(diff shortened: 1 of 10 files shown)


tools/seed/main.go
```diff
@@ -587,3 +607,29 @@ func createFirebaseUser(ctx context.Context, firebaseAuth *firebaseauth.Client,
...
+
+	h, err := render.New(ctx, nil, cfg.DevMode)
+	if err != nil {
+		return fmt.Errorf("failed to create renderer: %w", err)
+	}
+
+	rotationController := rotation.New(cfg, db, keyManager, secretManager, h)
+
+	if err := rotationController.RotateTokenSigningKey(ctx); err != nil {
+		return fmt.Errorf("failed to create initial token signing key: %w", err)
```
whaught, line 627:
```suggestion
		return fmt.Errorf("failed to rotate token signing key: %w", err)
```
are these always initial?

sethvargo: I don't think it's actually supported, but _technically_ if you ran the seed script again, it would rotate them.

whaught (approved):

PR #2039, opened 2021-04-20
Add automated secrets rotation

sethvargo:
**Release Note**

```release-note
**Add automated secrets rotation.** This introduces automated rotation for most application-level secrets. Whereas previously it was the responsibility of server administrators to rotate secrets, the application will now rotate a variety of secrets on regular intervals for improved security.
```

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -425,6 +425,14 @@ func (db *Database) SetRawDB(tx *gorm.DB) {
 	db.db = tx
 }
 
+// SetSecretResolver sets the underlying secret resolver. This is publicly exposed for
+// tests.
+func (db *Database) SetSecretResolver(r *SecretResolver) {
+	db.dbLock.Lock()
+	defer db.dbLock.Unlock()
+	db.secretResolver = r
+}
+
 // IsNotFound determines if an error is a record not found.
 func IsNotFound(err error) bool {
 	return errors.Is(err, gorm.ErrRecordNotFound) || gorm.IsRecordNotFoundError(err)
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -75,6 +75,9 @@ type Database struct {
 	// secretManager is used to resolve secrets.
 	secretManager secrets.SecretManager
 
+	// secretResolver is used for resolving secrets.
+	secretResolver *SecretResolver
+
 	statsCloser func()
 }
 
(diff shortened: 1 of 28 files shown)


pkg/database/database.go
```diff
@@ -103,6 +106,132 @@ func (db *Database) KeyManager() keys.KeyManager {
 	return db.keyManager
 }
 
+// GetCookieHashAndEncryptionKeys gets the cookie hash and encryption keys. The
+// first 32 bytes are the encryption key and the remaining bytes are the HMAC
+// key.
+func (db *Database) GetCookieHashAndEncryptionKeys(fallback []envconfig.Base64Bytes) ([][]byte, error) {
+	ctx, done := context.WithTimeout(context.Background(), 5*time.Second)
```
sethvargo, line 113: Ideally this context would flow through all callers, but that's a fairly large refactor that I didn't want to tackle here. What do y'all think?

pkg/database/database.go
```diff
@@ -103,6 +106,132 @@ func (db *Database) KeyManager() keys.KeyManager {
...
+
+	// TODO(sethvargo): remove in 0.28.0+
+	if len(results) == 0 {
+		existing := envconfigBytesToBytes(fallback)
+		if l := len(existing); l%2 != 0 {
+			return nil, fmt.Errorf("fallback bytes are not even (got %d)", l)
+		}
+
+		// In v1 cookies, the HMAC key is first, then the encryption key.
+		final := make([][]byte, len(existing)/2)
```
sethvargo, line 129: I hate this.

terraform/service_rotation.tf
```diff
@@ -57,10 +57,17 @@ locals {
   ])
 }
 
+resource "google_secret_manager_secret_iam_member" "rotation-secrets-accessor" {
```
sethvargo, line 60: I forgot that versionManager doesn't actually include permissions to access the secret. Oops.

pkg/database/database.go
mikehelmick: I can live w/ this

mikehelmick: and refactor later...

pkg/database/database.go
mikehelmick: it gets removed in the next release at least :)

mikehelmick (commented):
LGTM - I'll approve when tests are passing.

Question: How will our local dev environments upgrade / seed in the new world?

Should secrets be added to the seed script?

sethvargo:
> Question: How will our local dev environments upgrade / seed in the new world?

Local dev won't upgrade. Drop and re-create the database. You can remove the associated secrets from your .env file.

> Should secrets be added to the seed script?

They already are - that's the reason the Rotate function is public. It was in the previous PR.

sethvargo:
Don't merge. There's an occasional test failure when running in race mode that I'm looking into...

mikehelmick (approved):

PR #2040, opened 2021-04-20
add end to end runner for user-report test type

mikehelmick:
Last non web UI thing for  #1928

## Proposed Changes

* add end to end test for user-report
* ensure that user report SMS template is used for admin API requests codes (as a default)

**Release Note**

```release-note
Adds end to end test runner for user-report if that feature is enabled.
```

### DIFF
--- a/cmd/e2e-runner/main.go
+++ b/cmd/e2e-runner/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/internal/project"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller/middleware"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
 
 	"github.com/gorilla/handlers"
@@ -156,8 +157,9 @@ func realMain(ctx context.Context) error {
 	recovery := middleware.Recovery(h)
 	r.Use(recovery)
 
-	r.Handle("/default", handleDefault(cfg, h))
-	r.Handle("/revise", handleRevise(cfg, h))
+	r.Handle("/default", handleDefault(cfg, db, h))
+	r.Handle("/revise", handleRevise(cfg, db, h))
+	r.Handle("/user-report", handleUserReport(cfg, db, h))
 	r.Handle("/enx-redirect", handleENXRedirect(enxRedirectClient, h))
 
 	mux := http.Handler(r)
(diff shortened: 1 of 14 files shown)


cmd/e2e-runner/main.go
```diff
@@ -175,26 +177,63 @@ func realMain(ctx context.Context) error {
...
+
+// handleUserReport runs the end-to-end runner initiated by a user-report API request.
+func handleUserReport(cfg *config.E2ERunnerConfig, db *database.Database, h *render.Renderer) http.Handler {
+	if !cfg.Features.EnableUserReport {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+			logger := logging.FromContext(ctx)
+			logger.Warnw("user-report feature is not enabled on this server")
+			stats.Record(ctx, mUserReportSuccess.M(1))
+			h.RenderJSON(w, http.StatusOK, nil)
```
sethvargo, line 203: Should this actually return an error instead?

pkg/integration/enx_redirect_test.go
```diff
@@ -42,9 +42,14 @@ func TestENXRedirect(t *testing.T) {
...
-	realm.SMSTextTemplate = "[enslink]"
+	template := "[enslink]"
+	realm.SMSTextTemplate = template
+	for k := range realm.SMSTextAlternateTemplates {
+		realm.SMSTextAlternateTemplates[k] = &template
+	}
+
 	if err := server.Database.SaveRealm(realm, database.SystemTest); err != nil {
-		t.Fatal(err)
+		t.Fatalf("err: %v, errors: %+v", err, realm.ErrorMessages())
```
sethvargo, line 52: Can we add more content to this message if we're going to customize it?

cmd/e2e-runner/main.go
```diff
@@ -175,26 +177,63 @@ func realMain(ctx context.Context) error {
...
 // handleEndToEnd handles the common end-to-end scenario. m is incremented iff
 // the run succeeds.
-func handleEndToEnd(cfg *config.E2ERunnerConfig, h *render.Renderer, m *stats.Int64Measure) http.Handler {
+func handleEndToEnd(cfg *config.E2ERunnerConfig, db *database.Database, h *render.Renderer, m *stats.Int64Measure) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		logger := logging.FromContext(ctx)
 
+		if cfg.DoUserReport {
+			userReport, err := db.FindUserReport(db.RawDB(), project.TestPhoneNumber)
```
sethvargo, line 221: This API is a little weird to me. If `db` is the handler, why do we need to pass in a `db` as a param?

cmd/e2e-runner/main.go
mikehelmick: no, I think this is success.

it allows us to schedule the terraform and monitor the metric even if the feature is disabled.

cmd/e2e-runner/main.go
mikehelmick: I refactored this into user_report.go

pkg/integration/enx_redirect_test.go
mikehelmick: I put this back, this was debugging.

pkg/database/user_report.go
```diff
@@ -108,6 +108,22 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
 	return &ur, nil
 }
 
+// DeleteUserReport removes a specific phone number from the user report
+// de-duplication table.
+func (db *Database) DeleteUserReport(phoneNumber string) error {
+	return db.db.Transaction(func(tx *gorm.DB) error {
+		ur, err := db.FindUserReport(tx, phoneNumber)
```
sethvargo, line 115: Why do we need to lookup before deleting? It feels like this function is declarative (after I call it, the UR doesn't exist). Do we need to process the situation in which the record doesn't exist?

```go
db.db.Where("phone = ?", phoneNumber).Delete(&UserReport{})
```

mikehelmick: true - it also saves a transaction.

sethvargo (approved):

PR #2046, opened 2021-04-21
decreasing threshold for hostdown alert to improve alert times for DN…

bschlaman:
decreasing threshold for hostdown alert to improve alert times for DNS related service degradation

## Proposed Changes

* Set hostdown to alert when service degrades to under 60% availability over 5m instead of 20%

**Release Note**

```release-note
lowering threshold for hostdown alert
```


bschlaman:
@mariliamelo FYI

sethvargo (approved):

PR #2047, opened 2021-04-21
Timing

mikehelmick:
## Proposed Changes

* System admin can change a short code max expiration to between 60 and 120 minute s
* System admin can allow an ENX realm to change their short code expiration

Admin screen

image

ENX realm w/ editable code duration

image

**Release Note**

```release-note
System admins can allow for a domain to have longer short code expiration times (up to 2 hours) and for that realm to edit their short code expiration time even if ENX is enabled
```

### DIFF
--- a/assets/server/admin/realms/edit.html
+++ b/assets/server/admin/realms/edit.html
@@ -43,10 +43,13 @@
             <label for="code">Region code</label>
           </div>
 
-          <h6 class="mb-3">Verification Code Settings</h6>
-          <div class="form-label-group">
+          <h6 class="mb-3">Verification code settings</h6>
+          <div class="form-label-group input-group">
             <input name="short_code_max_minutes" id="short-code-max-minutes" type="text" class="form-control" value="{{$realm.ShortCodeMaxMinutes}}"  />
-            <label for="short-code-max-minutes">Short Code Max Minutes</label>
+            <span class="input-group-append">
+              <span class="input-group-text bg-transparent border-left-0">minutes</span>
+            </span>
+            <label for="short-code-max-minutes">Short code max duration</label>
           </div>
 
           <div class="form-check mb-3">
--- a/assets/server/admin/realms/edit.html
+++ b/assets/server/admin/realms/edit.html
@@ -43,6 +43,24 @@
             <label for="code">Region code</label>
           </div>
 
+          <h6 class="mb-3">Verification Code Settings</h6>
+          <div class="form-label-group">
+            <input name="short_code_max_minutes" id="short-code-max-minutes" type="text" class="form-control" value="{{$realm.ShortCodeMaxMinutes}}"  />
+            <label for="short-code-max-minutes">Short Code Max Minutes</label>
+          </div>
+
+          <div class="form-check mb-3">
+            <input type="checkbox" name="enx_code_expiration_configurable" id="enx-code-expiration-configurable" class="form-check-input" value="true" {{checkedIf ($realm.ENXCodeExpirationConfigurable)}} />
+            <label for="enx-code-expiration-configurable" class="form-check-label">
+              Allow realm short code configuration
+              <small class="form-text text-muted mb-3">
+                If EN Express is enabled, the short code expiration time is normally
+                fixed at 15 minutes. By enabling this setting, a realm admin can customize
+                their short code expiration time even with EN Express enabled.
+              </small>
+            </label>
+          </div>
+
           {{if .supportsPerRealmSigning}}
           <hr>
           <h6 class="mb-3">Certificate</h6>
--- a/pkg/controller/realmadmin/settings_modify.go
+++ b/pkg/controller/realmadmin/settings_modify.go
@@ -31,7 +31,6 @@ import (
 
 var (
 	shortCodeLengths            = []int{6, 7, 8}
-	shortCodeMinutes            = []int{}
 	longCodeLengths             = []int{12, 13, 14, 15, 16}
 	longCodeHours               = []int{}
 	mfaGracePeriod              = []int64{0, 1, 7, 30}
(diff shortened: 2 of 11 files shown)


pkg/database/migrations.go
```diff
@@ -2269,6 +2269,25 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
 		},
+		{
+			ID: "00104-AddCodeOverrides",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS short_code_max_minutes SMALLINT DEFAULT 60,
+						ADD COLUMN IF NOT EXISTS enx_code_expiration_configurable BOOL DEFAULT false`,
+					`ALTER TABLE realms
+						ALTER COLUMN short_code_max_minutes SET NOT NULL,
```
sethvargo, line 2280: Need to SET DEFAULT in another stanza (no default will be set if the column already exists)

pkg/database/migrations.go
```diff
@@ -2269,6 +2269,25 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+						ALTER COLUMN enx_code_expiration_configurable SET NOT NULL`,
+				)
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						DROP COLUMN short_code_max_minutes,
```
sethvargo, line 2287: IF EXISTS

pkg/database/realm.go
```diff
@@ -170,6 +169,13 @@ type Realm struct {
 	LongCodeLength   uint            `gorm:"type:smallint; not null; default: 16;"`
 	LongCodeDuration DurationSeconds `gorm:"type:bigint; not null; default: 86400;"` // default 24h
 
+	// ShortCodeMaxMinutes can only be set by system admins and allows for a
+	// realm to have a higher max short code duration
+	ShortCodeMaxMinutes uint
```
sethvargo, line 174: Unfortunately, since realm is already managed by gorm, you need the gorm annotations here. A drop and re-recreate of the database will create these fields because of AutoMigrate.

pkg/database/realm.go
```diff
@@ -299,6 +305,7 @@ func NewRealmWithDefaults(name string) *Realm {
 		CodeDuration:        FromDuration(15 * time.Minute),
 		LongCodeLength:      16,
 		LongCodeDuration:    FromDuration(24 * time.Hour),
+		ShortCodeMaxMinutes: 60,
```
sethvargo, line 308: const?

pkg/database/realm.go
```diff
@@ -384,11 +391,18 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 		r.AddError("passwordWarn", "may not be longer than password rotation period")
 	}
 
+	if r.ShortCodeMaxMinutes < 60 || r.ShortCodeMaxMinutes > 120 {
```
sethvargo, line 394: 60 and 120 should be constants or, more ideally, a runtime configuration as an envvar so we don't have to deploy new code.

sethvargo (changes requested):

assets/server/admin/realms/edit.html
```diff
@@ -43,6 +43,24 @@ <h6 class="mb-3">Name</h6>
             <label for="code">Region code</label>
           </div>
 
+          <h6 class="mb-3">Verification Code Settings</h6>
+          <div class="form-label-group">
+            <input name="short_code_max_minutes" id="short-code-max-minutes" type="text" class="form-control" value="{{$realm.ShortCodeMaxMinutes}}"  />
+            <label for="short-code-max-minutes">Short Code Max Minutes</label>
```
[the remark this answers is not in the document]
mikehelmick: image

pkg/controller/realmadmin/settings_modify.go
```diff
@@ -40,12 +40,14 @@ var (
 )
 
 const (
+	maxShortCodeMinutes = 60
```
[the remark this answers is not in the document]
mikehelmick: This is the default, and is really used just to optimize the create of the slice for the UI `[]int{5..60}`

I'm not sure we want to make this configurable - the other one, the max override limit, I can see making that configurable

pkg/controller/realmadmin/settings_show.go
```diff
@@ -145,7 +145,17 @@ func (c *Controller) renderSettings(
 	m["passwordWarnDays"] = passwordRotationWarningDays
 	// Valid settings for code parameters.
 	m["shortCodeLengths"] = shortCodeLengths
-	m["shortCodeMinutes"] = shortCodeMinutes
+	m["maxShortCodeMinutes"] = maxShortCodeMinutes
+	if realm.ShortCodeMaxMinutes > maxShortCodeMinutes {
```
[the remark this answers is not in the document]
mikehelmick: simplified - the caching was premature optimization anyway

pkg/database/migrations.go
mikehelmick: we've had this exact discussion before :)
can be same TX, but different statement.

sethvargo: Right, different statement is what I meant.

pkg/database/realm.go
mikehelmick: done.

pkg/database/realm.go
mikehelmick: see todo.

mikehelmick:
read for another look

sethvargo (approved):
I'm gonna approve, but see my follow-up comment about the migration.

I'm worried that "SET DEFAULT" won't apply if the column does not exist.

pkg/database/migrations.go
mikehelmick: This works as is for both a new and upgraded DB (tested both)

PR #2050, opened 2021-04-22
Check inactive secrets when comparing TTLs

sethvargo:
This fixes an issue where a secret would be created even if another one already existed but was waiting to be activated.

Also fixes an issue where tests would panic under failure conditions.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/rotation/handle_secrets_rotate.go
+++ b/pkg/controller/rotation/handle_secrets_rotate.go
@@ -194,8 +194,8 @@ func (c *Controller) rotateSecret(ctx context.Context, typ database.SecretType, 
 	// active. That would be at the end of the list.
 	latestSecretCreatedAt := time.Time{}
 	for _, secret := range existing {
-		if secret.CreatedAt.After(latestSecretCreatedAt) {
-			latestSecretCreatedAt = secret.CreatedAt
+		if secret.CreatedAt.UTC().After(latestSecretCreatedAt) {
+			latestSecretCreatedAt = secret.CreatedAt.UTC()
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/rotation/handle_secrets_rotate.go
```diff
@@ -189,9 +189,19 @@ func (c *Controller) rotateSecret(ctx context.Context, typ database.SecretType,
...
+	for _, secret := range existing {
+		if secret.CreatedAt.UTC().After(latestSecretCreatedAt) {
+			latestSecretCreatedAt = secret.CreatedAt.UTC()
+		}
+	}
+
 	// If there is no latest secret, or if the latest secret has existed for
 	// longer than the minimum age, create a new one.
-	if minTTL > 0 && (len(existing) == 0 || now.Sub(existing[0].CreatedAt) > minTTL) {
+	if minTTL > 0 && now.Sub(latestSecretCreatedAt) > minTTL {
```
mikehelmick, line 204: I think this is easier to read if it was

`now.Add(minTTL).After(lastScretCreatedAt)`

but those are the same thing.

mikehelmick: or Before I guess in this case...

mikehelmick: time is reletive

pkg/controller/rotation/handle_secrets_rotate.go
```diff
@@ -215,7 +225,7 @@ func (c *Controller) rotateSecret(ctx context.Context, typ database.SecretType,
 		// Activate any secrets that are ready to be activated. Please don't try to
 		// optimize this. Yes, this could be reduced into a single SQL statement,
 		// but we want to ensure the logging and AuditLog exist for debugging.
-		if !secret.Active && secret.CreatedAt == secret.UpdatedAt && now.Sub(secret.CreatedAt) > c.config.SecretActivationTTL {
+		if !secret.Active && secret.CreatedAt.UTC() == secret.UpdatedAt.UTC() && now.Sub(secret.CreatedAt.UTC()) > c.config.SecretActivationTTL {
```
mikehelmick, line 228: The exact time comparisons can be weird w/ DB values and truncation

mikehelmick (approved):

PR #2056, opened 2021-04-23
Display feature statuses on system admin info page

sethvargo:
**Release Note**

```release-note
Display feature statuses on system admin info page.
```


whaught (approved):
good idea!

mikehelmick (approved):

PR #2057, opened 2021-04-23
adding alamofire and androiddownloadmanager as user agent identifiers…

bschlaman:
## Proposed Changes

* adding "Alamofire" and "AndroidDownloadManager" as identifiers for user-agent OS detection middleware

**Release Note**

```release-note
Adding user agent strings to identify OS for additional common user-agent varieties
```

### DIFF
--- a/pkg/controller/middleware/operating_system.go
+++ b/pkg/controller/middleware/operating_system.go
@@ -25,9 +25,11 @@ import (
 
 func AddOperatingSystemFromUserAgent() mux.MiddlewareFunc {
 	userAgents := map[string]database.OSType{
-		"darwin": database.OSTypeIOS,
-		"iphone": database.OSTypeIOS,
-		"dalvik": database.OSTypeAndroid,
+		"darwin":                 database.OSTypeIOS,
+		"iphone":                 database.OSTypeIOS,
+		"alamofire":              database.OSTypeIOS,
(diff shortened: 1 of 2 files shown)


pkg/controller/middleware/operating_system.go
```diff
@@ -25,9 +25,11 @@ import (
...
 func AddOperatingSystemFromUserAgent() mux.MiddlewareFunc {
 	userAgents := map[string]database.OSType{
-		"darwin": database.OSTypeIOS,
-		"iphone": database.OSTypeIOS,
-		"dalvik": database.OSTypeAndroid,
+		"darwin":                 database.OSTypeIOS,
+		"iphone":                 database.OSTypeIOS,
+		"alamofire":              database.OSTypeIOS,
+		"dalvik":                 database.OSTypeAndroid,
+		"androiddownloadmanager": database.OSTypeAndroid,
```
mikehelmick, line 32: I would be surprised to see this UI making verify calls to the verification server - please confirm

bschlaman: good point, i got the user agent sample from the CDN.  i'll check on the apiserver side and update

bschlaman: I think it might be okhttp that we're looking for, but need to gather some more data or check with the states to confirm.  I'll close this for now pending confirmation

PR #2063, opened 2021-04-30
user report improvements

### LINKED ISSUE #2061: Add agency image and background color to app sync
### TL;DR

The agency image (a png) and background color will be published on the appsync URL from Google.

Consume these values and inject them into the user-report Webview.

mikehelmick:
Fixes #2061
Towards #1928

## Proposed Changes

* Sync agency image from appsync feed
* Display agency image on appropriate background color for user report
* Change `User Report` templates and guidance/restriction to require short codes `[code]`

image

**Release Note**

```release-note
Appsync service will pull agency images if available.
Agency images will show up on the user-report webview
Changed rules on user-report SMS text template
```

### DIFF
--- a/assets/enx-redirect/report/index.html
+++ b/assets/enx-redirect/report/index.html
@@ -12,7 +12,7 @@
   <main role="main" class="container">
     {{if $currentRealm.AgencyImage}}
     <div id="logo" style="background-color: {{$currentRealm.AgencyBackgroundColor}};" >
-      <img src="{{$currentRealm.AgencyImage}}" style="width: 80vw;" />
+      <img src="{{$currentRealm.AgencyImage}}" style="width: 80vw; max-width: 720px;" />
     </div>
     {{end}}
 
--- a/assets/enx-redirect/report/index.html
+++ b/assets/enx-redirect/report/index.html
@@ -10,6 +10,12 @@
 
 <body class="tab-content">
   <main role="main" class="container">
+    {{if $currentRealm.AgencyImage}}
+    <div id="logo" style="background-color: {{$currentRealm.AgencyBackgroundColor}};" >
+      <img src="{{$currentRealm.AgencyImage}}" style="width: 80vw;" />
+    </div>
+    {{end}}
+
     <h1>Request Exposure Notifications verification code from {{.realm.Name}} ({{.realm.RegionCode}})</h1>
 
     <form method="POST" action="/report/issue" class="floating-form">
@@ -53,7 +59,9 @@
                  <label for="agreement" class="form-check-label">
                    I agree that I have received a positive test result for COVID-19, that the phone
                    number provided belongs to me, and that I would like to receive a verification code
-                   for Exposure Notifications to share my information anonymously with others. 
+                   for Exposure Notifications to share my information anonymously with others. I understand
+                   that my phone number may be saved temporarily for anti-abuse purposes and this phone number
+                   will not be shared or used in any other way.
                  </label>              
               </div>
             </div>
--- a/internal/clients/app_sync.go
+++ b/internal/clients/app_sync.go
@@ -60,6 +60,8 @@ type App struct {
 	Region        string `json:"region"`
 	IsEnx         bool   `json:"is_enx,omitempty"`
 	AndroidTarget `json:"android_target"`
+	AgencyColor   string `json:"agencyColor"`
+	AgencyImage   string `json:"agencyImage"`
 }
 
 // AndroidTarget holds the android metadata for an App of AppResponse.
--- a/internal/clients/app_sync.go
+++ b/internal/clients/app_sync.go
@@ -60,8 +60,8 @@ type App struct {
 	Region        string `json:"region"`
 	IsEnx         bool   `json:"is_enx,omitempty"`
 	AndroidTarget `json:"android_target"`
-	AgencyColor   string `json:"agencyColor"`
-	AgencyImage   string `json:"agencyImage"`
+	AgencyColor   string `json:"agency_color"`
+	AgencyImage   string `json:"agency_image"`
 }
 
 // AndroidTarget holds the android metadata for an App of AppResponse.
--- a/pkg/controller/appsync/helpers.go
+++ b/pkg/controller/appsync/helpers.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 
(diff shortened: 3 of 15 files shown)


assets/enx-redirect/report/index.html
```diff
@@ -53,7 +59,9 @@ <h1>Request Exposure Notifications verification code from {{.realm.Name}} ({{.re
                  <label for="agreement" class="form-check-label">
                    I agree that I have received a positive test result for COVID-19, that the phone
                    number provided belongs to me, and that I would like to receive a verification code
-                   for Exposure Notifications to share my information anonymously with others. 
+                   for Exposure Notifications to share my information anonymously with others. I understand
+                   that my phone number may be saved temporarily for anti-abuse purposes and this phone number
```
sethvargo, line 63: I understand that a one-way hash of my phone number... ?

internal/clients/app_sync.go
```diff
@@ -60,6 +60,8 @@ type App struct {
 	Region        string `json:"region"`
 	IsEnx         bool   `json:"is_enx,omitempty"`
 	AndroidTarget `json:"android_target"`
+	AgencyColor   string `json:"agencyColor"`
```
sethvargo, line 63: Ugh are these really camelCase when the other fields are snake case?

pkg/controller/appsync/helpers.go
```diff
@@ -51,6 +56,20 @@ func (c *Controller) syncApps(ctx context.Context, apps *clients.AppsResponse) *
...
+		if _, found := agencyUpdatedAlready[realm.RegionCode]; !found {
+			// Sync the realm level items.
+			realm.AgencyBackgroundColor = app.AgencyColor
+			realm.AgencyImage = app.AgencyImage
+			if err := c.db.SaveRealm(realm, database.System); err != nil {
+				merr = multierror.Append(merr, fmt.Errorf("unable to update agency information: %w", err))
+				continue
+			}
+
+			if app.AgencyImage != "" {
```
sethvargo, line 68: This feels weird to me. What behavior are we trying to achieve when a realm has multiple apps with images? It feels like we should only execute the parent `if` if `AgencyImage != "`

pkg/controller/appsync/helpers.go
```diff
@@ -51,6 +56,20 @@ func (c *Controller) syncApps(ctx context.Context, apps *clients.AppsResponse) *
 			continue
 		}
 
+		if _, found := agencyUpdatedAlready[realm.RegionCode]; !found {
+			// Sync the realm level items.
+			realm.AgencyBackgroundColor = app.AgencyColor
```
sethvargo, line 61: We should add a BeforeSave or sanity check to make sure this is `/\a#[0-9]{6}\Z/`

pkg/database/migrations.go
```diff
@@ -2288,6 +2288,25 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 						DROP COLUMN IF EXISTS enx_code_expiration_configurable`)
 			},
 		},
+		{
+			ID: "00105-AddAgencyData",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS agency_background_color TEXT DEFAULT '#ffffff',
+						ADD COLUMN IF NOT EXISTS agency_image TEXT DEFAULT ''`,
```
sethvargo, line 2297: Why make this default to the empty string instead of NULL? Same with the agency color? I feel like they should be nullable fields.

pkg/database/realm.go
```diff
@@ -165,6 +165,11 @@ type Realm struct {
 	WelcomeMessage    string  `gorm:"-"`
 	WelcomeMessagePtr *string `gorm:"column:welcome_message; type:text;"`
 
+	// AgencyBackgroundColor and AgencyImage are synced from the Google
+	// ENX-Express sync source
+	AgencyBackgroundColor string `gorm:"column:agency_background_color; type:text; not null; default: '#ffffff'"`
```
sethvargo, line 170: I think these should be nullable fields (see WelcomeMessage and WelcomeMessagePtr above)

pkg/database/realm.go
```diff
@@ -449,7 +454,6 @@ func (r *Realm) BeforeSave(tx *gorm.DB) error {
 			if l == DefaultTemplateLabel {
 				r.AddError("smsTextTemplate", fmt.Sprintf("no label for template %s", *t))
 				r.AddError(l, fmt.Sprintf("label %s reserved for the default template", l))
-				continue
```
sethvargo, line 452: ?

sethvargo (changes requested):

assets/enx-redirect/report/index.html
```diff
@@ -10,6 +10,12 @@
 
 <body class="tab-content">
   <main role="main" class="container">
+    {{if $currentRealm.AgencyImage}}
+    <div id="logo" style="background-color: {{$currentRealm.AgencyBackgroundColor}};" >
```
flagxor, line 14: AgencyColor ?

pkg/database/migrations.go
```diff
@@ -2288,6 +2288,25 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 						DROP COLUMN IF EXISTS enx_code_expiration_configurable`)
 			},
 		},
+		{
+			ID: "00105-AddAgencyData",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS agency_background_color TEXT DEFAULT '#ffffff',
```
flagxor, line 2296: agency_color ?

internal/clients/app_sync.go
flagxor: This is being changed to agency_color

internal/clients/app_sync.go
```diff
@@ -60,6 +60,8 @@ type App struct {
...
+	AgencyImage   string `json:"agencyImage"`
```
flagxor, line 64: being changed to agency_image

flagxor (approved):

internal/clients/app_sync.go
mikehelmick: fixed upstream

pkg/controller/appsync/helpers.go
mikehelmick: I was trying to account for "removing" an agency image via sync as well

assets/enx-redirect/report/index.html
sethvargo: I don't think it will confuse the average user, but I think it will be used against us in the press.

internal/clients/app_sync.go
sethvargo: ❤️

pkg/controller/appsync/helpers.go
mikehelmick: added to realm BeforeSave

pkg/database/migrations.go
mikehelmick: made nullable.

pkg/controller/appsync/helpers.go
sethvargo: Is removing an image something we're thinking about? I guess I'd rather handle that as a one-off thing, because I think this logic is confusing

sethvargo (approved):

assets/enx-redirect/report/index.html
mikehelmick: in our model, this is background color

PR #2069, opened 2021-05-06
user report strings update

mikehelmick:
Towards #1928 

## Proposed Changes

* product requested string changes

**Release Note**

```release-note
Update user report webview strings.
```


whaught (approved):
this language does seem simpler...

PR #2081, opened 2021-05-13
Add support for returning generated SMS via the admin API

sethvargo:
This introduces new optional fields to the issue API request and response fields. If `onlyGenerateSMS` is specified in the request, the response will returned the compiled (and signed) SMS message. This feature must be enabled by a system admin on a per-realm basis.

**Release Note**

```release-note
Introduce new optional fields to the `/issue` API for requesting the generated SMS message. This feature must be enabled on a per-realm basis.
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2,20 +2,20 @@
 
 - [API access](#api-access)
 - [API usage](#api-usage)
-    - [Authenticating](#authenticating)
-    - [Error reporting](#error-reporting)
+  - [Authenticating](#authenticating)
+  - [Error reporting](#error-reporting)
 - [API Methods](#api-methods)
-    - [`/api/verify`](#apiverify)
-    - [`/api/certificate`](#apicertificate)
-    - [`/api/user-report`](#apiuser-report)
+  - [`/api/verify`](#apiverify)
+  - [`/api/certificate`](#apicertificate)
+  - [`/api/user-report`](#apiuser-report)
 - [Admin APIs](#admin-apis)
-    - [`/api/issue`](#apiissue)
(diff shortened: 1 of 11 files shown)


mikehelmick:
see test failures + hold for completion of design review

docs/api.md
```diff
@@ -325,6 +326,11 @@ Request a verification code to be issued. Accepts [optional] symptom date and te
     the caller should apply a cryptographic hash before sending that data. **The
     system does not sanitize or encrypt these external IDs, it is the caller's
     responsibility to do so.**
+* `onlyGenerateSMS` is an optional field. If supplied, the system will **not**
```
mikehelmick, line 329: also note that the phone number still needs to be passed in for the authenticated SMS piece.

pkg/controller/issueapi/issue.go
```diff
@@ -112,21 +116,45 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*IssueRequestInte
...
+		// Do not attempt to process things that have already errored.
+		if result.ErrorReturn != nil {
+			continue
+		}
+
+		// Get the associated request for this result.
+		issueReq := requests[i].IssueRequest
+
+		// Do not attempt to process requests that do not have a phone number.
+		if issueReq.Phone == "" {
```
mikehelmick, line 130: I think there's a missing error condition here where the phone number isn't provided BUT `OnlyGenerateSMS` is set to true

sethvargo: All the error cases are [handled in validate_code.go](https://github.com/google/exposure-notifications-verification-server/pull/2081/files#diff-2187f7760dcfc954d7452c4188c16f6c208e47ed70865ebd7bce576d2e7e34faR76), which is called above when the results list is created.

PR #2082, opened 2021-05-14
Add webhook functionality for user report

sethvargo:
**Release Note**

```release-note
Add functionality for PHAs to use custom SMS gateways for user report via webhooks. This feature requires a system admin to enable generated SMS on the realm first. If enabled, PHAs will be able to configure a Webhook URL for user report. When a user submits their diagnoses, the verification server will send a request to the PHAs server, and the PHA can issue the SMS. For more information, please see the [documentation](https://github.com/google/exposure-notifications-verification-server/blob/main/docs/api.md#user-report-webhooks).
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -623,6 +623,7 @@ the compiled SMS message and destination phone number to your server.
 - Must be **secured via TLS** (https)
 - Must accept a **POST** request
 - Must parse the response as JSON
+- Must send a 200 OK response **within 10 seconds**
 
 The request body will be identical to the [API /issue response](#apiissue). Your
 server should parse the JSON body and extract the `generatedSMS` field.
@@ -649,7 +650,8 @@ func acceptPayload(w http.ResponseWriter, r *http.Request) {
     return
   }
 
-  body, err := ioutil.ReadAll(r.Body)
+  lr := io.LimitReader(r.Body, 2_097_152) // 2 MB
+  body, err := ioutil.ReadAll(lr)
   if err != nil {
     w.WriteHeader(500)
     return
--- a/docs/api.md
+++ b/docs/api.md
@@ -16,6 +16,7 @@
   - [`/api/checkcodestatus`](#apicheckcodestatus)
   - [`/api/expirecode`](#apiexpirecode)
   - [`/api/stats/*`](#apistats)
+- [User report webhooks](#user-report-webhooks)
 - [Chaffing requests](#chaffing-requests)
 - [Response codes overview](#response-codes-overview)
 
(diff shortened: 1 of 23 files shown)


docs/api.md
```diff
@@ -611,6 +612,93 @@ endpoints without notice.
...
+func acceptPayload(w http.ResponseWriter, r *http.Request) {
+  defer r.Body.Close()
+
+  sig := w.Header.Get("X-Signature")
+  if sig == "" {
+    w.WriteHeader(400)
+    return
+  }
+
+  body, err := ioutil.ReadAll(r.Body)
```
mikehelmick, line 652: We should show max size read here instead of read all

docs/api.md
```diff
@@ -611,6 +612,93 @@ endpoints without notice.
...
+user completes a self report, the verification server will send a request with
+the compiled SMS message and destination phone number to your server.
+
+- Must be a **publicly-accessible** endpoint
+- Must be **unauthenticated**
+- Must be **secured via TLS** (https)
+- Must accept a **POST** request
+- Must parse the response as JSON
+
+The request body will be identical to the [API /issue response](#apiissue). Your
```
mikehelmick, line 627: It can’t be identical, we need phone number (unless you added that, haven’t gotten to that file yet)

pkg/api/api.go
```diff
@@ -272,6 +272,11 @@ type IssueCodeResponse struct {
 	// request.
 	GeneratedSMS string `json:"generatedSMS,omitempty"`
 
+	// Phone is the phone number in which to send the compiled SMS message. This
```
mikehelmick, line 275: Can we standardize on format, E. 164?

I put in utilities to do this recently.

pkg/controller/issueapi/issue.go
```diff
@@ -60,15 +60,20 @@ func (result *IssueResult) IssueCodeResponse() *api.IssueCodeResponse {
...
 		ExpiresAt:              v.ExpiresAt.Format(time.RFC1123),
 		ExpiresAtTimestamp:     v.ExpiresAt.UTC().Unix(),
 		LongExpiresAt:          v.LongExpiresAt.Format(time.RFC1123),
 		LongExpiresAtTimestamp: v.LongExpiresAt.UTC().Unix(),
-		GeneratedSMS:           result.GeneratedSMS,
 	}
+
+	if result.GeneratedSMS != "" {
+		resp.GeneratedSMS = result.GeneratedSMS
+		resp.Phone = v.PhoneNumber
```
mikehelmick, line 74: Let’s make sure phone number gets formatted consistently

pkg/controller/userreport/controller.go
```diff
@@ -55,11 +57,16 @@ func New(cacher cache.Cacher, cfg *config.RedirectConfig, db *database.Database,
 
 	localCache, _ := memcache.New(30 * time.Second)
 
+	httpClient := &http.Client{
+		Timeout: 10 * time.Second,
```
mikehelmick, line 61: We should document this in the webhook docs - if your processor doesn’t respond in 10 seconds, we will error out to the user.

pkg/controller/userreport/send.go
```diff
@@ -158,6 +171,16 @@ func (c *Controller) HandleSend() http.Handler {
 			}
 		}
 
+		// Compile and send the payload to the webhook URL.
+		if realm.UserReportWebhookURL != "" {
+			if err := sendWebhookRequest(ctx, c.httpClient, realm, result); err != nil {
```
mikehelmick, line 176: Can we add metrics for this?

docs/api.md
sethvargo: Yea, I added it. That field is only returned if OnlyGenerateSMS is supplied though.

docs/api.md
sethvargo: Done

pkg/api/api.go
sethvargo: There's a utility `CanonicalPhoneNumber`, but I don't see it actually used anywhere.

pkg/controller/userreport/controller.go
sethvargo: Done

pkg/api/api.go
sethvargo: The phone is pulled from the verification code, which already uses your standardized thing. I've updated the doc to reflect that it's E.164.

pkg/controller/issueapi/issue.go
sethvargo: See https://github.com/google/exposure-notifications-verification-server/pull/2082#discussion_r633613547

pkg/controller/userreport/send.go
sethvargo: Done

sethvargo:
Added metrics. I'll do a follow-up to create alerts and Terraform the metrics

mikehelmick (approved):

Issue #2102, opened 2021-05-25
 PHA configurable text for user-report webview


mikehelmick:
### TL;DR

Design not complete here, just tracking upcoming work.

1) PHA text + localizations will be configured upstream
2) Google will publish this in the existing appsync feed
3) Add new fields in the database for this
4) if present show on the user report web view

mikehelmick:
still todo, sync webReportLearnMoreURL from appsync feed.

need final JSON field name

PR #2108, opened 2021-05-27
start syncing translations

mikehelmick:
Towards #2102

## Proposed Changes

* add per-realm localization sync method
* attach to app sync job
* refactor data structures to remove circular dependency

**Release Note**

```release-note
Sync and store per-realm localization strings as part of the appsync process.
```

### DIFF
--- a/internal/appsync/app_sync.go
+++ b/internal/appsync/app_sync.go
@@ -29,8 +29,8 @@ type Translation struct {
 
 // Localization represents all known translations of a specific message ID
 type Localization struct {
-	MessageID    string        `json:"msgid"`
-	Translations []Translation `json:"translations"`
+	MessageID    string         `json:"msgid"`
+	Translations []*Translation `json:"translations"`
 }
 
 // App represents single app for the AppResponse body.
@@ -38,10 +38,10 @@ type App struct {
 	Region        string `json:"region"`
 	IsEnx         bool   `json:"is_enx,omitempty"`
 	AndroidTarget `json:"android_target"`
-	AgencyColor   string         `json:"agency_color"`
-	AgencyImage   string         `json:"agency_image"`
-	DefaultLocale string         `json:"default_locale"`
-	Localizations []Localization `json:"localizations"`
+	AgencyColor   string          `json:"agency_color"`
+	AgencyImage   string          `json:"agency_image"`
+	DefaultLocale string          `json:"default_locale"`
+	Localizations []*Localization `json:"localizations"`
 }
 
 // AndroidTarget holds the android metadata for an App of AppResponse.
--- a/internal/appsync/app_sync.go
+++ b/internal/appsync/app_sync.go
@@ -1,0 +1,53 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// package appsync contains API defintions for importing application data
+// from the Google export feed
+package appsync
+
+// AppsResponse is the body for the published list of android apps.
+type AppsResponse struct {
+	Apps []App `json:"apps"`
+}
+
+// Translation is a localized version of a message
+type Translation struct {
+	Language string `json:"lang"`
+	Message  string `json:"message"`
+}
+
+// Localization represents all known translations of a specific message ID
+type Localization struct {
+	MessageID    string        `json:"msgid"`
+	Translations []Translation `json:"translations"`
+}
+
+// App represents single app for the AppResponse body.
+type App struct {
+	Region        string `json:"region"`
+	IsEnx         bool   `json:"is_enx,omitempty"`
+	AndroidTarget `json:"android_target"`
+	AgencyColor   string         `json:"agency_color"`
+	AgencyImage   string         `json:"agency_image"`
+	DefaultLocale string         `json:"default_locale"`
+	Localizations []Localization `json:"localizations"`
+}
+
+// AndroidTarget holds the android metadata for an App of AppResponse.
+type AndroidTarget struct {
+	Namespace              string `json:"namespace"`
+	AppName                string `json:"app_name,omitempty"`
+	PackageName            string `json:"package_name"`
+	SHA256CertFingerprints string `json:"sha256_cert_fingerprints"`
+}
--- a/pkg/controller/appsync/helpers.go
+++ b/pkg/controller/appsync/helpers.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 
-	"github.com/google/exposure-notifications-verification-server/internal/clients"
+	"github.com/google/exposure-notifications-verification-server/internal/appsync"
 	"github.com/google/exposure-notifications-verification-server/internal/project"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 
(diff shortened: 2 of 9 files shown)


internal/appsync/app_sync.go
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// package appsync contains API defintions for importing application data
+// from the Google export feed
+package appsync
```
mikehelmick, line 17: moved to new package to avoid circular dependency

pkg/controller/appsync/helpers.go
```diff
@@ -54,11 +54,19 @@ func (c *Controller) syncApps(ctx context.Context, apps *clients.AppsResponse) *
 
 		realm.AgencyBackgroundColor = strings.ToLower(app.AgencyColor)
 		realm.AgencyImage = app.AgencyImage
+		realm.DefaultLocale = app.DefaultLocale
 		if err := c.db.SaveRealm(realm, database.System); err != nil {
 			merr = multierror.Append(merr, fmt.Errorf("unable to update agency information: %w", err))
 			continue
 		}
 
+		if res, err := c.db.SyncRealmTranslations(realm.ID, app.Localizations); err != nil {
```
sethvargo, line 63: Can we add a test for this in helper_test?

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+
+// DynamicTranslation stores a per-realm localized string that can be used
+// for user-facing content (currently only user-report webview).
+type DynamicTranslation struct {
+	Errorable
+
+	// ID is an auto-increment primary key
+	ID uint
+
+	// RelamID, MessageID, Locale is a unique index on this table.
```
sethvargo, line 34:
```suggestion
	// RealmID, MessageID, Locale is a unique index on this table.
```

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+	return fmt.Sprintf("%d-%s-%s", realmID, locale, msgID)
+}
+
+// LoadDynamicTranslations returns all of the dynamic translations for all realms.
+// The result of this read should be cached for some period of time.
+func (db *Database) LoadDynamicTranslations() ([]*DynamicTranslation, error) {
+	var translations []*DynamicTranslation
+	if err := db.db.
+		Model(&DynamicTranslation{}).
+		Find(&translations).
```
sethvargo, line 65: Order for predictability

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+// this applications localization strategy.
+func localeToLanguage(l string) string {
+	l = strings.ToLower(l)
+	if len(l) <= 2 {
+		return l
+	}
+	return l[0:2]
+}
+
+type TranslatcionSyncResult struct {
```
sethvargo, line 82: Typo

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+	return translationKey(d.RealmID, d.Locale, d.MessageID)
+}
+
+func translationKey(realmID uint, locale string, msgID string) string {
+	return fmt.Sprintf("%d-%s-%s", realmID, locale, msgID)
+}
+
+// LoadDynamicTranslations returns all of the dynamic translations for all realms.
+// The result of this read should be cached for some period of time.
+func (db *Database) LoadDynamicTranslations() ([]*DynamicTranslation, error) {
```
sethvargo, line 61: Can we add a LoadDynamicTranslationsCached helper like we do else where?

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+	return l[0:2]
+}
+
+type TranslatcionSyncResult struct {
+	Added   int
+	Updated int
+	Deleted int
+}
+
+func (db *Database) SyncRealmTranslations(realmID uint, localizations []appsync.Localization) (*TranslatcionSyncResult, error) {
```
sethvargo, line 88: Why not []*appsync.Localization

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+	var translations []*DynamicTranslation
+	if err := db.db.
+		Model(&DynamicTranslation{}).
+		Where("realm_id = ?", realmID).
+		Find(&translations).
+		Error; err != nil {
+		return nil, err
+	}
+
+	for _, t := range translations {
```
sethvargo, line 101: Nit: if you create the map here instead of before the query, you know how big to make it and can reduce allocs.

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+		existingTranslations[t.Key()] = t
+	}
+
+	// Load the incoming translations into a map for ease.
+	// This also ensures any de-duplication.
+	incomingTranslations := make(map[string]string)
```
sethvargo, line 107:
```suggestion
	incomingTranslations := make(map[string]string, len(localizations))
```

pkg/database/dynamic_translation.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, l := range localizations {
+		msgID := l.MessageID
+		for _, t := range l.Translations {
+			key := translationKey(realmID, localeToLanguage(t.Language), msgID)
+			incomingTranslations[key] = t.Message
+		}
+	}
+
+	var errors *multierror.Error = nil
```
sethvargo, line 116:
```suggestion
	var errors *multierror.Error
```

sethvargo: We also usually name this merr

pkg/database/migrations.go
```diff
@@ -2333,6 +2333,32 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
 		},
+		{
+			ID: "00108-AddDynamicTranslations",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS default_locale TEXT`,
+					`CREATE TABLE dynamic_translations (
+						id BIGSERIAL PRIMARY KEY,
+						realm_id INTEGER NOT NULL REFERENCES realms(id),
```
sethvargo, line 2344: ON DELETE CASCADE

We don't currently delete realms, but if we did...

pkg/database/realm.go
```diff
@@ -342,6 +344,7 @@ func NewRealmWithDefaults(name string) *Realm {
 		AllowedTestTypes:    TestTypeConfirmed | TestTypeLikely | TestTypeNegative,
 		CertificateDuration: FromDuration(15 * time.Minute),
 		RequireDate:         true, // Having dates is really important to risk scoring, encourage this by default true.
+		DefaultLocale:       "en",
```
sethvargo, line 347: Const here and below

pkg/controller/appsync/helpers.go
mikehelmick: added tests

pkg/database/dynamic_translation.go
mikehelmick: added, although unnecessary for intended usage (all needs to be put in maps anyway)

pkg/database/dynamic_translation.go
mikehelmick: I'm going to defer that until the next PR where I make use of this data.

pkg/database/dynamic_translation.go
mikehelmick: that's an undercount due to nested structure that is being unrolled. I added a size estiamte.

pkg/database/dynamic_translation.go
mikehelmick: I don't think it matters here, since the array itself is passed as a pointer. Made the change in case other uses would come up.

sethvargo (approved):

PR #2109, opened 2021-05-28
use dynamic translations in user report webview. Show translations in…

mikehelmick:
… admin screens

Towards #2102

## Proposed Changes

* Make use of dynamic translations in the user report UI
* Add middleware to reload and rebuild the translator structs on a 30 minute interval
* Show synced translations for a realm in the system admin page

**Release Note**

```release-note
For Exposure Notifications Express realms, localizations that are synced from the Google feed will be used when rendering the user report web view with sensible fallbacks.
```

### DIFF
--- a/pkg/controller/middleware/i18n.go
+++ b/pkg/controller/middleware/i18n.go
@@ -16,9 +16,15 @@ package middleware
 
 import (
 	"net/http"
+	"sync"
+	"time"
 
+	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/internal/i18n"
+	"github.com/google/exposure-notifications-verification-server/pkg/cache"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
+	"github.com/google/exposure-notifications-verification-server/pkg/database"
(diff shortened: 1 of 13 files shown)


pkg/controller/middleware/i18n.go
```diff
@@ -35,6 +41,69 @@ var rightAlignLanguages = map[string]struct{}{
...
+func LoadDynamicTranslations(locales *i18n.LocaleMap, db *database.Database, cacher cache.Cacher, period time.Duration) mux.MiddlewareFunc {
+	state := &translationReloader{
+		locales: locales,
+		db:      db,
+		cacher:  cacher,
+		period:  period,
+		// default lastUpdate to time.Zero
+	}
+	ctx := context.Background()
+	state.reload(ctx)
```
sethvargo, line 97: Shouldn't this return an error if it fails?

sethvargo (approved):

pkg/controller/middleware/i18n.go
mikehelmick: made it so it returns an error the first time (Will block startup), but then won't fail on subsequent runs

PR #2112, opened 2021-06-01
Make user report web view a realm setting

mikehelmick:
## Proposed Changes

* user report web view is a realm setting now
* remove system feature flag for this

**Release Note**

```release-note
User report web view moves from system setting to realm setting.
```

### DIFF
--- a/pkg/controller/realmadmin/settings_modify.go
+++ b/pkg/controller/realmadmin/settings_modify.go
@@ -64,6 +64,7 @@ type formData struct {
 	Codes                   bool              `form:"codes"`
 	AllowedTestTypes        database.TestType `form:"allowed_test_types"`
 	AllowUserReport         bool              `form:"allow_user_report"`
+	AllowUserReportWebView  bool              `form:"allow_user_report_web_view"`
 	AllowAdminUserReport    bool              `form:"allow_admin_user_report"`
 	UserReportWebhookURL    string            `form:"user_report_webhook_url"`
 	UserReportWebhookSecret string            `form:"user_report_webhook_secret"`
@@ -242,6 +243,7 @@ func (c *Controller) HandleSettings() http.Handler {
 			} else {
 				currentRealm.AllowAdminUserReport = false
 			}
+			currentRealm.AllowUserReportWebView = form.AllowUserReportWebView
 
 			// These fields can only be set if ENX is disabled
 			if !currentRealm.EnableENExpress {
--- a/pkg/controller/realmadmin/settings_modify.go
+++ b/pkg/controller/realmadmin/settings_modify.go
@@ -239,10 +239,10 @@ func (c *Controller) HandleSettings() http.Handler {
 
 			if form.AllowUserReport {
 				currentRealm.AddUserReportToAllowedTestTypes()
-				currentRealm.AllowAdminUserReport = form.AllowAdminUserReport
-			} else {
-				currentRealm.AllowAdminUserReport = false
 			}
+			// These are outside the conditional so that we can provide error feedback to the user for invalid settings combinations.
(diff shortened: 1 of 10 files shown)


pkg/controller/realmadmin/settings_modify.go
```diff
@@ -242,6 +243,7 @@ func (c *Controller) HandleSettings() http.Handler {
 			} else {
 				currentRealm.AllowAdminUserReport = false
 			}
+			currentRealm.AllowUserReportWebView = form.AllowUserReportWebView
```
sethvargo, line 246: Should this be moved up into the "AllowUserReport" block? It doesn't make sense to enable this if UserReport isn't also enabled.

pkg/database/migrations.go
```diff
@@ -2359,6 +2359,26 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
...
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS allow_user_report_web_view BOOLEAN`,
+					`ALTER TABLE realms
+						ALTER COLUMN allow_user_report_web_view SET DEFAULT false`,
+					`ALTER TABLE realms
+						ALTER COLUMN allow_user_report_web_view SET NOT NULL`,
+					// This was missing in a previous migration.
+					`ALTER TABLE realms
```
sethvargo, line 2373: Can you make this 109 by itself and then add 110 with the allow_user_report_web_view (just in case we need to rollback)

pkg/database/realm.go
```diff
@@ -288,6 +288,10 @@ type Realm struct {
 	// value is to allow all test types.
 	AllowedTestTypes TestType `gorm:"type:smallint; not null; default: 14;"`
 
+	// AllowUserReportWebView - if enabled, will use the user report web view
+	// on the redirect server for this realm. If disabled, it will 404.
+	AllowUserReportWebView bool
```
sethvargo, line 293: Still need to set all the `gorm` properties because of automigrate. Also need to do it for `AllowAdminUserReport` (I missed that earlier).

sethvargo, line 293:
```suggestion
	AllowUserReportWebView bool `gorm:"type:boolean; not null; default:false"`
```

pkg/database/realm.go
```diff
@@ -288,6 +288,10 @@ type Realm struct {
...
+
 	// AllowAdminUserReport - is the adminapi:/api/issue allowed to use the user-report
 	// test type if enabled on the realm.
 	AllowAdminUserReport bool
```
sethvargo, line 297:
```suggestion
	AllowAdminUserReport bool `gorm:"type:boolean; not null; default:false"`
```

sethvargo (changes requested):

pkg/controller/realmadmin/settings_modify.go
mikehelmick: by keeping it separate - we can show errors if they try to enable one w/o the other (realm.BeforeSave)

sethvargo: Should we change "AllowAdminUserReport" to behave the same? I don't have a strong opinion on which is the right way, but I think they should be the same

mikehelmick: yeah - I changed them to be the same.

mikehelmick: image

sethvargo (approved):

sethvargo (approved):

PR #2115, opened 2021-06-02
allow different language matchers for static and dynamic translations

mikehelmick:
## Proposed Changes

* Fixes bug where the matcher didn't allow a realm to have a wider language set than the system language set.
* Adds test coverage
* Fixes syntax issue in JA translations

**Release Note**

```release-note
Synced realm translations can be a wider set than the system translations (EN Express only)
```


sethvargo (approved):
There's a linter error, but otherwise LGTM

PR #2122, opened 2021-06-08
Expire all stats after 30 days

sethvargo:
**Release Note**

```release-note
Delete statistics that are more than 30 days old. This retention period can be configured by setting `STATS_MAX_AGE` on the cleanup service. The minimum value is 7 days and the maximum retention period is 60 days. This replaces the existing `KEY_SERVER_STATS_MAX_AGE` variable.
```


pkg/config/cleanup_server_config.go
```diff
@@ -117,6 +120,14 @@ func (c *CleanupConfig) Validate() error {
 			c.VerificationCodeStatusMaxAge.String(), c.VerificationCodeMaxAge.String())
 	}
 
+	// Stats must be valid for at least 7 days, but no more than 60 days.
+	if min := 7; c.StatsMaxAge < time.Duration(min)*24*time.Hour {
```
mikehelmick, line 124: I think the minimum should be 30

mikehelmick (approved):

PR #2124, opened 2021-06-09
Capture time during call to record chaff events

sethvargo:
This mitigates a (very rare) edge case where a chaff request came in near a UTC boundary. In that case, it's possible that the Cloud Run instance throttled CPU to 0 between when the go routine was scheduled and when it executed. Since "now" was relative to when the go routine executed, this could result in incrementing the statistic for the wrong UTC day. In practice, this wouldn't matter because you'd have more than one chaff event per each day.

**Release Note**

```release-note
Fix an issue where a realm chaff event might be recorded for the wrong date.
```


mikehelmick (approved):
raciest of race conditions...

Issue #2121, opened 2021-06-08
Add chaffing support to user-report


sethvargo:

mikehelmick:
I believe this works for the api already, we should discuss needs for the webview

PR #2126, opened 2021-06-09
Add chaff tracking to HTML user-report pages

sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/2121

**Release Note**

```release-note
Add optional chaff tracking to HTML user-report pages.
```

### DIFF
--- a/internal/routes/enx_redirect.go
+++ b/internal/routes/enx_redirect.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
 	"github.com/google/exposure-notifications-verification-server/pkg/ratelimit/limitware"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
+	"github.com/mikehelmick/go-chaff"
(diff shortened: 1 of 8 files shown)


internal/routes/enx_redirect.go
```diff
@@ -155,36 +158,54 @@ func ENXRedirect(
 		// Using a different name, makes it so cookies don't interfer in local dev.
 		requireSession := middleware.RequireNamedSession(sessions, "en-user-report", h)
 
-		{ // handler for /report/issue, required values must be in the established session.
+		// Setup chaffers
+		issueReportTracker := chaff.New()
+		showReportTracker := chaff.New()
```
mikehelmick, line 163: do we expect clients to chaff both requests?

and unless they're pulling all the assets, this doesn't help. it would still be possible to observe real vs chaff

mikehelmick: I'm leaning towards not supporting this, let's discuss offline.

PR #2131, opened 2021-06-12
complete user report strings redo

mikehelmick:
towards #2102 

## Proposed Changes

* Additional PHA customizable and localizable strings on user report page

**Release Note**

```release-note
More customization and localization of user-report web view.
```


pkg/database/migrations.go
```diff
@@ -2396,6 +2396,21 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 						DROP COLUMN IF EXISTS headless`)
 			},
 		},
+		{
+			ID: "00112-AddUserReportLearnMore",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE realms
+						ADD COLUMN IF NOT EXISTS user_report_learn_more TEXT`,
```
sethvargo, line 2404: Nit: this is a link right? It'd be good to name it *Link

sethvargo (approved):
One small nit

pkg/database/migrations.go
mikehelmick: added URL to the field and DB column

PR #2142, opened 2021-06-16
Split assets (again)

sethvargo:
Unrevert the asset split changes. I've deployed this six times now without issue, so I'm not entirely sure what broke the first time.

**Release Note**

```release-note
NONE
```


assets/assets.go
```diff
@@ -23,8 +23,12 @@ import (
 	"github.com/google/exposure-notifications-verification-server/internal/project"
 )
 
-//go:embed server server/**/*
-var serverFS embed.FS
+//go:embed server server/**/* server/**/**/*
```
sethvargo, line 26: This is the only "real" change from the previous one.

mikehelmick (approved):

PR #2146, opened 2021-06-17
Refactor flash javascript and styles for consistency

sethvargo:
**Release Note**

```release-note
Make alerts and notices more prominent in the UI.
```


assets/server/codes/bulk-issue.html
```diff
@@ -24,17 +24,10 @@
     <div class="card mb-3 shadow-sm">
       <div class="card-header">{{t $.locale "codes.bulk-issue.header"}}</div>
       <div class="card-body">
-        {{if not $hasSMSConfig}}
```
whaught, line 27: is there a replacement to this 'why' warning for disabled?

whaught: Just kidding, I see that it's now in the flash.

whaught (approved):

assets/server/codes/bulk-issue.html
sethvargo: Yea, I wanted to try and use the flash to make the style consistent.

PR #2156, opened 2021-06-29
Upgrade to bootstrap 5

sethvargo:
This diff is quite large, but it's mostly adding the new required wrapping for elements.

Benefits of upgrading:

- Keeps us up to date with security releases
- Full support for RTL languages (bootstrap 4 "kinda" works)
- Native floating-forms (although we still need to use our custom implementation for clippy fields for now)
- Better flex positioning and performance

The biggest drawback is that they removed the auto-markup from forms, meaning all forms need to specify a grid and row/col now. That's the majority of this PR. Reviewing without whitespace could be helpful.

**Release Note**

```release-note
Upgrade design framework to latest version.
```


mikehelmick (commented):
looks good except for test failure

mikehelmick (approved):

PR #2163, opened 2021-06-30
android picker region redirects when possible

mikehelmick:
**Release Note**

```release-note
For Android redirects, always hint to the region where possible when redirecting to the picker.
```


flagxor (approved):
lgtm

PR #2168, opened 2021-07-08
user report redesign + additional translations

mikehelmick:
## Proposed Changes

* UI changes for user report web view - rounded corners, floating form fields, label reorder.
* agreement text will not be PHA customizable

(just using CA logo as demonstration, it does not imply endorsement of this feature in any way)

image

**Release Note**

```release-note
User report web view redesign and refresh of default strings and translations.
```


whaught (approved):
I'll assume the translations are accurate

PR #2177, opened 2021-07-21
CSS changes for user report

mikehelmick:
image

**Release Note**

```release-note
CSS Tweaks for user report webview.
```

### DIFF
--- a/assets/enx-redirect/report/header.html
+++ b/assets/enx-redirect/report/header.html
@@ -17,12 +17,24 @@
 {{end}}
 
 <style>
+.ur-title {
+  font-size: calc(1.625rem + 4.5vw) !important;
+  font-weight: 400 !important;
+  line-height: 1.2 !important;
+}
+
+.ur-alert {
+  color: #0D3D17 !important;
+  background-color: #D6EBF0 !important;
+  border-color: #b6d4fe !important;
+}
+
 .rounded-corners {
   border-radius: 20px;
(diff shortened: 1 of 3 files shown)


whaught (approved):

allenx:
1. The alert has border-color, can we remove the border or make it transparent? 
2. It looks like I have some bad calculations on the continue button's corner radius. Can we make it a fixed value which is not based on screen size. The number looked good on my small screen phone, I didn't know it would look this much rounder on bigger screens.

mikehelmick:
I removed the border on the alert and made the button have the same fixed radius as the other elements, `20px`

assets/enx-redirect/report/header.html
```diff
@@ -40,6 +52,23 @@
...
+.continue-btn {
+  display: inline-block;
+  font-weight: 600;
+  line-height: 1.5;
+  color: #ffffff;
+  text-align: center;
+  text-decoration: none;
+  vertical-align: middle;
+  cursor: pointer;
+  -webkit-user-select: none;
```
sethvargo, line 65: You can add a class of `user-select-none` to the element instead of doing this.

sethvargo (commented):
> The number looked good on my small screen phone, I didn't know it would look this much rounder on bigger screens.

Given this will only ever appear on phones via a POST request in a webview, I think that's okay?

PR #2190, opened 2021-08-05
trim leading empty days from composite stats

mikehelmick:
**Release Note**

```release-note
Trim leading empty days from stats charts and exports. This has become more frequent with longer retention and display period.
```


sethvargo (approved):
Tests are unhappy

sethvargo (approved):

Issue #2195, opened 2021-08-18
Migrate to metrics-registrar service


sethvargo:
Same as https://github.com/google/exposure-notifications-server/pull/1549, but for this repo

PR #2199, opened 2021-08-18
Add metrics-registrar service for registering metrics

sethvargo:
This adds a new service that registers metrics on boot. It also removes metrics registration from the other services boot sequences. This service is deployed but inaccessible and purely exists to register metrics.

Fixes https://github.com/google/exposure-notifications-verification-server/issues/2195

**Release Note**

```release-note
**Upgrade notice!** This release introduces a new service which must be tagged and applied. You can run `terraform taint null_resource.build` before running Terraform to build this service. This service is responsible for metric registration and does not receive user traffic.
```


.golangci.yaml
```diff
@@ -56,7 +55,7 @@ linters:
   - 'paralleltest'
   - 'prealloc'
   - 'predeclared'
-  - 'scopelint'
+  - 'revive'
```
sethvargo, line 58: Apparently these are deprecated (was getting warnings running locally)

mikehelmick (approved):

mikehelmick (approved):

mikehelmick (approved):

PR #2206, opened 2021-08-25
Retry Twilio communication on transient network errors

sethvargo:
There are a few failed errors in our logs indicating that sometimes Twilio responds with an EOF. According to a few threads I found online and in GitHub issues, this can occur when the client Keep-Alive isn't honored by the server, but the server doesn't communicate this properly. To mitigate this, this commit does two things:

1. Set req.Close = true. This disables the client from requesting a Keep-Alive.

2. Retry on network-level connectivity issues. Note that we explicitly do NOT retry on HTTP responses, even if they are a non-200. http.Do only returns an error when the underlying transport failed at the network layer, which is explicitly what we want to retry.

**Release Note**

```release-note
Add resiliency against temporary network connectivity issues when sending SMS. 
```

### DIFF
--- a/pkg/sms/twilio.go
+++ b/pkg/sms/twilio.go
@@ -24,6 +24,8 @@ import (
 	"net/url"
 	"strings"
 	"time"
+
+	"github.com/sethvargo/go-retry"
 )
 
 // TwilioMessagingServiceSidPrefix is the prefix for a 34 character messaging service identifier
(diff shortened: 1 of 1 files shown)


sethvargo:
https://github.com/google/exposure-notifications-verification-server/pull/2206/files?w=1 is an easier diff

mikehelmick (approved):

PR #2255, opened 2021-11-02
Unify chart resizing and debouncing, move more javascript into external

sethvargo:
This unifies the chart redrawing logic (on window resize) into a more general-purpose `debounce` function and makes the repainting async so charts aren't redrawn in serial. This also moves some of our inline javascript into the larger javascript bundle for better compression and caching.

**Release Note**

```release-note
Improve performance for redrawing charts when the browser window is resized.
```


assets/server/static/js/application.js
```diff
@@ -513,33 +513,13 @@ function utcDate(str) {
...
-      redrawPending = true;
-      setTimeout(function () {
-        redraw();
-        redrawPending = false;
-      }, timeout);
-    }
+// debounce debounces the given function f for the given ts the provided event.
+// If no other resize events are created within ts, f will fire. Otherwise, f
+// will wait for another ts.
+const debounce = (event, f, ts = 300) => {
```
mikehelmick, line 519: nit: units for ts. I assume it's 300 ms?

mikehelmick (approved):

PR #2260, opened 2021-11-05
support GET when launching user report webview

mikehelmick:
## Proposed Changes

**Release Note**

```release-note
Allows HTTP GET request method (in addition to POST) for initiating the user report webview. The API key and nonce must still be passed as HTTP headers (unless dev mode is also enabled). dev mode should NOT be enabled in production to avoid logging the API key query params.
```

### DIFF
--- a/tools/user-report-web/main.go
+++ b/tools/user-report-web/main.go
@@ -39,6 +39,7 @@ var (
 	timeoutFlag = flag.Duration("timeout", 5*time.Second, "request time out duration in the format: 0h0m0s")
 	apikeyFlag  = flag.String("apikey", "", "API Key to use")
 	addrFlag    = flag.String("addr", "http://localhost:8080", "protocol, address and port on which to make the API call")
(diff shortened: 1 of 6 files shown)


tools/user-report-web/main.go
```diff
@@ -39,6 +39,7 @@ var (
 	timeoutFlag = flag.Duration("timeout", 5*time.Second, "request time out duration in the format: 0h0m0s")
 	apikeyFlag  = flag.String("apikey", "", "API Key to use")
 	addrFlag    = flag.String("addr", "http://localhost:8080", "protocol, address and port on which to make the API call")
+	sendGetFlag = flag.Bool("send-get", false, "If true, will use GET instead of POST to initiate the request")
```
sethvargo, line 42: Nit and can be follow-up. It'd probably be better to have the flag be http-method and accept that as a param on the client (instead of a bool)

sethvargo (approved):

PR #2263, opened 2021-11-08
switch code already expired response to BadRequest instead of an InternalError

bschlaman:
## Proposed Changes

* Switch code already expired or code already claimed scenario to a 4xx response (was previously 5xx).  Currently returning a 400; I think a 410 could also be an appropriate response.
* Fixing some documentation for this API.
* Added check for this response to the end of "success" subtest for code expiration API.  Code and tests for code expiration via form submission is unchanged.

**Release Note**

```release-note
Switch response code for expiring a code via API which is already expired or claimed
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -553,14 +553,27 @@ or
 
 ## `/api/expirecode`
 
-Expires an unclaimed code. If the code has been claimed an error is returned.
+Expires an unclaimed code. If the code has been claimed or is already expired, an error is returned.
 
 **ExpireCodeRequest**
 
 ```json
+{
+  "uuid": "UUID for code to expire",
+  "padding": "<bytes>"
+}
+```
+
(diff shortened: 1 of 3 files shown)


bschlaman:
Note: I chose `api.ErrVerifyCodeInvalid` since it's generic enough to capture both a code being expired or claimed already.  Could also introduce another error type like `api.ErrVerifyCodeNotExpirable` but that seems like overkill

sethvargo:
Hey @bschlaman thanks for the PR. Mike and I spotted something else here (unrelated) that we need to get fixed first. We'll share more in a bit.

That being said, I don't think this is the right approach here. We need to inspect the error (using `errors.Is`) to determine the proper code to return. But hold off on making any changes for now.

bschlaman:
Sounds good, just let me know

sethvargo:
#2264

PR #2265, opened 2021-11-09
Add FK constraints on verification codes

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2452,6 +2452,37 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 				)
 			},
 		},
+		{
+			ID: "00115-AddVerificationCodeConstraints",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx, `
+					ALTER SEQUENCE verification_codes_id_seq AS BIGINT;
+					ALTER TABLE verification_codes ALTER id TYPE BIGINT;
+					ALTER TABLE verification_codes ALTER COLUMN realm_id SET NOT NULL;
+					ALTER TABLE verification_codes ADD CONSTRAINT fk_verification_codes_realms FOREIGN KEY (realm_id) REFERENCES realms(id) ON DELETE CASCADE;
+					ALTER TABLE verification_codes ALTER COLUMN code TYPE TEXT;
+					ALTER TABLE verification_codes ALTER COLUMN long_code TYPE TEXT;
+					ALTER TABLE verification_codes ALTER COLUMN test_type TYPE TEXT;
+					ALTER TABLE verification_codes ALTER COLUMN issuing_external_id TYPE TEXT;
+					ALTER TABLE verification_codes ADD CONSTRAINT fk_issuing_user_id FOREIGN KEY(issuing_user_id) REFERENCES users(id) ON DELETE SET NULL;
+					ALTER TABLE verification_codes ADD CONSTRAINT fk_issuing_app_id FOREIGN KEY(issuing_app_id) REFERENCES authorized_apps(id) ON DELETE SET NULL;
+				`)
+			},
+			Rollback: func(tx *gorm.DB) error {
+				return multiExec(tx, `
+					ALTER SEQUENCE verification_codes_id_seq AS INT;
+					ALTER TABLE verification_codes ALTER id TYPE INT;
+					ALTER TABLE verification_codes ALTER COLUMN realm_id DROP NOT NULL;
+					ALTER TABLE verification_codes DROP CONSTRAINT fk_verification_codes_realms;
+					ALTER TABLE verification_codes ALTER COLUMN code TYPE VARCHAR(512);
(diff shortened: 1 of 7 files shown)


pkg/database/migrations.go
```diff
@@ -2452,6 +2452,37 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 				)
 			},
 		},
+		{
+			ID: "00115-AddVerificationCodeConstraints",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx, `
+					ALTER SEQUENCE verification_codes_id_seq AS BIGINT;
```
mikehelmick, line 2459: any concerns w/ these taking a long time to run?

mikehelmick: this could a long table lock over the verification_codes table.

pkg/database/vercode.go
```diff
@@ -89,29 +91,47 @@ type VerificationCode struct {
...
-	IssuingExternalID string `gorm:"column:issuing_external_id; type:varchar(255);"`
+	IssuingExternalID string `gorm:"column:issuing_external_id; type:text;"`
 }
 
 // BeforeSave is used by callbacks.
 func (v *VerificationCode) BeforeSave(tx *gorm.DB) error {
+	v.IssuingUserIDPtr = uintPtr(v.IssuingUserID)
+	v.IssuingAppIDPtr = uintPtr(v.IssuingAppID)
+
+	if v.RealmID == 0 && v.Realm == nil {
```
mikehelmick, line 115: would this be better to just

if v.Realm == nil {
  error
}
v.RealmID = v.Realm.ID

pkg/database/vercode.go
```diff
@@ -89,29 +91,47 @@ type VerificationCode struct {
...
-		return fmt.Errorf("validation failed: %s", strings.Join(msgs, ", "))
+	return v.ErrorOrNil()
+}
+
+// AfterFind runs after the verification code is found.
+func (v *VerificationCode) AfterFind(tx *gorm.DB) error {
+	v.IssuingUserID = uintValue(v.IssuingUserIDPtr)
+	v.IssuingAppID = uintValue(v.IssuingAppIDPtr)
+
+	if v.Realm == nil {
```
mikehelmick, line 131: do we really need to load the realm for every verification code read?

pkg/database/vercode.go
```diff
@@ -69,12 +68,15 @@ type VerificationCode struct {
...
 
-	RealmID       uint   // VerificationCodes belong to exactly one realm when issued.
-	Code          string `gorm:"type:varchar(512)"`
-	LongCode      string `gorm:"type:varchar(512)"`
-	UUID          string `gorm:"type:uuid;default:null"`
-	Claimed       bool   `gorm:"default:false"`
-	TestType      string `gorm:"type:varchar(20)"`
+	// RealmID is the realm in which this verification code was issued. Codes belong to exactly one realm.
+	RealmID uint `gorm:"column:realm_id; type:integer; not null;"`
+	Realm   *Realm
```
mikehelmick, line 73: I'm not sure the loading of the realm relation is 100% necessary here...

pkg/database/vercode.go
```diff
@@ -204,7 +224,8 @@ func (r *Realm) FindVerificationCode(db *Database, code string) (*VerificationCo
 
 	var vc VerificationCode
 	if err := db.db.
-		Where("realm_id = ? AND (code IN (?) OR long_code IN (?))", r.ID, hmacedCodes, hmacedCodes).
+		Joins("JOIN realms ON realms.id = verification_codes.realm_id").
```
mikehelmick, line 227: per the earlier comment - not loading the realm relation here can avoid a join.

pkg/database/vercode.go
sethvargo: Some places set `Realm` and some places set `RealmID`. We really only care if `RealmID` is set during saving, but this catches both. It's also what we do in MobileApp and a few other models already.

sethvargo: But doesnt matter because removed realm

pkg/database/vercode.go
sethvargo: Done

pkg/database/vercode.go
sethvargo: Done

pkg/database/vercode.go
sethvargo: Done

pkg/database/migrations.go
mikehelmick: discussed & benchmarked offline - all good.

mikehelmick (approved):

pkg/database/migrations.go
sethvargo: Upgrading column types is constant (int -> bigint and varchar -> text), so I'm not concerned there.

Adding the FK constraints will lock for a full table scan for each foreign key. I ran this migration on local dev against a seed run with stats (which created 120k verification codes) and it took ~0.15s.

PR #2266, opened 2021-11-09
Fix splitting for new alerts and aggregate over full modeler interval for CodesClaimedRatioAnomaly

bschlaman:
## Proposed Changes

* Change `alignment_period` for the CodesClaimedRatioAnomaly alert so that counts can be added up over the full duration of the alerting period.  Otherwise, the alert will never fire, since the modeler runs every 4 hrs
* Fix `group_by_fields` to match correct field
* Change `alignment_period` for ElevatedSMSErrors to 5 min

**Release Note**

```release-note
Fix parameters for the CodesClaimedRatioAnomaly and ElevatedSMSErrors alerts
```

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -270,9 +270,9 @@ resource "google_monitoring_alert_policy" "CodesClaimedRatioAnomaly" {
       threshold_value = 1
 
       aggregations {
-        alignment_period     = "60s"
+        alignment_period     = "${8 * local.hour + 10 * local.minute}s"
         per_series_aligner   = "ALIGN_DELTA"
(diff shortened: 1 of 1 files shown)


bschlaman:
This investigation also made me realize that the thresholds for ElevatedSMSErrors might be too low.  I think that `alignment_period` being set to 60s means we require 5 straight minutes of at least 50 errors / min which seems quite high.  Was the intent to alert at 50 errors / 5 min?

sethvargo:
The intent was if we received 50 errors over a 5 min period, yea

bschlaman:
Ok.  I'll update this PR

sethvargo (approved):

PR #2268, opened 2021-11-09
Add FK constraints on verification codes

sethvargo:
Reverting the reverting while including a fix. Check last commit.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/database/migrations.go
+++ b/pkg/database/migrations.go
@@ -2456,29 +2456,14 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 			ID: "00115-NormalizeVerificationCodeReferences",
 			Migrate: func(tx *gorm.DB) error {
 				return multiExec(tx, `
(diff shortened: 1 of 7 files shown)


pkg/database/migrations.go
```diff
@@ -2453,7 +2453,20 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 			},
 		},
 		{
-			ID: "00115-AddVerificationCodeConstraints",
+			ID: "00115-NormalizeVerificationCodeReferences",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx, `
+					DELETE FROM verification_codes WHERE realm_id IS NULL OR realm_id NOT IN (SELECT id FROM realms);
```
mikehelmick, line 2459: having these be in a separate transaction still introduces a race condition where a new row that is not conformant could be added. we may need to roll this out over multiple releases.

thoughts?

mikehelmick (approved):

PR #2270, opened 2021-11-11
allow system admins to purge user report phone numbers

mikehelmick:
## Proposed Changes

* Allow system admins to remove a phone number from deny list prior to expiration.
* If there is a code still associated, expire the code (ads DB index to make more efficient)  
* Removes AND attempted removes generate audit logs

New screen

image

Sample audit logs

image

**Release Note**

```release-note
System administrators can now remove a phone number from the user report deduplication list before the phone number ages out.
```


pkg/controller/admin/user_report.go
```diff
@@ -0,0 +1,79 @@
...
+
+		var form FormData
+		if err := controller.BindForm(w, r, &form); err != nil {
+			flash.Error("Failed to process form: %v", err)
+			c.renderUserReport(ctx, w)
+			return
+		}
+
+		form.PhoneNumber = strings.TrimSpace(form.PhoneNumber)
+		if err := c.db.DeleteUserReport(form.PhoneNumber, currentUser); err != nil {
```
mikehelmick, line 62: I didn't add controller tests because all the logic is in this call.

pkg/database/user_report.go
```diff
@@ -110,13 +120,74 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+		var ur UserReport
+		if err := tx.
+			Set("gorm:query_option", "FOR UPDATE").
+			Model(&UserReport{}).
+			Where("phone_hash IN (?)", hmacedCodes).
+			First(&ur).
+			Error; err != nil {
+			if IsNotFound(err) {
+				// Log an audit record on attempted purges as well.
+				if actor != nil {
```
sethvargo, line 142: I'm not entirely convinced this makes sense. Deleting a phone number is already a highly-privileged operation. I don't think we need to generate an audit log when deleting the record failed due to an invalid phone number. If we're worried about abuse, let's put a rate limit on the number of times this API can be called in a minute/hour.

pkg/database/user_report.go
```diff
@@ -110,13 +120,74 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+			Where("user_report_id = ?", ur.ID).
+			First(vc).
+			Error; err != nil {
+			if IsNotFound(err) {
+				vc = nil
+			} else {
+				return err
+			}
+		}
+		if vc != nil && !vc.IsExpired() {
```
sethvargo, line 171: It might be better to call `realm.ExpireCode` here instead of share the logic.

pkg/database/user_report.go
```diff
@@ -110,13 +120,74 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+		if err := tx.
+			Set("gorm:query_option", "FOR UPDATE").
+			Model(&VerificationCode{}).
+			Where("user_report_id = ?", ur.ID).
+			First(vc).
+			Error; err != nil {
+			if IsNotFound(err) {
+				vc = nil
+			} else {
+				return err
```
sethvargo, line 168: Wrap

pkg/database/user_report_test.go
```diff
@@ -124,21 +124,49 @@ func TestDeleteUserReport(t *testing.T) {
...
+	realm.AddUserReportToAllowedTestTypes()
+	if err := db.SaveRealm(realm, SystemTest); err != nil {
+		t.Fatal(err)
+	}
 
 	phoneNumber := "+8128675309"
-	userReport, err := db.NewUserReport(phoneNumber, generateNonce(t), true)
-	if err != nil {
-		t.Fatalf("error creating user report: %v", err)
+	verCode := VerificationCode{
```
sethvargo, line 137:
```suggestion
	verCode := &VerificationCode{
```

We usually create a pointer in tests instead of taking a pointer later.

pkg/database/user_report_test.go
```diff
@@ -124,21 +124,49 @@ func TestDeleteUserReport(t *testing.T) {
...
-	if err := db.db.Save(userReport).Error; err != nil {
-		t.Fatalf("error writing user report: %v", err)
+
+	if err := realm.SaveVerificationCode(db, &verCode); err != nil {
+		t.Fatalf("unable to save code and user report: %v", err)
 	}
 
-	if err := db.DeleteUserReport(phoneNumber); err != nil {
+	var got *UserReport
+	if err := db.db.Transaction(
```
sethvargo, line 152: Why use a tx here?

pkg/database/user_report.go
```diff
@@ -53,6 +55,14 @@ type UserReport struct {
 	UpdatedAt time.Time
 }
 
+func (ur *UserReport) AuditID() string {
+	return fmt.Sprintf("user_report:%d", ur.ID)
+}
+
+func (ur *UserReport) AuditDisplay() string {
+	return fmt.Sprintf("hash: %q claimed: %v", ur.PhoneHash, ur.CodeClaimed)
```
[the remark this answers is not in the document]
mikehelmick: This was intentional - if a system admin purged a bunch of phone numbers, this is the only way to recover what they did. If we log the row ID, that requires going to backup to reconstruct, and the reconstruct the HMAC.

We get alerted on someone accessing the HMAC secret as well.

I guess we need to prioritize which thing we want to protect against the most.

pkg/database/user_report.go
mikehelmick: but it's not...
the two existing cases 
- user report fails at the SMS stage
- e2e runner

I don't want to log those two cases.

pkg/database/user_report.go
```diff
@@ -110,13 +120,74 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+						PhoneHash:   hmacedCodes[0],
+						CodeClaimed: false,
+					}
+					audit := BuildAuditEntry(actor, "attempted to purge user report phone", &ur, 0)
+					if err := tx.Save(audit).Error; err != nil {
+						return fmt.Errorf("failed to save audits: %w", err)
+					}
+				}
+			}
+			return nil
```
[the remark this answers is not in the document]
mikehelmick: I indented this to "return success" either way.

pkg/database/user_report.go
mikehelmick: it's a different entry point - UUID vs user_report_id, that a substantial refactor. 

I think we want it this way w/ the audit record being associated w/ the user report and not w/ the verification code

pkg/database/user_report_test.go
mikehelmick: because the FindUserReport requires a tx?

pkg/controller/admin/user_report.go
```diff
@@ -0,0 +1,79 @@
...
+
+		currentUser := controller.UserFromContext(ctx)
+		if currentUser == nil {
+			controller.MissingUser(w, r, c.h)
+			return
+		}
+
+		var form FormData
+		if err := controller.BindForm(w, r, &form); err != nil {
+			flash.Error("Failed to process form: %v", err)
```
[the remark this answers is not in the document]
sethvargo: We do it in quite a few places 😄 : https://github.com/google/exposure-notifications-verification-server/blob/ed4cbf5aab6ffbdf1e715f6432aa39ea4c58e32e/pkg/controller/user/create.go#L60

pkg/database/user_report.go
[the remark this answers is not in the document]
sethvargo: If an admin did that, I think we would probably care more about the timestamp of the audit and just restore all phone numbers from a backup (ignoring conflicts that already exist). It wouldn't be necessary to access the HMAC key or insert specific phone numbers. I think we're be accessing the backup either way, but logging the row ID makes it less likely to leak the HMAC secret.

Also, keep in mind that the HMAC secret rotates over time, so there's no guarantee that, if we discovered this after a rotation, that the _current_ HMAC secret is the value that the phone number was written as (you'd have to check all HMAC keys), which would be rather toilsome.

pkg/database/user_report.go
sethvargo: I'm suggesting we _don't_ generate an audit when deleting a user report if the user report is not found. I'm advocating we delete L142-153 entirely. The audit log should really only contain actions that mutated data (and this would not have mutated data since the data didn't exist).

pkg/database/user_report.go
```diff
@@ -110,13 +120,75 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
-	return db.db.Where("phone_hash IN (?)", hmacedCodes).Delete(&UserReport{}).Error
+	return db.db.Transaction(func(tx *gorm.DB) error {
+		var ur UserReport
+		if err := tx.
+			Set("gorm:query_option", "FOR UPDATE").
+			Model(&UserReport{}).
+			Where("phone_hash IN (?)", hmacedCodes).
+			First(&ur).
+			Error; err != nil {
+			if IsNotFound(err) {
```
sethvargo, line 140: We tend to just return a wrapped error here and let the caller handle whether it's `NotFound` or `ValidationError` in the controller (since it could indicate a different response).

pkg/database/user_report.go
```diff
@@ -110,13 +120,75 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+		if err := tx.
+			Set("gorm:query_option", "FOR UPDATE").
+			Model(&VerificationCode{}).
+			Where("user_report_id = ?", ur.ID).
+			First(vc).
+			Error; err != nil {
+			if IsNotFound(err) {
+				vc = nil
+			} else {
+				return fmt.Errorf("unable to search for related verification codes: %w", err)
```
sethvargo, line 169:
```suggestion
				return fmt.Errorf("failed to find related verification codes: %w", err)
```

pkg/database/user_report.go
```diff
@@ -110,13 +120,75 @@ func (db *Database) FindUserReport(tx *gorm.DB, phoneNumber string) (*UserReport
...
+			}
+		}
+		if vc != nil && !vc.IsExpired() {
+			// The associated code needs to be expired so that
+			// the code can't be used by anyone else.
+			exp := time.Now().UTC()
+			vc.ExpiresAt = exp
+			vc.LongExpiresAt = exp
+			if err := tx.Save(vc).Error; err != nil {
+				return fmt.Errorf("unable to expire code: %w", err)
```
sethvargo, line 179:
```suggestion
				return fmt.Errorf("failed to expire code: %w", err)
```

pkg/controller/e2erunner/handlers.go
```diff
@@ -57,7 +57,7 @@ func (c *Controller) handleEndToEnd(cfg *config.E2ERunnerConfig, m *stats.Int64M
 		logger := logging.FromContext(ctx)
 
 		if cfg.DoUserReport {
-			if err := c.db.DeleteUserReport(project.TestPhoneNumber); err != nil {
+			if err := c.db.DeleteUserReport(project.TestPhoneNumber, nil); err != nil {
```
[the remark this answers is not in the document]
sethvargo: Okay fair enough. I think I'd still like to _require_ `actor` to always be non-nil. If we want to create a new Actor that we can directly compare against, that'd be better.

```go
// NullActor is an actor to use as an Auditable when an audit entry should not be generated.
var NullActor Auditable = new(nullActor)

type nullActor struct{}

func (n *nullActor) AuditID() string {
	return ""
}

func (n *nullActor) AuditDisplay() string {
	return ""
}

// IsNullActor returns true if the given Auditable is the null actor.
func IsNullActor(a Auditable) bool {
  _, ok := a.(*nullActor)
  return ok
}
```

pkg/controller/admin/user_report.go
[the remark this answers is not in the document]
mikehelmick: most of the BindForm locales do not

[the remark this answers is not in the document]
mikehelmick: added here, don't want to get all of the other sites in this PR

pkg/database/user_report.go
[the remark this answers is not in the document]
mikehelmick: right, only good for a certain period of time.

before I change to row ID.

This would prevent recovering the phone numbers O(time between backups)

pkg/database/user_report.go
mikehelmick: we should discuss offline. This is to catch an admin attempting to delete a bunch of phone numbers even if they weren't there.

pkg/database/user_report.go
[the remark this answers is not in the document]
sethvargo: Talked live - gonna take the first 6-8 characters for display.

```
abcd1234 (claimed: false)
```

sethvargo (approved):

Issue #2273, opened 2021-11-11
change missing actor to common error


mikehelmick:
### TL;DR

right now it's a fmt.Errof w/ different messages
this should be a common error instance

PR #2277, opened 2021-11-11
Return a common error when Auditable actor is missing

sethvargo:
**Release Note**

```release-note
NONE
```


sethvargo:
Fixes https://github.com/google/exposure-notifications-verification-server/issues/2273

mikehelmick (approved):

PR #2281, opened 2021-11-15
Add configuration for number of stdevs to be considered anomaly

sethvargo:
**Release Note**

```release-note
Add a global configuration option (applies to all realms) to configure the number of standard deviations away from the norm before behavior is consider an anomaly. You can configure this value by setting `ANOMALY_ALLOWED_STDEVS` on the `server` and `modeler` components to any positive float value. The default value is 2.0.
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -109,6 +109,23 @@ var (
 
 	ENXRedirectDomain = os.Getenv("ENX_REDIRECT_DOMAIN")
 
+	AnomalyAllowedStdevs = func() float64 {
+		v := os.Getenv("ANOMALY_ALLOWED_STDEVS")
+		if v == "" {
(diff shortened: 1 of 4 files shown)


pkg/database/realm.go
```diff
@@ -109,6 +109,23 @@ var (
 
 	ENXRedirectDomain = os.Getenv("ENX_REDIRECT_DOMAIN")
 
+	AnomalyAllowedStdevs = func() float64 {
+		v := os.Getenv("ANOMALY_ALLOWED_STDEVS")
```
mikehelmick, line 113: if you add this to one of the config structs, I think startup will fail in a more graceful way (probably need to leave these here too)

mikehelmick (approved):

Issue #2283, opened 2021-11-15
Code claim ratio anomaly detection should factor in overall issue volume


bschlaman:
### TL;DR
Currently, per-realm UTC day code claim ratios are considered anomalous if they stray >=2 standard deviations from a rolling 14 day mean.  Realms with low overall issue rates have a naturally greater degree of statistical uncertainty in their daily claim ratios, and are likely to trigger false positives in the `codes_claimed_ratio_anomaly` metric.

Additionally, I think it's sensible to omit the default e2e test realm from this metric, as it has a comparatively high issue rate, but will trigger the alert typically with even a single failure, as near 100% success is expected.  While it is important to be alerted on this, this behavior is a duplicate of the `ForwardProgress-e2e-default` alert.

### Design

**Proposal**
I propose there be an inverse relationship between the average code issue rate of a realm and how many consecutive anomaly days are required to trigger the alert.  (Perhaps the scalar constant can be a configurable value?  Or set dynamically based on the average realm issue rate across the server?)  To simplify, maybe each realm would fall into 1 of 2 categories - high and low issue realms.

The downsides here would be a potential delay in alerting low-issue realms of an outage.

**Alternatives considered**
Alternatively, you could consider creating an inverse relationship between average code issue rate and number of standard deviations required to increment the metric; e.g. for a low-issue realm, it may require 3 standard deviations instead of the default 2.  This is probably much easier to code but may not eliminate all false-positives.

bschlaman:
@sethvargo

sethvargo:
> Additionally, I think it's sensible to omit the default e2e test realm from this metric, as it has a comparatively high issue rate, but will trigger the alert typically with even a single failure, as near 100% success is expected. While it is important to be alerted on this, this behavior is a duplicate of the ForwardProgress-e2e-default alert.

This is done and will be included in the next release (https://github.com/google/exposure-notifications-verification-server/pull/2284).

sethvargo:
Two other alternatives (open to ideas):

1. Allow per-realm configurable thresholds. This adds additional overhead to realm admins to configure their desired threshold (much like the abuse factor), but it also gives realms that want to be alerted on the slightest anomalies the ability to configure that as such.

1. Change the calculation such that it's over _total_ codes claimed / codes_issued for the 14-day window (instead of per day). Right now, the calculation is:

    ```text
    Mean = Sum(ratio.day1, ratio.day2, ratio.day3, ...) / N
    ```

[...]

PR #2284, opened 2021-11-15
Ignore e2e realm in anomaly detection notifications

sethvargo:
Alternatively, I could inline this in `realm.CodesClaimedRatioAnomalous` so the e2e realm would never be considered anomalous. Open to either.

**Release Note**

```release-note
Ignore e2e realm in anomaly notifications.
```

### DIFF
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -453,6 +453,12 @@ func (db *Database) E2ERealm() (*Realm, error) {
 	return &realm, nil
 }
 
+// IsE2ERealm returns true if this realm is the "e2e" realm, false otherwise.
+func (r *Realm) IsE2ERealm() bool {
+	return strings.HasPrefix(strings.ToUpper(r.RegionCode), "E2E-") ||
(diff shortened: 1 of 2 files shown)


sethvargo:
Part of https://github.com/google/exposure-notifications-verification-server/issues/2283

pkg/database/realm.go
```diff
@@ -453,6 +453,12 @@ func (db *Database) E2ERealm() (*Realm, error) {
 	return &realm, nil
 }
 
+// IsE2ERealm returns true if this realm is the "e2e" realm, false otherwise.
+func (r *Realm) IsE2ERealm() bool {
+	return strings.HasPrefix(strings.ToUpper(r.RegionCode), "E2E-") ||
+		strings.HasPrefix(strings.ToLower(r.Name), "e2e-test-")
```
mikehelmick, line 459: we should probably prevent creation of realms w/ this name in the UI? no?

sethvargo: Yea - I can do that in a follow-up

mikehelmick (approved):

PR #2294, opened 2021-12-06
Use a custom transport on all clients

sethvargo:
Even if this doesn't _fix_ the issue, I think it's a reasonable change. My suspicion is that, since `http.DefaultTransport` is a global variable, another package is modifying the value of the dial timeout to be very low, and we are inheriting it everywhere. I haven't been able to _prove_ that, but that's my theory as to why we're seeing timeouts for such small times. 

This introduces a new project-level helper, `DefaultHTTPTransport` which returns a pointer to a new allocation of an `http.Transport` on each invocation with no shared state. Even if this doesn't end up being the root cause, it will prevent the situation I described above from ever arising.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/userreport/send_test.go
+++ b/pkg/controller/userreport/send_test.go
@@ -80,7 +80,8 @@ func TestSendWebhookRequest(t *testing.T) {
 	ctx := project.TestContext(t)
 
 	client := &http.Client{
-		Timeout: 2 * time.Second,
+		Timeout:   2 * time.Second,
+		Transport: project.DefaultHTTPTransport(),
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/controller/userreport/send_test.go
```diff
@@ -80,7 +80,8 @@ func TestSendWebhookRequest(t *testing.T) {
 	ctx := project.TestContext(t)
 
 	client := &http.Client{
-		Timeout: 2 * time.Second,
+		Timeout:   2 * time.Second,
```
sethvargo, line 83: We should probably bump this to 5s to match other timeouts, but wanted to limit changes in this PR

mikehelmick (approved):

PR #2296, opened 2021-12-06
Extract GCP trace ID onto context and inject in clients

sethvargo:
**Release Note**

```release-note
Improve tracing of requests that span multiple services (like the e2e runer)
```

### DIFF
--- a/pkg/controller/context.go
+++ b/pkg/controller/context.go
@@ -32,16 +32,17 @@ type contextKey string
 const (
 	contextKeyAuthorizedApp = contextKey("authorizedApp")
 	contextKeyFirebaseUser  = contextKey("firebaseUser")
+	contextKeyLocale        = contextKey("locale")
 	contextKeyMembership    = contextKey("membership")
 	contextKeyMemberships   = contextKey("memberships")
+	contextKeyNonce         = contextKey("nonce")
+	contextKeyOS            = contextKey("os")
 	contextKeyRealm         = contextKey("realm")
 	contextKeyRequestID     = contextKey("requestID")
 	contextKeySession       = contextKey("session")
 	contextKeyTemplate      = contextKey("template")
+	contextKeyTraceID       = contextKey("traceID")
 	contextKeyUser          = contextKey("user")
-	contextKeyOS            = contextKey("os")
-	contextKeyNonce         = contextKey("nonce")
-	contextKeyLocale        = contextKey("locale")
 )
 
 // WithOperatingSystem stores the operating system enum in the context.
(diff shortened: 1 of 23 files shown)


pkg/controller/context.go
```diff
@@ -32,16 +32,17 @@ type contextKey string
...
+	contextKeyOS            = contextKey("os")
 	contextKeyRealm         = contextKey("realm")
 	contextKeyRequestID     = contextKey("requestID")
 	contextKeySession       = contextKey("session")
 	contextKeyTemplate      = contextKey("template")
+	contextKeyTraceID       = contextKey("traceID")
 	contextKeyUser          = contextKey("user")
-	contextKeyOS            = contextKey("os")
-	contextKeyNonce         = contextKey("nonce")
-	contextKeyLocale        = contextKey("locale")
```
sethvargo, line 44: Sorry, my editor auto-sorted them

pkg/controller/userreport/controller.go
```diff
@@ -61,10 +60,7 @@ func New(locales *i18n.LocaleMap, cacher cache.Cacher, cfg *config.RedirectConfi
 
 	localCache, _ := memcache.New(30 * time.Second)
 
-	httpClient := &http.Client{
-		Timeout:   10 * time.Second,
-		Transport: project.DefaultHTTPTransport(),
-	}
+	httpClient := controller.TracedHTTPClient(10 * time.Second)
```
mikehelmick, line 63: This will propagate the trace ID to external clients (used for webhook)

Do we want to do that?

pkg/controller/middleware/logger.go
```diff
@@ -61,12 +57,9 @@ func PopulateLogger(originalLogger *zap.SugaredLogger) mux.MiddlewareFunc {
 			}
 
 			// On Google Cloud, extract the trace context and add it to the logger.
-			if v := r.Header.Get(googleCloudTraceHeader); v != "" && googleCloudProjectID != "" {
-				parts := strings.Split(v, "/")
-				if len(parts) > 0 && len(parts[0]) > 0 {
-					val := fmt.Sprintf("projects/%s/traces/%s", googleCloudProjectID, parts[0])
-					logger = logger.With(googleCloudTraceKey, val)
-				}
+			if id := controller.TraceIDFromContext(ctx); id != "" && googleCloudProjectID != "" {
```
mikehelmick, line 60: This introduces an undocumented middleware ordering issue...

pkg/controller/userreport/controller.go
sethvargo: I was interested in getting the request ID, but I realize that might be a bad idea now. Reverting.

pkg/controller/middleware/logger.go
sethvargo: We already had one (for request ID generation). Documented both.

mikehelmick (approved):

PR #2303, opened 2022-01-04
read header color instead of agency color

mikehelmick:
**Release Note**

```release-note
For user-report webview, use agency background color instead of agency color.
```


sethvargo (approved):
LGTM, disabled auto-merge since we're waiting for the internal CL

PR #2306, opened 2022-01-06
Allow ignoring Twilio codes from alerts

sethvargo:
```release-note
Allow server operators to ignore specific Twilio error codes in SMS alerts. Server operators can set the `ignored_twilio_error_codes` variable in the alerting module to a string list of error codes that should be ignored when determined if there are elevated SMS errors for a realm. This can be helpful for Twilio errors that are non-actionable, such as "30006" which means a message was sent to a landline phone number. This is a global configuration that applies to all realms on the server; there is no realm-specific configuration. This change affects the monitoring and alerting policy, not the data collection. Realms will still see all SMS errors in their statistics dashboard, even if that error code is excluded from the alerting policy.
```

### DIFF
--- a/docs/playbooks/alerts/ElevatedSMSErrors.md
+++ b/docs/playbooks/alerts/ElevatedSMSErrors.md
@@ -1,9 +1,12 @@
 # Elevated SMS Errors Alert
 
-This alert fires when one or more realms have an increased number of errors returned from the Twilio API.
+This alert fires when one or more realms have an increased number of errors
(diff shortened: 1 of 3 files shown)


sethvargo:
@bschlaman FYI - this is the approach we're taking for ignoring specific Twilio error codes.

sethvargo:
@mikehelmick ready for review - tested in dev and the regex seems to work.

mikehelmick (approved):

PR #2307, opened 2022-01-07
Fix NPE when queue is full

sethvargo:
**Release Note**

```release-note
Fix an issue where the server might crash when a large number of codes were submitted to the SMS provider and the SMS provider rejects the request due to a queueing issue.
```


mikehelmick (approved):

sethvargo:
@mikehelmick PTAL - I added a test for this specific issue.

whaught (approved):

PR #2312, opened 2022-01-12
allow user report to use different SMS queue

mikehelmick:
## Proposed Changes

* Address queuing issues we've seen as case #s have risen
* Allow for user report to use a different phone number

**Release Note**

```release-note
Allow user report to utilize a different "from" number for SMS sending. This can help reduce queuing and self report codes being blocked by a large bulk issue in a jurisdiction.
```

### DIFF
--- a/assets/server/realmadmin/_form_sms.html
+++ b/assets/server/realmadmin/_form_sms.html
@@ -73,7 +73,21 @@
           reduce the chance that your message will be flagged as spam.
         </small>
       </div>
+
+      {{if $realm.AllowsUserReport}}
+      <div class="form-floating mb-3">
+        <input type="tel" name="twilio_user_report_from_number" id="twilio-user-report-from-number" class="form-control font-monospace {{invalidIf ($smsConfig.ErrorsFor "twilioUserReportFromNumber")}}"
+          placeholder="Twilio user report number" value="{{$smsConfig.TwilioUserReportFromNumber}}">
+        <label for="twilio-user-report-from-number">Twilio user report number</label>
+        {{template "errorable" $smsConfig.ErrorsFor "twilioUserReportFromNumber"}}
+        <small class="form-text text-muted">
+          This is the Twilio From Number only for <strong>User Report</strong>. If present, user report codes will be
+          sent on a different queue from regular and bulk issue codes.
+        </small>
+      </div>
+      {{end}}
     </div>
+
 
     <div class="sms-system-form collapse{{if $realm.UseSystemSMSConfig}} show{{end}}">
       <div class="form-floating mb-3">
--- a/assets/server/realmadmin/_form_sms.html
+++ b/assets/server/realmadmin/_form_sms.html
@@ -82,7 +82,8 @@
         {{template "errorable" $smsConfig.ErrorsFor "twilioUserReportFromNumber"}}
         <small class="form-text text-muted">
           This is the Twilio From Number only for <strong>User Report</strong>. If present, user report codes will be
-          sent on a different queue from regular and bulk issue codes.
+          sent on a different queue from regular and bulk issue codes. If unspecified, user report will use the same
+          queue and phone number as regular and bulk issue codes.
         </small>
       </div>
       {{end}}
--- a/pkg/controller/issueapi/issue.go
+++ b/pkg/controller/issueapi/issue.go
@@ -106,14 +106,14 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*IssueRequestInte
 	defer c.recordStats(ctx, results)
 
 	// Send SMS messages if there's an SMS provider.
-	smsProvider, err := c.smsProviderFor(ctx, realm, false)
+	smsProvider, err := c.smsProviderFor(ctx, realm)
 	if err != nil {
 		logger.Errorw("failed to get sms provider", "error", err)
 		errorAll(results, api.InternalError())
 		return results
 	}
 	// If there is no user report SMS provider, it will be the same as the regular provider.
-	smsProviderUserReport, err := c.smsProviderFor(ctx, realm, true)
+	smsProviderUserReport, err := c.smsProviderFor(ctx, realm, &database.SMSProviderUserReport{})
 	if err != nil {
 		logger.Errorw("failed to get user report sms provider", "error", err)
 		errorAll(results, api.InternalError())
@@ -162,11 +162,11 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*IssueRequestInte
 			wg.Add(1)
 			go func(request *api.IssueCodeRequest, r *IssueResult) {
 				defer wg.Done()
+				provider := smsProvider
 				if request.TestType == api.TestTypeUserReport {
-					c.SendSMS(ctx, realm, smsProviderUserReport, smsSigner, keyID, request, r)
-					return
+					provider = smsProviderUserReport
 				}
-				c.SendSMS(ctx, realm, smsProvider, smsSigner, keyID, request, r)
+				c.SendSMS(ctx, realm, provider, smsSigner, keyID, request, r)
 			}(issueReq, result)
 		}
 	}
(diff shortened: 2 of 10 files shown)


mikehelmick:
🛑 not done testing yet, don't merge yet.

sethvargo (commented):
In my head, I was thinking we create an entirely separate SMS Provider for self-report. That would, theoretically, allow for using a different Twilio _account_, not just the from number.

assets/server/realmadmin/_form_sms.html
```diff
@@ -73,8 +73,22 @@ <h5 class="mb-3">Credentials</h5>
...
+
+      {{if $realm.AllowsUserReport}}
+      <div class="form-floating mb-3">
+        <input type="tel" name="twilio_user_report_from_number" id="twilio-user-report-from-number" class="form-control font-monospace {{invalidIf ($smsConfig.ErrorsFor "twilioUserReportFromNumber")}}"
+          placeholder="Twilio user report number" value="{{$smsConfig.TwilioUserReportFromNumber}}">
+        <label for="twilio-user-report-from-number">Twilio user report number</label>
+        {{template "errorable" $smsConfig.ErrorsFor "twilioUserReportFromNumber"}}
+        <small class="form-text text-muted">
+          This is the Twilio From Number only for <strong>User Report</strong>. If present, user report codes will be
+          sent on a different queue from regular and bulk issue codes.
```
sethvargo, line 85: "If unspecified, user-report will use the same queue and phone number as regular and bulk issue codes."

pkg/controller/issueapi/issue.go
```diff
@@ -106,12 +106,19 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*IssueRequestInte
 	defer c.recordStats(ctx, results)
 
 	// Send SMS messages if there's an SMS provider.
-	smsProvider, err := c.smsProviderFor(ctx, realm)
+	smsProvider, err := c.smsProviderFor(ctx, realm, false)
```
sethvargo, line 109: I'm not a huge fan of arbitrary booleans that control behavior. Can we make a helper off of Realm like `smsUserReportProviderFor` instead? The arbitrary true/false is going to cause a bug somewhere.

pkg/controller/issueapi/issue.go
```diff
@@ -155,6 +162,10 @@ func (c *Controller) IssueMany(ctx context.Context, requests []*IssueRequestInte
 			wg.Add(1)
 			go func(request *api.IssueCodeRequest, r *IssueResult) {
 				defer wg.Done()
+				if request.TestType == api.TestTypeUserReport {
```
sethvargo, line 165: Nit, I think it'd be clearer if we did something like:

```go
defer wg.Done()

provider := smsProvider
if request.TestType == api.TestTypeUserReport {
  provider = smsProviderUserReport
}

c.SendSMS(ctx, realm, provider, smsSigner, keyID, request, r)
```

pkg/controller/issueapi/issue.go
```diff
@@ -177,10 +188,10 @@ func (c *Controller) recordStats(ctx context.Context, results []*IssueResult) {
 
 // smsProviderFor returns the sms provider for the given realm. It pulls the
 // value from a local in-memory cache.
-func (c *Controller) smsProviderFor(ctx context.Context, realm *database.Realm) (sms.Provider, error) {
-	key := fmt.Sprintf("realm:%d:sms_provider", realm.ID)
+func (c *Controller) smsProviderFor(ctx context.Context, realm *database.Realm, userReport bool) (sms.Provider, error) {
+	key := fmt.Sprintf("realm:%d:sms_provider:user_report:%v", realm.ID, userReport)
```
sethvargo, line 192: One of the advantages of making user report a separate sms provider is that we can keep this logic the same everywhere, since it'd have a unique ID.

pkg/controller/issueapi/validate_code.go
```diff
@@ -109,7 +109,7 @@ func (c *Controller) BuildVerificationCode(ctx context.Context, internalRequest
 	var smsProvider sms.Provider
 	if !request.OnlyGenerateSMS && request.Phone != "" {
 		var err error
-		smsProvider, err = realm.SMSProvider(c.db)
+		smsProvider, err = realm.SMSProvider(c.db, vCode.IsUserReport())
```
sethvargo, line 112: I don't think we care _which_ sms provider we get back here, just that we got one.

pkg/database/migrations.go
```diff
@@ -2521,6 +2521,19 @@ func (db *Database) Migrations(ctx context.Context) []*gormigrate.Migration {
 				)
 			},
 		},
+		{
+			ID: "00119-UserReportFromNumber",
+			Migrate: func(tx *gorm.DB) error {
+				return multiExec(tx,
+					`ALTER TABLE sms_configs ADD COLUMN IF NOT EXISTS twilio_user_report_from_number VARCHAR(255) NOT NULL DEFAULT ''`,
```
sethvargo, line 2528:
```suggestion
					`ALTER TABLE sms_configs ADD COLUMN IF NOT EXISTS twilio_user_report_from_number TEXT NOT NULL DEFAULT ''`,
```

sethvargo, line 2528: I also think it _can_ be null and we should probably allow that.

pkg/database/sms_config.go
```diff
@@ -40,6 +40,11 @@ type SMSConfig struct {
 	// Twilio messaging service identifier see: https://support.twilio.com/hc/en-us/articles/223134387-What-is-a-Message-SID-
 	TwilioFromNumber string `gorm:"type:varchar(255)"`
 
+	// E.164 format telephone number or
+	// Twilio messaging service identifier see: https://support.twilio.com/hc/en-us/articles/223134387-What-is-a-Message-SID-
+	// If present, used for user report only.
+	TwilioUserReportFromNumber string `gorm:"type:varchar(255)"`
```
sethvargo, line 46:
```suggestion
	TwilioUserReportFromNumber string `gorm:"type:TEXT"`
```

pkg/database/vercode.go
```diff
@@ -44,13 +44,17 @@ const (
 	CodeTypeLong
 )
 
+const (
```
sethvargo, line 47: We should probably make all the test types a public const.

pkg/database/vercode.go
```diff
@@ -137,6 +141,10 @@ func (v *VerificationCode) FormatSymptomDate() string {
 	return v.SymptomDate.Format(project.RFC3339Date)
 }
 
+func (v *VerificationCode) IsUserReport() bool {
```
sethvargo, line 144: I think it'd be better to make the consts public

pkg/controller/issueapi/issue.go
mikehelmick: replaced w/ Options pattern. bool was for speedy POC & testing.

pkg/controller/issueapi/issue.go
mikehelmick: I think we agreed not to do that, right?

I've made this contingent on the options applied and the order they appear in - best future proofing I can think of ATM.

pkg/database/migrations.go
mikehelmick: the existing twilio_from_number is still varchar(255), so I'm being consistent

pkg/controller/issueapi/issue.go
sethvargo: Yea, agreed.

pkg/database/sms_config_test.go
```diff
@@ -221,7 +221,7 @@ func TestSMSProvider(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	provider, err := realm.SMSProvider(db)
+	provider, err := realm.SMSProvider(db, false)
```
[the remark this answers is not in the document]
mikehelmick: yeah...

pkg/database/vercode.go
mikehelmick: separate PR ok?

sethvargo (approved):

PR #2313, opened 2022-01-13
Fix small emailer issues

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -35,49 +35,53 @@ locals {
     ]
   ))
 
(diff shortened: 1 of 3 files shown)


terraform/alerting/alerts.tf
```diff
@@ -35,49 +35,53 @@ locals {
     ]
   ))
 
-  forward_progress_indicators = merge({
```
sethvargo, line 38: Review this with https://github.com/google/exposure-notifications-verification-server/pull/2313/files?w=1

PR #2320, opened 2022-01-14
Add support for cc and bcc on system emails

sethvargo:
**Release Note**

```release-note
Add support for setting CC and BCC on all system-sent emails. To configure this, see the production guide for sending system emails.
```


sethvargo:
We should probably deploy this and test it. I deleted my test env, and it'd be faster just to deploy in dev and test it out.

mikehelmick (approved):

PR #2321, opened 2022-01-15
Send a single email to multiple recipients

sethvargo:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -390,8 +390,8 @@ The verification server is capable of sending email through a Google Workspace S
 
       service_environment = {
         emailer = {
-          EMAIL_CC  = "cc@example.com"
-          EMAIL_BCC = "bcc@example.com"
+          CC_ADDRESSES  = "cc@example.com"
+          BCC_ADDRESSES = "bcc@example.com"
         }
       }
     }
--- a/pkg/controller/emailer/emailer.go
+++ b/pkg/controller/emailer/emailer.go
@@ -46,10 +46,14 @@ func New(cfg *config.EmailerConfig, db *database.Database, h *render.Renderer) *
 }
 
 // sendMail sends a single message through the Google Workspace SMTP relay.
-func (c *Controller) sendMail(to string, msg []byte) error {
-	toAddr, err := mail.ParseAddress(to)
-	if err != nil {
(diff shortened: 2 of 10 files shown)


docs/production.md
```diff
@@ -390,8 +390,8 @@ The verification server is capable of sending email through a Google Workspace S
 
       service_environment = {
         emailer = {
-          EMAIL_CC  = "cc@example.com"
-          EMAIL_BCC = "bcc@example.com"
+          CC_ADDRESSES  = "cc@example.com"
```
sethvargo, line 393: This isn't really "breaking" since we never cut a release

pkg/controller/emailer/emailer.go
```diff
@@ -46,10 +46,14 @@ func New(cfg *config.EmailerConfig, db *database.Database, h *render.Renderer) *
...
-func (c *Controller) sendMail(to string, msg []byte) error {
-	toAddr, err := mail.ParseAddress(to)
-	if err != nil {
-		return fmt.Errorf("failed to parse to address %q: %w", to, err)
+func (c *Controller) sendMail(tos []string, msg []byte) error {
+	toAddrs := make([]*mail.Address, 0, len(tos))
+	for _, to := range tos {
+		toAddr, err := mail.ParseAddress(to)
+		if err != nil {
+			return fmt.Errorf("failed to parse to address %q: %w", to, err)
```
sethvargo, line 54: Thoughts? Should I just log here and continue along? If we don't return an error, I'm not sure how'd we'd be alerted of an invalid email. FWIW, emails here are also parsed in a callback when added to the realm, so _theoretically_ this should never fail.

mikehelmick: we could even include failed email addresses in the email for an admin to fix up

sethvargo: We can't include the failed emails in the email because we won't know they've failed until after we've sent. We'd need to store the results somewhere and then alert on the _next_ email.

mikehelmick: not failed - I meant invalid, ones that fail to parse.

mikehelmick (approved):

PR #2336, opened 2022-03-02
Moar cookies

### LINKED ISSUE #2335: Session cookie too big
There are some scenarios where the session cookie can become too large and fail to persist correctly.

This occurs with particularly long email addresses when creating an API key

mikehelmick:
Fixes #2335

## Proposed Changes

* for storage purposes only - split the auth cookie into a separate session so that the main session cookie (Contains flash for example) doesn't exceed 4k bytes
* This is handled in the require session middleware by joining the value pre-processing and splitting the value post-processing (just before save)

Before this change:

A sufficiently long email address creates a large session cookie:
image

And creating a new API key fails

image

Live upgrading that user, by splitting the cookies

image

Successfully creates an API key and shows it

image

**Release Note**

```release-note
Fixes an issue where some users might not see a newly created API key.
```

### DIFF
--- a/pkg/controller/middleware/sessions.go
+++ b/pkg/controller/middleware/sessions.go
@@ -20,8 +20,10 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/google/exposure-notifications-verification-server/internal/auth"
 	"github.com/google/exposure-notifications-verification-server/pkg/controller"
 	"github.com/google/exposure-notifications-verification-server/pkg/render"
+	"github.com/hashicorp/go-multierror"
 
 	"github.com/gorilla/mux"
 	"github.com/gorilla/sessions"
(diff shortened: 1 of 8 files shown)


pkg/controller/middleware/sessions.go
```diff
@@ -30,24 +32,25 @@ import (
 
 const (
 	// sessionName is the name of the session.
-	sessionName = "verification-server-session"
+	sessionName     = "verification-server-session"
```
sethvargo, line 35: Since these are included in the max length, should we pick something shorter like `_vs` and `_vsu`? That would force a logout on deploy, but I'm not super worried about that.

pkg/controller/middleware/sessions.go
```diff
@@ -65,6 +68,17 @@ func buildHandler(store sessions.Store, name string, h *render.Renderer) func(ht
...
 				session, _ = store.New(r, name)
 			}
+			var authSession *sessions.Session
+			if authProvider != nil {
+				authSession, err = store.Get(r, authSessionName)
+				if err != nil {
+					logger.Errorw("failed to get auth session", "error", err)
+					authSession, _ = store.New(r, name)
+				}
+				if err = authProvider.JoinAuthCookie(session, authSession); err != nil {
```
sethvargo, line 78: I think you can do the cookie splitting and joining here, and avoid needing to hang this off of the `auth.Provider` entirely:

On save:

- Extract and remove fields from session.Values and put into a new session that has the same store. These are public fields so we can easily write a Clone function or something.
- Save both sessions (or 3 or 4 if we wanted in the future)

On load:

- Download all cookies and merge into a single session.Values map.

sethvargo (changes requested):

pkg/controller/middleware/sessions.go
mikehelmick: I left it as is so that logged in users don't get logged out.

pkg/controller/middleware/sessions.go
mikehelmick: splitting all values is unnecessary overhead for network and computation. I think it's best to do this surgical fix.

alternatively - I could have the session provider take in a list of keys that should be split to own cookies.

how does that sound?

sethvargo: Sorry - I didn't mean _every_ key. We know that the firebase key is a problem. We can explicitly pull that one out into its own cookie and explicitly merge it back in.

sethvargo: TL;DR - I'd prefer to keep this out of the `authProvider` if possible.

mikehelmick: simplified.

sethvargo (approved):

PR #2348, opened 2022-05-17
Secret agent man

mikehelmick:
## Proposed Changes

* accept macintosh desktop user agents for redirect

**Release Note**

```release-note
Accept desktop user agents from Macs. iPhone where the browser is set to "always request desktop" mode enabled, currently get an error if they tap on the link in the SMS.
```


sethvargo (approved):
I think this is fine? I guess worst case scenario, someone on desktop could be redirected inaccurately.

PR #2354, opened 2022-07-12
add charting for bad nonce / device mismatch

mikehelmick:
## Proposed Changes

* Adds chart for bad nonce / device mismatch
* Adds corresponding help
* Fixes bug introduced in previous commit

Added line to chart

image

Added help text

image

**Release Note**

```release-note
Adds a chart on the realm statistics page that indicates when a user-report code was requested from one device, but sent to another (and therefore not able to be claimed).
```

### DIFF
--- a/pkg/database/token.go
+++ b/pkg/database/token.go
@@ -271,7 +271,6 @@ func (db *Database) VerifyCodeAndIssueToken(request *IssueTokenRequest) (*Token,
 		// Check associated UserReport record if necessary.
 		if vc.UserReportID != nil {
 			providedNonce := base64.StdEncoding.EncodeToString(request.Nonce)
-			required := providedNonce != ""
 
 			var ur UserReport
 			if err := tx.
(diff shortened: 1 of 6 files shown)


pkg/database/token.go
```diff
@@ -271,7 +271,6 @@ func (db *Database) VerifyCodeAndIssueToken(request *IssueTokenRequest) (*Token,
 		// Check associated UserReport record if necessary.
 		if vc.UserReportID != nil {
 			providedNonce := base64.StdEncoding.EncodeToString(request.Nonce)
-			required := providedNonce != ""
```
mikehelmick, line 274: this was the bug - in the previous version it worked because it was all enforced in SQL. Now we need to check it after the query.

tools/seed/main.go
```diff
@@ -444,7 +463,7 @@ func generateCodesAndStats(ctx context.Context, db *database.Database, realm *da
 					IssuingExternalID: issuingExternalID,
 				}
 				if isUserReport {
-					phoneNumber++
+					phoneNumber := phoneNumberGen.next()
```
mikehelmick, line 466: I don't know why this didn't see contention before, but we need to sync phone number issue...

sethvargo (approved):

Issue #2357, opened 2022-07-13
Upgrade to Firebase 9


sethvargo:
Firebase 8 is deprecated and no longer supported.

PR #2359, opened 2022-07-13
Update CSS/JS elements to their latest versions

sethvargo:
The new version of firebase and firebase-auth require a significant refactoring of our Javascript code; we would need to migrate to modules and use a bundler like Webpack. Fortunately there's a compat layer we can use for now to minimize code changes, but this is something we'll need to think about long term.

Fixes GH-2357

**Release Note**

```release-note
Update Javascript and CSS elements to bootstrap to 5.2.0 and Firebase to 9.9.0.
```

### DIFF
--- a/assets/enx-redirect/header.html
+++ b/assets/enx-redirect/header.html
@@ -14,12 +14,12 @@
 <meta name="msapplication-config" content="/static/browserconfig.xml?{{.buildID}}">
 <meta name="theme-color" content="#ffffff">
 
-{{if eq $.textDirection "rtl"}}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.rtl.min.css"
-  integrity="sha384-gXt9imSW0VcJVHezoNQsP+TNrjYXoGcrqBZJpry9zJt8PCQjobwmhMGaDHTASo9N" crossorigin="anonymous">
+{{if eq (printf "%s" $.textDirection) "rtl"}}
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/5.2.0-beta1/css/bootstrap.rtl.min.css"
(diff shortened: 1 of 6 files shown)


assets/enx-redirect/header.html
```diff
@@ -14,12 +14,12 @@
 <meta name="msapplication-config" content="/static/browserconfig.xml?{{.buildID}}">
 <meta name="theme-color" content="#ffffff">
 
-{{if eq $.textDirection "rtl"}}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.rtl.min.css"
-  integrity="sha384-gXt9imSW0VcJVHezoNQsP+TNrjYXoGcrqBZJpry9zJt8PCQjobwmhMGaDHTASo9N" crossorigin="anonymous">
+{{if eq (printf "%s" $.textDirection) "rtl"}}
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/5.2.0-beta1/css/bootstrap.rtl.min.css"
```
sethvargo, line 18: Yes, it's "beta", but it's the currently recommended version.

assets/enx-redirect/header.html
```diff
@@ -14,12 +14,12 @@
...
+  integrity="sha512-Swm+W9nAnhHjRr94CSIoZSWqt3SQfImX8IoMjn01dLoYiEgeSSVBuS+GqVupSA5nSfGZrUfUkb8rilBv4WQ6bQ==" crossorigin="anonymous" referrerpolicy="no-referrer" />
```
sethvargo, line 19: Referrer policy is a new security recommendation.

pkg/render/assets.go
```diff
@@ -30,13 +30,13 @@ import (
 	"github.com/google/exposure-notifications-verification-server/internal/buildinfo"
 )
 
-const sriPrefix = "sha384-"
```
sethvargo, line 33: 384 is deprecated.

mikehelmick (approved):

Issue #2358, opened 2022-07-13
Modeler tests have different results on ARM


mikehelmick:
### TL;DR

One of the modeler tests has a different results on ARM (Apple M1) vs Intel processors.

```
--- FAIL: TestRebuildAbusePreventionModel (4.95s)
    modeler_test.go:160: expected 21 to be 22
```

**Expected behavior**

Math works the same everywhere

**Observed behavior**

Math doesn't work the same everywhere

### Reproduction

go test ./...

**Environment**

Mac OS, M1 Pro processor

sethvargo:
Interestingly, this returns 21 even in Rosetta (ARM operating in Intel mode): 

```text
arch && go test -run TestRebuildAbusePreventionModel ./pkg/controller/modeler/...
i386

--- FAIL: TestRebuildAbusePreventionModel (0.68s)
    modeler_test.go:160: expected 21 to be 22
FAIL
FAIL	github.com/google/exposure-notifications-verification-server/pkg/controller/modeler	5.919s
FAIL
```

sethvargo:
https://github.com/gonum/gonum/issues/1812

PR #2361, opened 2022-07-13
Round modeler results

sethvargo:
Due to floating point math differences between Intel and Arm, we want to round the modeler result before storing it. This is mostly for tests, since in practice the value is +-0.000000000001.

Fixes GH-2358

**Release Note**

```release-note
Update modeler code to handle floating point math differences between Intel and Arm processors.
```

### DIFF
--- a/pkg/controller/modeler/modeler.go
+++ b/pkg/controller/modeler/modeler.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/gonum/matrix/mat64"
 	"github.com/google/exposure-notifications-server/pkg/logging"
 	"github.com/google/exposure-notifications-verification-server/pkg/config"
 	"github.com/google/exposure-notifications-verification-server/pkg/database"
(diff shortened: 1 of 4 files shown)


pkg/controller/modeler/modeler.go
```diff
@@ -173,13 +173,20 @@ func (c *Controller) rebuildAbusePreventionModel(ctx context.Context, realm *dat
 	// potentially less than zero. We need to do the negative check against the
 	// float value before casting to a uint, or else risk overflowing if this
 	// value is negative.
-	nextFloat := math.Ceil(curve(float64(len(ys))))
+	raw := curve(float64(len(ys)))
+	logger.Debugw("computed raw next curve", "next", raw)
```
mikehelmick, line 177: do you want this to stay in?

sethvargo: Yes. If it was in, I would have found this a lot sooner because I would have seen `21.00000000000004`

PR #2362, opened 2022-07-14
remove unnecessary code and increase test coverage

mikehelmick:
## Proposed Changes

* Remove unnecessary key size calculation

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/jwthelper/jwthelper.go
+++ b/pkg/jwthelper/jwthelper.go
@@ -29,6 +29,9 @@ import (
 	"github.com/golang-jwt/jwt"
 )
 
+// keyBytes is 256 bits / 32 bytes.
+const keyBytes = 32
+
 // TODO(mikehelmick): Add ValidateJWT, accepting a serialized token + signer.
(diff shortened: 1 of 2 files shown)


pkg/jwthelper/jwthelper.go
```diff
@@ -63,11 +63,6 @@ func SignJWT(token *jwt.Token, signer crypto.Signer) (string, error) {
 		return "", fmt.Errorf("unable to unmarshal signature: %w", err)
 	}
 
-	keyBytes := 256 / 8
-	if 256%8 > 0 {
```
mikehelmick, line 67: this calculation is unnecessary when we always use 256 key size.

sethvargo:
After some digging, I found that this likely came from https://github.com/someone1/gcp-jwt-go/blob/d78d05e25456d531f06afd7da65bc69ddfed4fbb/kms.go#L238-L259, which handles multiple ECDSA curves. Since we only handle P256, this is probably fine.

sethvargo (approved):

PR #2367, opened 2022-08-05
User report cleanup

mikehelmick:
## Proposed Changes

* Fix user report cleanup

**Release Note**

```release-note
Fix a potential race condition where the system could be configured to clean up user report records (delete nonce) while the user report code is still valid. This ensures that user reports across all realms are not cleaned up sooner than the realm with the longest short code timeout AND user report enabled.
```

### DIFF
--- a/pkg/controller/cleanup/handle_cleanup.go
+++ b/pkg/controller/cleanup/handle_cleanup.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
 	enobs "github.com/google/exposure-notifications-server/pkg/observability"
-	"github.com/google/exposure-notifications-verification-server/pkg/pagination"
 	"github.com/hashicorp/go-multierror"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/tag"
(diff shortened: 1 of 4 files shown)


pkg/controller/cleanup/handle_cleanup.go
```diff
@@ -266,7 +267,27 @@ func (c *Controller) HandleCleanup() http.Handler {
 		func() {
 			defer enobs.RecordLatency(ctx, time.Now(), mLatencyMs, &result, &item)
 			item = tag.Upsert(itemTagKey, "UNCLAIMED_USER_REPORTS")
-			if count, err := c.db.PurgeUnclaimedUserReports(c.config.UserReportUnclaimedMaxAge); err != nil {
+
+			realms, _, err := c.db.ListRealms(pagination.UnlimitedResults)
```
sethvargo, line 271: Instead of looping over all the realms, can we just `SELECT MAX(user_report_unclaimed_max_age)`?

sethvargo: Sorry, I meant `MAX(realm.short_code_max_minutes)`

mikehelmick: updated

pkg/database/realm.go
```diff
@@ -1294,6 +1294,25 @@ func (r *Realm) ValidTestType(typ string) bool {
 	}
 }
 
+func (db *Database) MaximumUserReportTimeout() (time.Duration, error) {
+	// TestTypeUSerReport == 16 == 0b10000
```
sethvargo, line 1298:
```suggestion
	// TestTypeUserReport == 16 == 0b10000
```

pkg/database/realm.go
```diff
@@ -1294,6 +1294,25 @@ func (r *Realm) ValidTestType(typ string) bool {
...
+	sql := `
+	SELECT
+		MAX(code_duration) AS duration
+	FROM realms
+	WHERE
+	    allowed_test_types >= 16;`
+
+	var result struct {
+		Quantity int64 `gorm:"column:duration;"`
```
sethvargo, line 1307: Quantity?

sethvargo (approved):

pkg/database/realm.go
mikehelmick: copy paste...

PR #2370, opened 2022-08-17
Don't allow longexpires in user report

### LINKED ISSUE #2369: SMS Validation and long expires
### TL;DR

The SMS validation for the `User Report` template should only allow the short expiry time to appear in the message and block the long expiry time from appearing in the message.

I haven't reproduced this yet, but the code appears to allow either.

mikehelmick:
Fixes #2369

## Proposed Changes

* don't allow `[longexpires]` in the `User Report` sms template (since it always uses the short expires)
* add visual feedback to the UI for this.

image

**Release Note**

```release-note
Stop allowing `longexpires` in the User Report SMS template since it always uses the `expires` value.
```

### DIFF
--- a/assets/server/realmadmin/_form_sms.html
+++ b/assets/server/realmadmin/_form_sms.html
@@ -444,7 +444,7 @@
         val = val.replace(/\[longcode\]/g, longCode);
       }
 
-      isUserReport = $(target).attr('label') == "User Report";
+      isUserReport = $(target).attr('label') === "User Report";
       if (isUserReport) {
         if (val.includes(("[longexpires]"))) {
           errors.push("'User Report' template cannot contain `[longexpires]` since user report always uses the short expiration time `[expires]`, which is always in minutes.");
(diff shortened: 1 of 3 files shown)


assets/server/realmadmin/_form_sms.html
```diff
@@ -438,11 +438,19 @@ <h5 class="mb-3">SMS templates</h5>
...
         if (!(hasSC || hasLC) || (hasSC && hasLC)) {
           errors.push('must contain exactly one of `[code]` or `[longcode]`');
-        }
+        }        
 
         val = val.replace(/\[region\]/g, region);
         val = val.replace(/\[longcode\]/g, longCode);
       }
+
+      isUserReport = $(target).attr('label') == "User Report";
```
sethvargo, line 447:
```suggestion
      isUserReport = $(target).attr('label') === "User Report";
```

pkg/database/realm.go
```diff
@@ -718,6 +718,13 @@ func (r *Realm) validateSMSTemplate(label, t string) string {
 		}
 	}
 
+	if label == UserReportTemplateLabel {
```
sethvargo, line 721: Can you add a test for this too please?

assets/server/realmadmin/_form_sms.html
mikehelmick: Javascript!!!!!

sethvargo (approved):

PR #2374, opened 2022-09-12
No e2 e stats

mikehelmick:
## Proposed Changes

* Exclude E2E-TEST realm from system code issue stats

**Release Note**

```release-note
Note to operators, this will exclude a realm with a region code of E2E-TEST from system wide code issue/code claimed stats. If you have named this realm something else, you will need to override the EXCLUDE_FROM_SYSTEM_STATS environment variable.
```

### DIFF
--- a/pkg/config/server_config.go
+++ b/pkg/config/server_config.go
@@ -172,7 +172,7 @@ func (c *ServerConfig) Validate() error {
 		return fmt.Errorf("failed to validate issue API configuration: %w", err)
 	}
 
-	if c.MinRealmsForSystemStatistics < 1 {
+	if c.MinRealmsForSystemStatistics < 2 {
 		return fmt.Errorf("MIN_REALMS_FOR_SYSTEM_STATS cannot be set lower than 2")
 	}
 
--- a/pkg/config/server_config.go
+++ b/pkg/config/server_config.go
@@ -103,6 +103,9 @@ type ServerConfig struct {
 	// MinRealmsForSystemStatistics gives a minimum threshold for displaying system
 	// admin level statistics
 	MinRealmsForSystemStatistics uint `env:"MIN_REALMS_FOR_SYSTEM_STATS, default=2"`
+	// ExcludeFromSystemStatistics is the realm location to exclude from the system stats.
+	// This is the e2e testing realm by default.
+	ExcludeFromSystemStatistics string `env:"EXCLUDE_FROM_SYSTEM_STATS, default=E2E-TEST"`
 
 	// Rate limiting configuration
 	RateLimit ratelimit.Config
--- a/pkg/config/server_config.go
+++ b/pkg/config/server_config.go
@@ -103,9 +103,9 @@ type ServerConfig struct {
 	// MinRealmsForSystemStatistics gives a minimum threshold for displaying system
 	// admin level statistics
 	MinRealmsForSystemStatistics uint `env:"MIN_REALMS_FOR_SYSTEM_STATS, default=2"`
-	// ExcludeFromSystemStatistics is the realm location to exclude from the system stats.
-	// This is the e2e testing realm by default.
-	ExcludeFromSystemStatistics string `env:"EXCLUDE_FROM_SYSTEM_STATS, default=E2E-TEST"`
+	// ExcludeFromSystemStatistics is a list of realm IDs to exclude from system stats, in addition
+	// to any e2e test realms that are found.
+	ExcludeFromSystemStatistics []uint `env:"EXCLUDE_FROM_SYSTEM_STATS"`
 
 	// Rate limiting configuration
 	RateLimit ratelimit.Config
@@ -172,7 +172,7 @@ func (c *ServerConfig) Validate() error {
 		return fmt.Errorf("failed to validate issue API configuration: %w", err)
 	}
 
-	if c.MinRealmsForSystemStatistics < 2 {
+	if c.MinRealmsForSystemStatistics < 1 {
 		return fmt.Errorf("MIN_REALMS_FOR_SYSTEM_STATS cannot be set lower than 2")
 	}
 
--- a/pkg/database/realm.go
+++ b/pkg/database/realm.go
@@ -1323,16 +1323,16 @@ func (db *Database) MaximumUserReportTimeout() (time.Duration, error) {
 	return timeout, nil
 }
 
-func (db *Database) FindE2ETestRealms() ([]*Realm, error) {
-	var realms []*Realm
+func (db *Database) FindE2ETestRealm() (*Realm, error) {
+	var realm Realm
 	if err := db.db.
 		Model(&Realm{}).
 		Where("is_e2e = ?", true).
-		First(&realms).
+		First(&realm).
 		Error; err != nil {
 		return nil, err
 	}
-	return realms, nil
+	return &realm, nil
 }
 
 func (db *Database) FindRealmByRegion(region string) (*Realm, error) {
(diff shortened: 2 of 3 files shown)


pkg/config/server_config.go
```diff
@@ -103,6 +103,9 @@ type ServerConfig struct {
 	// MinRealmsForSystemStatistics gives a minimum threshold for displaying system
 	// admin level statistics
 	MinRealmsForSystemStatistics uint `env:"MIN_REALMS_FOR_SYSTEM_STATS, default=2"`
+	// ExcludeFromSystemStatistics is the realm location to exclude from the system stats.
+	// This is the e2e testing realm by default.
+	ExcludeFromSystemStatistics string `env:"EXCLUDE_FROM_SYSTEM_STATS, default=E2E-TEST"`
```
sethvargo, line 108: Is this realm names or realm IDs? It should really be IDs to be stable.

mikehelmick: updated to realm IDs

pkg/config/server_config.go
```diff
@@ -172,7 +172,7 @@ func (c *ServerConfig) Validate() error {
 		return fmt.Errorf("failed to validate issue API configuration: %w", err)
 	}
 
-	if c.MinRealmsForSystemStatistics < 2 {
+	if c.MinRealmsForSystemStatistics < 1 {
```
sethvargo, line 175: Need to update the error. You can cheat and inline with:

```go
if min := 1; c.MinRealmsForSystemStatistics < min {
  return fmt.Errorf("MIN_REALMS_FOR_SYSTEM_STATS cannot be set lower than %d", min)
}
```

pkg/database/realm.go
```diff
@@ -1323,6 +1323,18 @@ func (db *Database) MaximumUserReportTimeout() (time.Duration, error) {
 	return timeout, nil
 }
 
+func (db *Database) FindE2ETestRealms() ([]*Realm, error) {
```
sethvargo, line 1326: Nit: Should be `ListE2ETestRealms()`. Also, there can only be one e2e test realm (database constraint), and I thought we already had a function to find it.

pkg/database/realm.go
```diff
@@ -1947,17 +1959,26 @@ func (db *Database) AllRealmCodeStatsCached(ctx context.Context, cacher cache.Ca
 // This reuses the RealmStats data structure and the realm_id is the COUNT
 // of all realms that contributed to statistics on that day, so that filtering
 // can be done if there are not enough data points.
-func (db *Database) AllRealmCodeStats(ctx context.Context, requiredRealms int, excludeReam string) (RealmStats, error) {
+func (db *Database) AllRealmCodeStats(ctx context.Context, requiredRealms int, excludeRealms []uint) (RealmStats, error) {
```
sethvargo, line 1962: Nit: excludeRealmIDs or something to indicate the variable better

pkg/database/realm.go
```diff
@@ -1947,17 +1959,26 @@ func (db *Database) AllRealmCodeStatsCached(ctx context.Context, cacher cache.Ca
...
-		db.logger.Warnw("unable to find realm to exclude from system stats", "excludeRealm", excludeReam, "error", err)
-		realm = nil
+		db.logger.Warnf("unable o find e2e test realms for exclusion from system stats", "error", err)
+	} else {
+		for _, r := range e2eTestRealms {
+			excludedRealmIDs = append(excludedRealmIDs, r.ID)
+		}
+	}
+	// Exclude other explicitly listed realms
+	for _, rID := range excludeRealms {
```
sethvargo:
```go
excludeRealmIds = append(excludeRealmIDs, excludeRealms...)
```

pkg/database/realm.go
```diff
@@ -1981,17 +2002,17 @@ func (db *Database) AllRealmCodeStats(ctx context.Context, requiredRealms int, e
 		) d
 		LEFT JOIN realm_stats s ON s.date = d.date
 		`
-	if realm != nil {
+	if len(excludedRealmIDs) > 0 {
 		// This needs the newlines in it
-		sql = sql + `WHERE s.realm_id != $3
+		sql = sql + `WHERE NOT s.realm_id = ANY($3)
```
sethvargo, line 2005: I think you might want `NOT IN` instead of `ANY`? I forget which is more performant.

pkg/database/realm.go
```diff
@@ -1981,17 +2002,17 @@ func (db *Database) AllRealmCodeStats(ctx context.Context, requiredRealms int, e
 		) d
 		LEFT JOIN realm_stats s ON s.date = d.date
 		`
-	if realm != nil {
+	if len(excludedRealmIDs) > 0 {
```
sethvargo, line 2003: This should always be appended to the query, and the values appended to the values list. SQL will optimize away a `NOT INT []` call to be a noop search.

pkg/config/server_config.go
mikehelmick: debug code that didn't get reverted

pkg/database/realm.go
mikehelmick: `WHERE s.realm_id NOT IN $3` gives a sql error

When $3 is an empty array - which could very well happen

```
{"severity":"DEBUG","timestamp":"2022-09-12T18:11:52.690161-07:00","logger":"database","message":"SELECT d.date AS date, COUNT(DISTINCT(realm_id)) as realm_id, SUM(s.codes_issued) AS codes_issued, SUM(s.codes_claimed) AS codes_claimed, 0 AS codes_invalid, 0 AS user_reports_issued, 0 AS user_reports_claimed, 0 AS tokens_claimed, 0 AS tokens_invalid, 0 AS user_report_tokens_claimed, array[]::integer[] AS code_claim_age_distribution, 0 AS code_claim_mean_age, array[0,0,0]::bigint[] AS codes_invalid_by_os, 0 AS user_reports_invalid_nonce FROM ( SELECT date::date FROM generate_series($1, $2, '1 day'::interval) date ) d LEFT JOIN realm_stats s ON s.date = d.date WHERE s.realm_id NOT IN ($3) GROUP BY d.date ORDER BY date DESC","build_id":"unknown","build_tag":"unknown","caller":"pkg/database/realm.go:2009","duration":0.0079645,"values":["2022-06-15T00:00:00Z","2022-09-13T00:00:00Z","'{}'"],"rows":0}
{"severity":"ERROR","timestamp":"2022-09-12T18:11:52.690191-07:00","caller":"controller/controller.go:80","message":"internal error","request_id":"b589fa51-5da8-4452-a55e-fc2e10ea2023","error":"pq: invalid input syntax for type integer: \"{}\""
```

pkg/database/realm.go
sethvargo: So maybe change to `*Realm` instead

pkg/database/realm.go
```diff
@@ -1974,11 +1999,14 @@ func (db *Database) AllRealmCodeStats(ctx context.Context, requiredRealms int) (
 			SELECT date::date FROM generate_series($1, $2, '1 day'::interval) date
 		) d
 		LEFT JOIN realm_stats s ON s.date = d.date
+		WHERE NOT s.realm_id = ANY ($3)
 		GROUP BY d.date
 		ORDER BY date DESC`
 
+	values := []interface{}{start, stop, pq.Array(allExcludedRealmIDs)}
```
sethvargo, line 2006:
```suggestion
	values := []any{}{start, stop, pq.Array(allExcludedRealmIDs)}
```

pkg/database/realm.go
mikehelmick: I missed the DB constraint - I could not find an existing function for this

pkg/database/realm.go
```diff
@@ -1947,13 +1959,24 @@ func (db *Database) AllRealmCodeStatsCached(ctx context.Context, cacher cache.Ca
...
 	start := stop.Add(project.StatsDisplayDays * -24 * time.Hour)
 	if start.After(stop) {
 		return nil, ErrBadDateRange
 	}
 
+	allExcludedRealmIDs := make([]uint, 0)
+	// Find and exclude e2e test realm if it exists.
+	e2eTestRealm, err := db.FindE2ETestRealm()
+	if err != nil {
+		db.logger.Warnf("unable o find e2e test realms for exclusion from system stats", "error", err)
```
sethvargo, line 1973:
```suggestion
		db.logger.Warnf("failed to find e2e test realms for exclusion from system stats", "error", err)
```

sethvargo (approved):

PR #2381, opened 2022-11-15
Allow sys operators to override generated SMS

mikehelmick:
**Release Note**

```release-note
Allows system operators to override the generated SMS setting on a per-realm basis.
```


mikehelmick:
keep open

